Issue #69, opened 2019-04-01
Command for index management (add/remove)


mosteo:
We need a way to work with different indexes, local or remote. Since remote indexes will imply a download prior to being usable, it seems sensible to store them for reuse in the current configuration folder (see #68), instead of specifying them every time in the command line. Local ones could be soft-linked.

Thus a possible approach is having an `alr index` command to add/remove/update indexes to the working set.

N.b.: there is `alr update` already that optionally refreshes the index, although the main purpose is to recompute a solution to dependencies

mosteo:
Just saw that ocam manages this through `opam repository`

mosteo:
`alr update --online` currently updates the index and then solves dependencies (after addition/modification). It's probably cleaner to separate both functionalities:

- `alr index --update` downloads up-to-date indexes
- `alr update` solves dependencies (perhaps this should be part of `with`: `alr with --solve`?)

mosteo:
Implemented by #121.

PR #72, opened 2019-04-05
Config and Source folders overriding

### LINKED ISSUE #68: Allow overriding of the configuration folder location
This enables:
- Testing different alr versions simultaneously
- Having different sets of indexes (public/private/mixed)

mosteo:
Config folder can be overriden with env var ALR_CONFIG or switch -c, --config
Source folder (currently used for index and templates) can be overriden with env var ALR_SOURCE

Closes #68


src/alire/alire-config.adb
```diff
@@ -0,0 +1,32 @@
...
+   end Path;
+
+   procedure Set_Path (Path : String) is
+   begin
+      if Config_Path /= null then
+         raise Constraint_Error with "Custom path already set";
+      else
+         Config_Path := new String'(Path);
+      end if;
+   end Set_Path;
```
pmderodat: In Alr.Platforms, a comment says that the config directory is used to store alr’s source code. Given that we’re moving away from recompilation, is the config directory still relevant?

src/alr/alr-bootstrap.adb
```diff
@@ -27,6 +29,28 @@ package body Alr.Bootstrap is
...
+
+      OS_Lib.Spawn ("git", "clone " & Paths.Alr_Repo & " " & To_Path);
+
+      declare
+         Enter : OS_Lib.Folder_Guard (OS_Lib.Enter_Folder (To_Path)) with Unreferenced;
+      begin
+         OS_Lib.Spawn ("git", "submodule update --init --recursive");
+      end;
+   end Checkout_Alr_Sources;
+
```
pmderodat: If bootstrap (and self-recompilation in general) is to become obsolescent, why do we do this? I’m not familiar with this part, so not sure if getting rid of self-recompilation is something that is easy to do or if it will require more work ^^

mosteo: There is still a `templates` folder used by the `init` command. At some point we can download only things really needed; right now to minimize changes I kept the checkout as it was.

Recompilation should not happen any more after #71. Commands that relied on it are broken until I finish #73 though.

src/alire/alire-config.adb
mosteo: I also thought that we could get rid of it. Then I remembered a couple things we need the config folder for:
- Templates for initialization of projects (inside the source checkout in the other comment).
- The downloaded indexes (related: #69).

The source override is only meant for us developers, so we can tinker with resources under version control directly in the source tree (the aforementioned templates), if the need ever arises.

mosteo:
The commits are really mostly nonsensical in this one because of the rebase (why are there repetitions?). I am still trying to get my bearings about the best way of submitting changes, so bear with me for a while ;-)

src/alire/alire-config.adb
pmderodat: Ah, good point! It would look more natural to me if the downloaded indexes went to the `cache` directory, but that’s for another time :)

src/alr/alr-bootstrap.adb
pmderodat: Ok, thanks for the clarification!

pmderodat:
As I said on Gitter: no problem regarding commits, I understand it’s difficult to keep things in order during such breaking changes. :-)

PR #76, opened 2019-04-09
Work on repairing pre-TOML migration features

pmderodat:


alr.gpr
```diff
@@ -18,9 +21,56 @@ project Alr is
...
+      when "release" =>
+         --  Build with lots of optimizations, generate debug info (useful for
+         --  tracebacks) and use maximum runtime checks. Also enable all
+         --  warnings, but do not turn them into errors.
+         Ada_Switches := (
+            --  Build with lots of optimizations. Generate debug info (useful
+            --  for tracebacks).
+            "-O2", "-g",
+
+            --  Force the use of Ada 2012
```
mosteo: Shouldn't this be common to any build mode?

mosteo: I mean the -gnat2012 switch (bad context)

pmderodat: Ah yes, you are right.

Before I try to fix that, I would like to know: what is the purpose of using this flag? Since GNAT Community 2012, only the latest Ada version is supported, so this will break builds with GNAT Community for the next release that supports Ada 202X. Also, whether there is this flag or not, the build will stop if using a compiler that is old enough not to support Ada 2012 by default.

mosteo: There is no need then, I thought you were adding it for some reason but then it must be some remainder from my old projects, so let's get rid of it.

pmderodat: Ok thanks! I’ll take care of it ASAP

PR #77, opened 2019-04-10
Separate alire.gpr from alr.gpr

### LINKED ISSUE #70: Separate alire from alr as library within the same repo
To simplify development we have reunited all codebase in one repository. However, we still have two separate package hierarchies:

- `alire.*` packages contained the index (in Ada specs) and related data types.
- `alr.*` contained all other functionality.

Since now the index is textual, we decided to reuse these hierarchies in the following sense:

- `alire.*` will contain all reusable functionality (from alr or any other code, hypothetically a GUI).
- `alr.*` will contain only stuff used by the command-line tool.

The migration will be progressive for old functionality, and new one will respect this separation. This issue is concerned only with the creation of the alire.gpr library project that separates compilation (which is currently centralized in alr.gpr).

mosteo:
And related changes to avoid existing inverse dependencies alire -> alr

Also removed the dependency on global status in Alire.Features.Index.
The idea is to progressively remove other globals that muddle the current code base.

Closes #70

### DIFF
--- a/alire.gpr
+++ b/alire.gpr
@@ -1,27 +1,26 @@
 with "aaa";
 with "ada_toml";
-with "alire";
 with "ajunitgen";
 with "semantic_versioning";
 with "simple_logging";
 with "xml_ez_out";
 
-project Alr is
+library project Alire is
+
+   for Library_Name use "alire";
 
    type Any_Build_Mode is ("debug", "release");
    Build_Mode : Any_Build_Mode := external ("ALR_BUILD_MODE", "debug");
 
-   for Source_Dirs use ("src/alr", 
-                        "src/alr/os_linux");
+   for Source_Dirs use ("src/alire", 
+                        "src/alire/os_linux");
 
+   for Library_Dir use "lib";
    for Object_Dir use "obj";
-   for Exec_Dir use "bin";
-   for Main use ("alr-main.adb");
 
-   for Languages use ("Ada", "C");
+   for Languages use ("Ada");
 
    Ada_Switches := ();
-   C_Switches := ();
 
    case Build_Mode is
       when "debug" =>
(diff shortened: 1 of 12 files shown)


alire.gpr
```diff
@@ -0,0 +1,73 @@
...
+with "semantic_versioning";
+with "simple_logging";
+with "xml_ez_out";
+
+library project Alire is
+
+   for Library_Name use "alire";
+
+   type Any_Build_Mode is ("debug", "release");
+   Build_Mode : Any_Build_Mode := external ("ALR_BUILD_MODE", "debug");
```
pmderodat: I guess the variable should be called `ALIRE_BUILD_MODE` instead

src/alire/alire-features-index.adb
```diff
@@ -1,35 +1,27 @@
...
-
 package body Alire.Features.Index is
 
    --------------
    -- Load_All --
    --------------
 
-   procedure Load_All is
+   procedure Load_All (Platform : Environment.Setup;
+                       From    : Absolute_Path) is
```
pmderodat: Misaligned colons :)

src/alire/alire.ads
```diff
@@ -55,6 +55,22 @@ package Alire with Preelaborate is
    --  This type is used to ensure that folder separators are externally always '/',
    --  and internally properly converted to the platform one
 
+   --  To clarify constants/functions declared herein:
+
+   subtype Absolute_File is Platform_Independent_Path;
+   --  Filenames with full path
+
+   subtype Absolute_Path is Platform_Independent_Path;
```
pmderodat: What is the difference between Absolute_File and Absolute_Path? (same question for Relative_*)

mosteo: The intent is that *_File is a regular file and *_Path is a folder. Perhaps the latter should be renamed to *_Folder (or *_Directory)

PR #81, opened 2019-04-14
Exporting of releases to TOML file

mosteo:
The gist of this PR is to re-enable the capability of exporting an in-memory release to a file (previously an Ada spec, now a TOML index file).

This was the major blocker towards reinstating most commands that rely on the existence of this file.

There are some unrelated commits in this PR due to my previously wrong understanding of how dependent PRs work. Also there are a few big commits that affect many files when a fundamental data structure is modified. There are even some backtracks. Sorry about all of those.

For these reasons I'm not asking for an in-detail review, it's more so you can see where things are going. Also, since most things should start working again soon, it will be easier to do more focused PRs.


mosteo:
Obsoleted by PRs #83 #84 #85 #86 #87

Issue #80, opened 2019-04-11
Dump in-memory release to TOML file


mosteo:
This is the main blocker for #73

mosteo:
Completed by #97 in a practical sense. Remaining issues to be addressed separately.

Issue #73, opened 2019-04-05
Reinstate pre-toml functionality


mosteo:
Although index loading and listing are mostly functional, most other commands are now broken. Passing tests in ./stress/selftest.sh are the bare minimum to be back on track.

- [x] build (commit c06fa7a267951e2584062d6c87e175a413f604ce)
- [x] clean (commit ac932d8e8db4514c57766356702cb632c7293792)
- [x] compile (commit c06fa7a267951e2584062d6c87e175a413f604ce)
- [x] get (commit ac932d8e8db4514c57766356702cb632c7293792)
  - [x] --only (commit ac932d8e8db4514c57766356702cb632c7293792)
- [x] init (commit aeed330)
- [x] list
- [x] pin (commit 47d9953)
- [x] run (commit c06fa7a267951e2584062d6c87e175a413f604ce)
- [x] search
- [x] show (commit b0b8e4b)
  - [x] show \<project\> (commit 17adb7)
  - [x] --native
- [x] test (commit e4fcb0159d160df122e31d6180e7063f6326f509)
- [x] update (commit c06fa7a267951e2584062d6c87e175a413f604ce)
- [x] version
- [x] with (commit e6eddfef366811ccbc7c1942a17729f795e21075)

PR #84, opened 2019-04-15
Export releases to TOML #1

mosteo:
First half of patches towards fixing #80 and #73


.gitmodules
```diff
@@ -18,7 +18,7 @@
 	url = https://github.com/alire-project/simple_logging.git
 [submodule "deps/ada-toml"]
 	path = deps/ada-toml
-	url = https://github.com/pmderodat/ada-toml.git
+	url = https://github.com/mosteo/ada-toml.git
```
pmderodat: It’s not clear to me why we need this, but if it’s important to have a separate fork, maybe hosting this fork in alire-project would be better, no?

mosteo: It's until my two PRs to your repo are merged, at which point I plan to switch back to yours. Is there a better way to do this?

I thought about forking under alire, but since I didn't intend for it to become the main one I thought it could cause confusion.

pmderodat: Understood. I think the best way would be to close the loop on your PR for ada-toml :-) This shouldn’t take long: the first one is already merged: do you need help for what’s missing? (only the testcase IIRC).

PR #86, opened 2019-04-15
Export releases to TOML #2

mosteo:
Second half of good changes

### DIFF
--- a/src/alire/alire-dependencies.adb
+++ b/src/alire/alire-dependencies.adb
@@ -1,0 +1,18 @@
+with Alire.TOML_Adapters;
+
+package body Alire.Dependencies is
+
+   overriding function To_TOML (Dep : Dependency) return TOML.TOML_Value is
+      use Semantic_Versioning;
+      use TOML_Adapters;
+   begin
+      if Dep.Versions = Any then
+         return +"*";
+      elsif Length (Dep.Versions) > 1 then
+         raise Unimplemented; -- TODO (but not yet in index format)
+      else
+         return +Operator_Image (Element (Dep.Versions, 1));
+      end if;
+   end To_TOML;
+
+end Alire.Dependencies;
--- a/src/alire/alire-dependencies.adb
+++ b/src/alire/alire-dependencies.adb
@@ -7,7 +7,7 @@ package body Alire.Dependencies is
       use TOML_Adapters;
    begin
       if Dep.Versions = Any then
-         return +"*";
+         return +"any";
       elsif Length (Dep.Versions) > 1 then
          raise Unimplemented; -- TODO (but not yet in index format)
       else
--- a/src/alire/alire-properties-labeled.ads
+++ b/src/alire/alire-properties-labeled.ads
@@ -44,6 +44,8 @@ package Alire.Properties.Labeled with Preelaborate is
 
    overriding function Image (L : Label) return String;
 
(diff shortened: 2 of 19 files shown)


src/alire/alire-properties-labeled.ads
```diff
@@ -88,6 +90,16 @@ private
...
 
+   overriding function Key (L : Label) return String is
+     (case L.Name is
+         when Author       => "authors",
+         when Comment      => "comment",
+         when Executable   => "executables",
+         when Maintainer   => "maintainers",
+         when Path         => "paths",
+         when Project_File => "project-files",
+         when Website      => "website");
```
pmderodat: Looks like this should directly reference `Alire.TOML_Keys`.

src/alire/alire-properties-licenses.ads
```diff
@@ -1,11 +1,17 @@
 with Alire.Licensing;
+with Alire.TOML_Keys;
 
 package Alire.Properties.Licenses is
 
    function Image (L : Licensing.Licenses) return String;
 
+   pragma Warnings (Off);
+   function Key (L : Licensing.Licenses) return String is (TOML_Keys.License);
+   pragma Warnings (On); -- Unreferenced => L
```
pmderodat: Protip: rename `L` to `Dummy` (or `Dummy_L`), and the warning will go away.

src/alire/alire-releases.adb
```diff
@@ -459,42 +459,44 @@ package body Alire.Releases is
...
+--                                        when Multiple => Values.To_TOML);
+--                 begin
+--                    General.Set (TOML_Keys.Labels (Label),
+--                                 TOML_Values);
+--                 end;
+--              end if;
+--           end;
+--        end loop;
+--
+--        Root.Set ("general",       General);
```
pmderodat: Have you left this by accident? Commented code looks messy, and no explanation about why it’s there will create doubt in the future when we’ll wonder if we have to leave it here or remove it ^^

src/alire/alire-dependencies.adb
```diff
@@ -0,0 +1,18 @@
+with Alire.TOML_Adapters;
+
+package body Alire.Dependencies is
+
+   overriding function To_TOML (Dep : Dependency) return TOML.TOML_Value is
+      use Semantic_Versioning;
+      use TOML_Adapters;
+   begin
+      if Dep.Versions = Any then
+         return +"*";
```
pmderodat: The string should be `any` (that’s what `Alire.TOML_Index` expects)

src/alire/alire-properties-labeled.ads
mosteo: Indeed!

src/alire/alire-properties-licenses.ads
mosteo: Is this some gnat magic? :D

src/alire/alire-releases.adb
mosteo: Right you are, to be deleted

src/alire/alire-properties-licenses.ads
pmderodat: Yes indeed. ;-) See https://docs.adacore.com/gnat_rm-docs/html/gnat_rm/gnat_rm/implementation_defined_pragmas.html#pragma-unmodified

PR #90, opened 2019-04-17
Load a single release from a non-index TOML file

mosteo:
Commands that apply to a downloaded project rely on getting the release info from a TOML file (since this can be an unreleased project, or a release that has been modified).

I have moved things a bit around in TOML_Index to be able to load a release without indexing it. Since this is your turf, maybe I have missed something in the process.

The main change is adding an extra step after decoding a Pkg, so the releases are not directly indexed.

### DIFF
--- a/src/alire/alire-toml_index.adb
+++ b/src/alire/alire-toml_index.adb
@@ -347,10 +347,11 @@ package body Alire.TOML_Index is
    -- Load_Release_From_File --
    ----------------------------
 
-   function Load_Release_From_File
+   procedure Load_Release_From_File
      (Filename    : String;
-      Environment : Environment_Variables)
-      return        Releases.Release
+      Environment : Environment_Variables;
+      Release     : out Containers.Release_Holders.Holder;
+      Result      : out Load_Result)
    is
       Pkg      : Package_Type;
       Value    : TOML.TOML_Value;
--- a/src/alire/alire-toml_index.adb
+++ b/src/alire/alire-toml_index.adb
@@ -376,7 +376,7 @@ package body Alire.TOML_Index is
 
       --  Generate the releases to be imported
 
-      Import_TOML_Package_As_Releases (Pkg, Environment, Releases);
+      Decode_TOML_Package_As_Releases (Pkg, Environment, Releases);
 
       --  Verify and return
       if Releases.Is_Empty then
@@ -507,7 +507,7 @@ package body Alire.TOML_Index is
 
       --  Generate the releases to be imported
 
-      Import_TOML_Package_As_Releases (Pkg, Environment, Releases);
+      Decode_TOML_Package_As_Releases (Pkg, Environment, Releases);
 
       --  Finally import them to the catalog
 
@@ -1594,7 +1594,7 @@ package body Alire.TOML_Index is
    -- Import_TOML_Package_As_Releases --
    -------------------------------------
 
-   procedure Import_TOML_Package_As_Releases
+   procedure Decode_TOML_Package_As_Releases
      (Pkg         : Package_Type;
       Environment : Environment_Variables;
       Releases    : out Containers.Release_Sets.Set)
@@ -1888,7 +1888,7 @@ package body Alire.TOML_Index is
       when Evaluation_Error =>
          Trace.Error (+Error_Message);
          null;
-   end Import_TOML_Package_As_Releases;
+   end Decode_TOML_Package_As_Releases;
 
    --------------------
    -- Index_Releases --
--- a/src/alire/alire-toml_index.adb
+++ b/src/alire/alire-toml_index.adb
@@ -1868,7 +1868,11 @@ package body Alire.TOML_Index is
                  (Project            => +(+Pkg.Name),
                   Version            => R.Version,
                   Origin             => Origin,
-                  Notes              => +Pkg.Common.Notes,
+                  Notes              => +US.Head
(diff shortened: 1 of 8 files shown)


src/alire/alire-toml_index.adb
```diff
@@ -343,6 +343,48 @@ package body Alire.TOML_Index is
...
+      if not Result.Success then
+         raise Constraint_Error with "could not load " & Filename;
+      end if;
+
+      --  Convert it to our intermediate data structures
+
+      Decode_TOML_Package
+        (Filename, Ada.Directories.Base_Name (Filename), Environment, Value, Pkg, Result);
+      if not Result.Success then
+         raise Constraint_Error with +Result.Message;
```
pmderodat: Except when the language enforces the use of a Constraint_Error exception (for instance when decoding an integer from a string) and if practicle in general, I think it’s better to use specific types to convey errors to handle, as I did here with `Load_Result`, and leave exceptions for bona fide bugs.

The reasons for that is that 1) it’s just too easy to just forget exception handling in callers, and in general exceptions make it hard to have a good understanding of what can raise and when/how to handle errors, and 2) IIRC the language does not guarantee that it will preserve the whole exception message; I think GNAT will preserve only the first 200 characters: it’s just too easy to hit this limit with user error messages, especially when they include filenames!

In this specific case, I would advise to use an out `Load_Result` parameter to communicate to callers whether this is successful.

src/alire/alire-toml_index.adb
```diff
@@ -1860,7 +1860,11 @@ package body Alire.TOML_Index is
                  (Project            => +(+Pkg.Name),
                   Version            => R.Version,
                   Origin             => Origin,
-                  Notes              => +Pkg.Common.Notes,
+                  Notes              => +US.Head
+                    (Pkg.Common.Notes, Alire.Max_Description_Length),
+                  -- It crops too long notes, so something TODO about this
+                  -- Since it didn't fail before, I guess they weren't added
+                  --  this way (only as property?)
```
pmderodat: Exactly :-) Given that there are more powerful Notes properties, what is the purpose of this argument? Could we merge the two?

src/alire/alire-toml_index.adb
```diff
@@ -343,6 +343,48 @@ package body Alire.TOML_Index is
...
+        (Filename, Ada.Directories.Base_Name (Filename), Environment, Value, Pkg, Result);
+      if not Result.Success then
+         raise Constraint_Error with +Result.Message;
+      end if;
+
+      --  Generate the releases to be imported
+
+      Import_TOML_Package_As_Releases (Pkg, Environment, Releases);
+
+      --  Verify and return
```
pmderodat: These checks happen after we have imported the releases in the in-memory index, so this procedure kind of corrupts the state in case of error. To fix this, what about moving these checks before the call to `Import_TOML_Packages_As_Releases`? (checking `Pkg.Releases` instead of `Releases`)

src/alire/alire-toml_index.adb
mosteo: Note that this is a truly exceptional situation, since the file this is intended to load is generated by alr and should always be correct. That said, I'll follow suit with the rest of the package and use the Result record.

src/alire/alire-toml_index.adb
mosteo: IIRC, in the previous index, Note was not a property but a unique field of the release. Comments OTOH where managed as properties (and there could be several in the same release). 

I see in line 1497 that you are importing the note as a comment. This is conflating the two, so when a release is later exported it will have been modified by alr, which to me is a no-no.

I propose, for now, to leave notes as they were originally (I can fix that in this same PR) and, if we want to homogenize the notes/comments situation, to open a specific issue for it. It's certainly confusing having notes and comments; perhaps we can change to general(one line descr + long desc) + release(notes) and that's it? But we can discuss that in the issue, if you agree.

src/alire/alire-toml_index.adb
mosteo: The intention is that Import_TOML_Packages_As_Releases no longer imports into the in-memory index (the Register call is moved to the new Index_Releases, lines 503-507). So Load_Release_From_File doesn't touch the in-memory index. (I'll verify that.)

If the Import_* in the procedure name is confusing I could rename to Convert_TOML_Package_To_Releases or something like that, if you prefer.

src/alire/alire-toml_index.adb
mosteo: And just remembered that users can edit the file manually so I retract my comment :s

src/alire/alire-toml_index.adb
mosteo: Renamed to Decode_TOML_Package_To_Releases

src/alire/alire-toml_index.adb
pmderodat: :-) This objective is hard to reach, but I think it’s valuable to avoid as much as possible uncaught exceptions in error cases: they make the tool look unpolished…

src/alire/alire-toml_index.adb
pmderodat: Thanks for the clarification and for opening #92! I agree we should do this as a next step.

src/alire/alire-toml_index.adb
pmderodat: Looks good, thanks!

PR #125, opened 2019-05-29
 Second pass on coding style 

Fabien-Chouteau:
As of this patch, all the style errors are fixed.

A scenario variable is also introduced in the project file to control style checking.


mosteo:
Thanks a lot!

PR #126, opened 2019-05-29
Remove dependencies on external template files

pmderodat:


src/alr/alr-commands-init.adb
```diff
@@ -18,38 +19,144 @@ package body Alr.Commands.Init is
...
+         Put_Line ("      for Switches (""ada"") use");
+         Put_Line ("        (""-gnatVa"", ""-gnatwa"", ""-g"", ""-O2"",");
+         Put_Line ("         ""-gnata"", ""-gnato"", ""-fstack-check"");");
+         Put_Line ("   end Compiler;");
+         Put_New_Line;
+         Put_Line ("   package Binder is");
+         Put_Line ("      for Switches (""ada"") use (""-Es"");");
+         Put_Line ("   end Binder;");
+         Put_New_Line;
+         Put_Line ("end project " & Mixed_Name & ";");
```
mosteo: "end project " --> "end "

src/alr/alr-commands-init.adb
```diff
@@ -43,12 +43,15 @@ package body Alr.Commands.Init is
       end if;
 
       declare
-         Guard : Folder_Guard (
```
mosteo: Removing this one causes a spurious 'alire' folder to be created in the CWD. Possible fix is to change line 163 from 
`OS_Lib.Create_Folder (Paths.Alr_Working_Folder);` 
to 
`OS_Lib.Create_Folder (Name / Paths.Alr_Working_Folder);`
(without undoing the deletion)

src/alr/alr-commands-init.adb
pmderodat: Oops, I tested the init command but forgot to do a build after it… will fix, thanks!

src/alr/alr-commands-init.adb
pmderodat: Thank you for catching this! I’ll have a look.

mosteo:
Sorry for having this one so long in the cooker, and thanks!

pmderodat:
No problem… thanks for the review!

PR #127, opened 2019-05-30
Third pass on coding style

mosteo:
I've enabled -gnatDIO, only O needed fixes.

@Fabien-Chouteau, now that the style is fixed I'd prefer that the checks be enabled by default; your thoughts?

### DIFF
--- a/src/alire/alire-actions.adb
+++ b/src/alire/alire-actions.adb
@@ -17,7 +17,7 @@ package body Alire.Actions is
    -- To_TOML --
    -------------
 
-   function To_TOML (This : Run) return TOML.TOML_Value is
+   overriding function To_TOML (This : Run) return TOML.TOML_Value is
 
       use TOML_Adapters;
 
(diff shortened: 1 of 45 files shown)


pmderodat:
Sorry if I forgot guys, but what do you think about using GNAT’s own style checks here? (`-gnatyg`)

Also I’m all for enabling the checks once the style is fixed :) Thank you both for your work on this!

mosteo:
I've tried -gnatyg and it only complains about multiple blank lines. I guess these can be improved by adding a comment about what that extra separation intended to draw attention to, if anything.

So, if no objection from @Fabien-Chouteau, I'm going to enable and fix it in this PR.

mosteo:
I finally didn't use -gnatyD, since it doesn't allow acronyms.

src/alire/alire-actions.adb
```diff
@@ -17,7 +17,7 @@ package body Alire.Actions is
    -- To_TOML --
    -------------
 
-   function To_TOML (This : Run) return TOML.TOML_Value is
+   overriding function To_TOML (This : Run) return TOML.TOML_Value is
```
Fabien-Chouteau: I always put `overriding` on the line above `function`.

mosteo: OK, will do in the future, thanks.

PR #131, opened 2019-06-24
Enable new test suite in CI

mosteo:


pmderodat:
Nice! Just in case: don’t we also need some configuration update to install the e3-core and e3-testsuite Python packages, that the testsuite requires?

mosteo:
I'm including that in our custom docker images in https://cloud.docker.com/u/alire/repository/docker/alire/gnat/ so it doesn't get fetched every time. It seems to work with only e3-testsuite, I guess pip brings in e3-core as a dependency?

pmderodat:
Exactly. Okay, so good to go I guess :-) (except that the shippable run failed…)

mosteo:
Yep, that's because the new docker files are still building. I thought that pushing a draft PR would not trigger CI, but it seems shippable is not discriminating.

mosteo:
Hi @pmderodat, while debugging a failing test here I noticed a couple of things:

1) For operations that may take some time (e.g. cloning of repos), in the default verbosity level of alr, there's an ASCII spinner that makes the output unusable for testing, since it depends on timing. In those cases (and probably in general), it will be better to use 'alr -q' to get output only in case of error (see last commit in the PR).

2) In case of test failure, there is no way (I think) to retrieve after the fact the testsuite/out folder in Shippable. For now I have added a dump to stdout of all the test logs in case of failure. But:

Shippable (and some other CI tools) use junit-formatted output files to report individual test results and logs (see http://docs.shippable.com/ci/test-code-coverage-reports/). For example, I'm using that for crate testing in https://app.shippable.com/github/alire-project/alire-crates-ci/runs/38/summary/console . I was wondering if the YAML output that e3 generates is some kind of standard?

As an interim measure, it could be convenient if the e3 test suite dumped the log of failed tests also to stdout/stderr (perhaps activated on demand by some switch?).  Although I'm thinking now that if you're using e3 with Travis in some AdaCore projects, you'll already have a solution for this kind of situation?

Long-term, it would be neat if e3 could output junit-formatted results...

pmderodat:
Thanks for adding `-q`! I think it could be worth adding an environment variable for this so that the testsuite sets it once and all testcase benefits from it. I’d like to try that, if that’s okay for you.

Regarding JUnit, I don’t think `e3.testsuite` has support to generate such reports, but this sounds definitely useful. I’ll start a discussion with maintainers. :-) The YAML format is completely custom IIRC. I’ll see what I can do to somehow dump the results and logs somewhere convenient for CI tools to pick them up.

mosteo:
Sure, go ahead.

Issue #129, opened 2019-06-24
Migrate ./stress/selftest.sh to new test suite


mosteo:
With the new test suite in place, it would be nice to have all tests done there and get rid of the old custom test script.

pmderodat:
I can have a look at this: this will be a good opportunity to add missing features to the testsuite framework.

PR #133, opened 2019-06-25
Add testcases and fix minor issues

pmderodat:
Hello @mosteo 

I think I have ported most tests from `selftest.sh` to the new testsuite, except tests that cover the `update` command. I’m not familiar with it and need more time to know what it’s supposed to do. Please tell me if I missed something. Thanks!


mosteo:
Actually the update command is something that I wanted to revisit. How it goes:

It reads dependencies, resolves them, and if necessary fetches any new versions of crates to generate a buildable environment. Changes in regard to the original `get` or `init` may come from the user modifying dependencies/versions, or from new index contents. For the latter, there was (before the new index management) and `update --online` that pulled the master branch of the alire-index repo. That ceased to exist quite a while ago.

Right now, we have an `index` command to manage indexes, including one option to fetch updates; and a `with` command that is used to add/remove dependencies. This `with` command has no direct equivalent in the base `cargo`, to my knowledge.

So I nowadays tend to favor the idea of getting rid of `update` and integrating it into, e.g., `with --update`, since that's the umbrella command for dependency edition. Basically, it would update indexes and the re-resolve dependencies (in current, `alr`, equivalent to `$ alr index --update-all; alr update`. Also, `update` is too generic whereas `with` has a clearer Ada meaning.

I will review your PR today, thanks!

pmderodat:
Understood, so I’ll leave testcases as they are for now, considering that there’s nothing left to do for #129, thanks! Reading your message, I realize that there are a lot of tests to write for index management commands: I’ll have a look at this.

testsuite/tests/workflows/init-with-pin/test.py
```diff
@@ -0,0 +1,67 @@
...
+os.chdir('xxx')
+
+# Make it depend on libhello
+session_file = os.path.join('alire', 'xxx.toml')
+run_alr('with', 'libhello')
+check_line_in(session_file, 'libhello = "any"')
+
+# Add the corresponding "with" line in xxx.gpr.
+#
+# TODO: maybe "alr with" should do that automatically?
```
mosteo: This is something I have considered in the past. For releases with a single GPR it seems the obvious thing to do. For ones with several... I guess we shouldn't (or we might ask the user interactively?)

As con, I don't like to touch user files directly. Again, we could ask beforehand (incidentally, we have -n to avoid interactivity -- which currently only happens before doing a `sudo apt-get install`)

mosteo:
Merged, thanks for the comprehensive tests.

pmderodat:
And thank you for the review :)

testsuite/tests/workflows/init-with-pin/test.py
pmderodat: Good points. Anyway, tt would be nice to have but clearly not needed :)

PR #138, opened 2019-06-27
Fix possibility of forbidden chars in folder name

mosteo:
Remove the patch/build parts of a version from the crate deployment folder.
Crate folder is still unique due to the commit part


mosteo:
I've gone ahead and merged this because it's minor. Don't hesitate to bring up any issues.

PR #143, opened 2019-07-05
Start to use GNATCOLL.VFS to switch to platform-independent path processings

pmderodat:


src/alire/alire-features-index.ads
```diff
@@ -10,8 +10,13 @@ package Alire.Features.Index is
 
    subtype Index_On_Disk_Set is Sets.Set;
 
-   function Find_All (Under : Absolute_Path) return Index_On_Disk_Set;
-   --  Find all indexes available on a disk location
+   function Find_All
+     (Under  : Absolute_Path;
+      Result : out Outcome) return Index_On_Disk_Set;
+   --  Find all indexes available on a disk location. If valid indexes are
+   --  found or none, set Result to Outcome_Success and retun the corresponding
```
mosteo: misspelled "return"

mosteo:
Lots of improvements here, thanks. I see two kind of errors in CI:

1. Compiling with the community edition results in license inconsistencies with GPR
2. The non-community compiler versions (debian/ubuntu) can't find project GPR for GNATCOLL. 

Now I remember that 2) was the reason behing the "gnatcoll-slim" crate (having gnatcoll with no other dependencies). I see that ubuntu (and presumably debian) include those:
```
libgpr1-dev - support for programs processing GNAT projects (development)
libgpr16.1.2016 - support for programs processing GNAT projects (runtime)
```

Still, not depending on GPR, since that's not currently needed by alire, would be better. Any chances that gnatcoll-gpr might be split from gnatcoll-core? Or any other ideas?

src/alire/alire-features-index.ads
pmderodat: Nicely spot :-) Will fix, thanks!

pmderodat:
Ah sorry, I forgot about that. GNATcoll has its own dependencies, but they are all contained in project files, so fixing this should be easy. I’ll complete the commits…

pmderodat:
I forgot to answer the other questions :)

There are plans to remove the dependency on LibGPR (with LibGPR2 depending itself on GNATCOLL), but not in the near future. My plan is to add the dependency on LibGPR and XML/Ada just like the other Alire dependencies, as they are simple project files like the other Alire’s deps beyond the initial compilation time (which isn’t that short already, anyway), this should be transparent to people using `alr_env.gpr`.

pmderodat:
Unfortunately, I was too optimistic! XML/Ada still requires to run a `./configure` script before running `gprbuild`. I started an internal discussion at AdaCore, and people would be favorable to patch XML/Ada remove this requirement.

Now, I’ll be on vacation from Friday, 12th to Sunday, 21st (both included). We probably won’t be able to do this XML/Ada enhancement before then: are you fine temporarily requesting Alire developers (i.e. mainly you and me these days ;-)) to build and install XML/Ada before building using `alr_env.gpr`? Otherwise I’m afraid my patches will bit rot for two weeks… ^^’ I’ve just updated the pull request assuming it’s okay in the meantime.

pmderodat:
Arf, likewise for LibGPR actually: there is a clash between `gnatcoll.gpr` and `gpr.gpr` regarding the possible values for the `BUILD` external variable. I’ll need to patch one of them too in order to solve this issue.

The workaround should be the same, though: build and install LibGPR before running `gprbuild` on `alr_env.gpr` (and I would need to remove the `gprbuild` submodule in the meantime).

mosteo:
Yup, these problems are starting to ring some bells. It probably doesn't help us that gnat community editions transparently bundle some projects, while debian (and, I guess, other distros) gnat's doesn't. I fear this will come to bite us in dependency packaging at some point.

I'm thinking that perhaps the eviscerated gnatcoll-slim, although a bit old, has what we need until these issues are solved upstream? I might take a stab at it while you're gone (enjoy, btw).

mosteo:
Closing, superseded by #152.

PR #150, opened 2019-07-09
Improvements to help given by alr

### LINKED ISSUE #139: Improve help command
- [x] Document what `procedure Display_Help_Details (Cmd : Command)` in `alr-commands.ads:35` does.
- [x] Fill it in for commands not using it yet.

mosteo:
- Mandatory explanation for every command
  - With better formatting
- Made `alr command -h` equivalent to `alr help command`
- Test for said equivalence.

Closes #139


pmderodat:
I know we can already do that running `alr` without arguments. Maybe running `alr --help` should do that as well?

Fabien-Chouteau:
Right, I didn't know about that.

I think it would be nice to have the list when doing `alr --help` and `alr help`.

Fabien-Chouteau:
Change proposed in #166

PR #153, opened 2019-07-11
Detect duplicate definition of switches

### LINKED ISSUE #111: Detect conflicting global and command switches
The way switches are parsed right now is in two stages (and frankly, a bit convoluted):

1. There are global switches, that affect alr as a whole (e.g., verbosity level). Those are parsed first, ignoring unknown switches.
    1. At this stage, words not starting with '-' are also stored as alr arguments.
1. Then, the command is ascertained (first argument), and the command gets a turn to set up its own switches (via its overriding `Setup_Switches`).
    1. Switches are re-parsed, and any unknown switch is reported as an error.

This was done this way so commands are self-contained in the sense that own switch management is delegated to the command itself.

We need to somehow detect that a switch definition has changed after 2, or generally rework switch parsing. Simplest solution would be if `Gnat.Command_Line.Define_Switch erred on switch re-definition (to be looked into). Or perhaps use sections where the command name acts as separator (but then a switch before/after the command would change meaning, seems looking for trouble).

mosteo:
Ensure that a command does not redefine a previously defined global switch. More generally, check that the same switch is not defined twice, in either short or long form.

Fixes #111

### DIFF
--- a/src/alr/gnat-command_line-extra.ads
+++ b/src/alr/gnat-command_line-extra.ads
@@ -1,0 +1,11 @@
+package GNAT.Command_Line.Extra is
+
+   function Verify_No_Duplicates (Config : Command_Line_Configuration)
+                                  return Boolean;
+   --  Returns True if there are no duplicates.
+   --  Check that no switch is given twice in Config. This is used to ensure
+   --  that command switches are not stepping on global switches, which would
+   --  lead to some undefined behavior. This manual check is necessary because
+   --  the GNAT library does not perform it.
+
+end GNAT.Command_Line.Extra;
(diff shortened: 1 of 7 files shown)


src/alr/gnat-command_line-extra.ads
```diff
@@ -0,0 +1,11 @@
+package GNAT.Command_Line.Extra is
```
Fabien-Chouteau: I am not sure if it is a good idea to make add a child package to a run-time unit.

mosteo: I could not think of other "lightweight" solution... I considered as option B to have a wrapper on GNAT.Command_Line.Command_Line_Configuration, but I am not sure if it's worth the effort.

Any other ideas?

PR #156, opened 2019-07-12
Enhance -d/--debug with optional debugging scopes

mosteo:
This patch allows passing optional filters to `-d/--debug`, which are substrings that will be matched against the enclosing entity and source file of regular verbosity logging messages.

By passing e.g. `-d+main`, `-d-alr`, etc, the logging output can be filtered out in both blacklist/whitelist mode. Several filters can be given: `-d+main+commands`. 

Using the contrary sign acts as an exception to the filtering: `-d+alr-commands` will whitelist all `alr` messages, but will filter out `commands` messages.

The first sign in the argument puts the filtering in whitelist (+) or blacklist (-) mode. As special cases, giving no substrings illustrate these modes: `-d-` will use an empty blacklist, thus allowing all logging messages to pass through. Contrarily, `-d+` will use an empty whitelist, blocking all messages.

These filters act on the normal verbosity logging, so one most likely wants `-vv -d[something]` in practice.

See example testcase therein for other examples.

### DIFF
--- a/src/alire/alire_early_elaboration.adb
+++ b/src/alire/alire_early_elaboration.adb
@@ -1,12 +1,46 @@
+with Ada.Text_IO;
+
 with Alire;
---  Kind of circularity here but somehow it slips past static GNAT rules??
 
 with GNAT.Command_Line;
 with GNAT.OS_Lib;
 
-with Simple_Logging;
+with Simple_Logging.Filtering;
 
 package body Alire_Early_Elaboration is
+
+   ----------------
+   -- Add_Scopes --
+   ----------------
+
+   procedure Add_Scopes (Debug_Arg : String) is
+      --  Receives as-is the --debug/-d[ARG] argument. This is a list of
+      --  optionally comma-separated, plus/minus prefixed substrings that will
+      --  be used for filtering against the enclosing entity/source location.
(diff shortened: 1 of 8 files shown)


src/alire/alire_early_elaboration.adb
```diff
@@ -2,12 +2,46 @@ with Alire;
...
+      --  Once scopes are used, we activate logging of enclosing entity and
+      --  location to provide full logging information.
+
+   begin
+      if not Simple_Logging.Filtering.Add_From_String (Debug_Arg,
+                                                       Say => True)
+      then
+         --  Bypass debug channel, which was not entirely set up.
+         --  Otherwise we get unwanted location/entity info already.
+         GNAT.IO.Put_Line ("ERROR: Invalid logging filters.");
```
Fabien-Chouteau: Why using `GNAT.IO` instead of `Ada.Text_IO`?

Fabien-Chouteau: Why using GNAT.IO instead of Ada.Text_IO?

mosteo: I guess at some point that code was in a preelaborable library. I'll change it to plain Ada.Text_IO.

PR #176, opened 2019-09-12
Set up gnatelim for future use

mosteo:
I expect it will be useful to locate scattered dead code from the removal of the old index.

The PR adds a few mock project files since it seems gnatelim does not yet understand aggregate projects.


mosteo:
Any concerns about merging this one, @Fabien-Chouteau ?

PR #186, opened 2019-09-14
Load the index without resolving case  expressions

### LINKED ISSUE #109: Platform-independent index loading
The old index, despite the apparent plain Ada syntax, was designed to preserve conditional expressions and case statements as-is when loaded into memory. This way, the full index information was preserved until the point where it needed to be resolved for a particular platform (in essence, for resolution of dependencies).

Although this made data structures a bit more complex and memory intensive, it enabled several features/advantages:
- Complete information could be presented to the user (e.g., difference between `alr show` and `alr show --native`)
- When checking out a release for modification, the full information can be preserved in the exported toml file, so the import-export-import cycle is complete, also raising developer awareness of any platform-specific issues.
- When a project is unavailable, the user can ascertain the reason from `alr` output without having to go to the index files (see example below).
- A hypothetical future tool that managed the index in a platform-independent way would have all the information available. This may be needed for the future web interface, for example.

Example output from old index:
```
$ bin/alr show whitakers_words                              
whitakers_words=2017.9.10: William Whitaker's WORDS, a Latin dictionary
Origin: commit 27be95b8a06d7b22c0600c824cf929ab43efcf25 from https://github.com/mk270/whitakers-words.git
Available when: (not (Compiler < Gnat_Fsf_7_3_Or_Newer or Compiler = Gnat_Fsf_7_3_Or_Newer))
```
and currently:
```
whitakers_words=2017.9.10: William Whitaker's WORDS, a Latin dictionary
Origin: commit 27be95b8a06d7b22c0600c824cf929ab43efcf25 from https://github.com/mk270/whitakers-words.git
Available when: False
```

This affects both dependency expressions (that resolve to dependencies) and requisites expressions (that resolve to a boolean).

@pmderodat, this one is my fault because I should have realized from your first patches that you were evaluating the index during loading, and raised the issue sooner. So I would like to tackle it (also I guess I will have an easier time with alr internals).

My idea is not going back to the full boolean expressions, but simply keep the current index case statements during loading. I can simplify/adapt the old structures to the new index syntax at the same time.

Related to #101, depending on what we do about that.

mosteo:
Fixes #109.

This is a quite large PR, for which I apologize. Truth is, the new index syntax does not fit that well with the old data structures, since a fundamental inversion of concepts has occurred:

- Previously we had a bunch of expressions (including the trivial one: no expression) that, once evaluated, gave you the properties/dependencies to apply to the release.
- Currently, we have the name of the property, optional case expressions, and the values to apply for each case to the property.

That is, we have gone from a expr -> key -> value structure to a key -> expr -> value structure. With this PR the in-memory information is again abstract (the visible change is that expression cases are visible through `alr show`) but there are no other consequences for users. The good news are that during this rework I have seen big opportunities for code simplification that we should apply when we do the dead code removal of the old index. Since we are pressed for time, I leave that for latter as an opportunistic optimization, since that refactoring should be transparent for users.

Another point of note is that in a particular case the index syntax has changed slightly. When setting externals we have gone from:

```
[gpr-set-externals.VAR.case(os)]
linux = VALUE1
windows = VALUE2
```

to:

```
[gpr-set-externals.case(os)]
linux = { VAR = VALUE1 } # inline TOML table syntax
windows = { VAR = VALUE2 }
```

I did this change out of regularity, since this way the key -> expr -> [data to build a property] structure is maintained for all properties. I think that it's less confusing when reading, too. There are a few crates in the index affected by this. This makes necessary to update the index repository together with this PR. Once in production, for any similar breakage, we should use the versioning of indexes to transparently go from one index version to another.

In the old index, as said, expressions could appear everywhere, whereas in the own one they are applied only to specifically chosen properties. During this rework I have made dynamic a few more properties that weren't before (because they seemed to me likely to benefit from being dynamic):

- `executables` (executables that are built)
- `actions` (executing actions depending on environment)

I have also reintroduced the architecture word size as a parameter usable in case expressions.

I must say that the TOML syntax we are getting from nested tables (particularly when combined with arrays, as is the case for actions) is hard.


Fabien-Chouteau:
> gone from:
> 
> ```
> [gpr-set-externals.VAR.case(os)]
> linux = VALUE1
> windows = VALUE2
> ```
> 
> to:
> 
> ```
> [gpr-set-externals.case(os)]
> linux = { VAR = VALUE1 } # inline TOML table syntax
> windows = { VAR = VALUE2 }
> ```
> 
> I did this change out of regularity, since this way the key -> expr -> [data to build a property] structure is maintained for all properties.

What will happen if I write:
 ```
 [gpr-set-externals.case(os)]
 linux = { VAR = VALUE1 }
 windows = { ANOTHER_VAR = VALUE2 }
```

File: doc/catalog-format-spec.rst [hunk 1]
```diff
@@ -379,3 +398,4 @@ Parameters
 * ``compiler``: name of the current compiler. Currently supported values are:
   ``gnat-unknown``, ``gnat-fsf-old``, ``gnat-fsf-7.2``, ``gnat-fsf-7.3``,
   ``gnat-gpl-old``, ``gnat-gpl-2017``, ``gnat-community-2018``.
+* ``word-size``: architecture word size. Currently supported values are: ``bits-32``, ``bits-64``, ``bits-unknown``
```
Fabien-Chouteau: How will that play with cross compilation?

I guess you will need the user to specify the target architecture.

mosteo: Yes. My idea is that there will be some way for the user to override the detection of the native platform, to instead provide whatever applies to the target.

mosteo:
> What will happen if I write:
> 
> ```
> [gpr-set-externals.case(os)]
> linux = { VAR = VALUE1 }
> windows = { ANOTHER_VAR = VALUE2 }
> ```

That will result in the given variable being defined just for the corresponding case.

Fabien-Chouteau:
`VAR` will be defined only on linux and `ANOTHER_VAR` only on windows?

That is what I would expect.

mosteo:
Yes, that's it.

PR #194, opened 2019-09-16
Document use of log levels and corresponding output tweaks.

mosteo:
Foreseeing that at some point we may want nicer output, perhaps with colorization, I have explicitly commented what is the intended use of logging levels. This has not always been adhered to strictly, but by putting it in writing it should be easier to be consistent moving forward.

This should appear at some point wherever we put the information for wannabe contributors.

### DIFF
--- a/ISSUES.txt
+++ b/ISSUES.txt
@@ -1,1 +1,0 @@
-* Remove the Alr.Root package and fix dependent functionality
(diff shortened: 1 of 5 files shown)


mosteo:
The second commit removes a couple of dangling text files which information is better found in issues.

PR #196, opened 2019-09-16
Allow detection of build mode from Ada code.

mosteo:
This is intended to turn some errors into warnings during development. For example, when dealing with changes in the index it is quite cumbersome to test partial changes with older commits.

### DIFF
--- a/alire.gpr
+++ b/alire.gpr
@@ -11,7 +11,9 @@ library project Alire is
 
    for Library_Name use "alire";
 
-   for Source_Dirs use ("src/alire", "src/alire/os_linux");
+   for Source_Dirs use ("src/alire",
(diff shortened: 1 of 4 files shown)


Fabien-Chouteau:
Why do we need a compile time difference between Debug and Release mode?

mosteo:
The idea is that some checks that results in alr stopping with an error message would instead be a warning and continuing, but only during development. We could do this instead with some environment variable or command-line switch... Any preference?

Fabien-Chouteau:
My question was more on the compile time aspect of this, so yes a `--keep-going` switch sounds better in my opinion.

mosteo:
OK, will redo.

PR #199, opened 2019-09-17
Fix case insensitivity for crate names

mosteo:
This was manifesting in some dependencies being reported as unavailable.

We might go even further and force all crate names to be lowercase upon loading. For now, following the Ada example, case is preserved.

### DIFF
--- a/src/alire/alire-dependencies.adb
+++ b/src/alire/alire-dependencies.adb
@@ -10,7 +10,7 @@ package body Alire.Dependencies is
       package SV renames Semantic_Versioning;
       Version_Str : constant String := Value.As_String;
    begin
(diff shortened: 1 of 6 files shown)


Fabien-Chouteau:
How will we detect in the index if there are clashing crates (e.g. `Hello` and `hello`)?

mosteo:
At loading time the first one takes precedence (same mechanism as earlier indexes releases take precedence). But we might want to detect and report this as an error/warning?

Fabien-Chouteau:
> we might want to detect and report this as an error/warning?

At this point I would say that enforcing a lowercase name for all crates is a good solution.

Issue #175, opened 2019-09-09
[RFC] Better granularity in index for reuse of information


mosteo:
Context: in the old index you could use a release as the base for a new one ([example](https://github.com/alire-project/old-index/blob/master/oldindex/alire-index-adayaml.ads)), either by replacing all or adding new dependencies/properties. This, to some extent, is doable in the new index because the `[general]` sections allows to specify properties/dependencies that apply to all releases in the crate.

I see a few shortcomings with the current status:
* If you specify some dependencies/properties in `[general]`, these will apply forever to all releases; so, in turn:
* You cannot in practice modify the [general] section once you have a few releases, since that would apply to past releases, or
* If you do want to modify the general section, you'll have to move its defaults to all existing releases, and start afresh.
* It seems unlikely to me that, in long-lived crates, with possibly several major point releases, `[general]` deps/props can be reused; in practice every release will have to list all its information, even if there is some extent of commonality.

TL;DR: some way of having more precise inheritance seems likely to be needed. I see a couple obvious possibilities, haven't given too much thinking to it, waiting for comments:
1. Be able to refer to an existing release, as in the past, to add new info incrementally.
1. Have some kind of "named sections" which are not releases to which a release can refer as a base; thus several releases may reuse a common section. Once breaking changes appear, a new common section is created as baseline, without the need to change anything anywhere.

And as how to express that, I can see:
1. Using explicit names and a way of referring to them (`inherits-from = "whatever"` or the like)
1. Have an extra nesting level in which several crates can be grouped within a common section. Trouble is that TOML is somewhat cumbersome for nested structures, but that cat is already out of the bag. If we where using YAML I would go for this one but with TOML I feel 1 may be clearer.

If some solution were adopted, I would suggest to reserve the `[general]` section for truly unique values (authors, maintainers) that once changed really apply to all crates, but not for deps/props.

Fabien-Chouteau:
How is this done in other package managers?

mosteo:
Rust: not done, apparently: https://github.com/rust-lang/crates.io-index/blob/master/ra/nd/rand

OPAM: not done, and besides they have one file per release: https://github.com/ocaml/opam-repository/tree/master/packages/mirage-random

I guess that if the ultimate goal is to only edit those files through the tooling it doesn't make sense to make them amenable to user edition (or even reading, as in the Rust case).

PR #200, opened 2019-09-17
Re-enable child crates feature

### LINKED ISSUE #108: Parent-child projects
The old index used child packages as a way to denote a more intimate relation between projects. An example is the `adayaml` and `adayaml.server`, and `simple_components.*`.

***Edit***
I'm currently edging towards using parent-child.toml crate file names, a la gnat sources, for parent.child crates. Nothing major changes in regard to the status quo. If this were to happen:

[ ] Reindex `Simple_Components` as a collection of child packages.

***Original discussion follows:***

The main difference and use case for these child projects was that, if they shared the same commit, they weren't checked out as a stand-alone project, but reused the parent checkout. In turn, this required that their corresponding project files were not conflicting (e.g. reusing source files).

In other words, it was a way to split several related components, each one with their own GPR file, in different alire subprojects, without checking out a copy for each one.

Advantages were:
- Possibility of having different dependencies per child
- Reuse of the checkout
- Greater visibility of related components in a large project
- When features were added in later releases, they allowed the previous ones to stand alone (if the project maintainer provided new GPR files for new features, naturally).
- They could still be declared in a single file that contained all child projects (although separate child package files were in the roadmap).

Right now, these kind of projects (e.g., `adayaml` and `adayaml_server`) still work but they behave as totally independent projects: they are checked out separately, even being the same commit from the same repository, and the parent-child info does not make into `alr`.

I can see several approaches about this:
1. Do nothing, and lose these features, if we deem them not important enough.
1. Include a new field in the [general] section to make explicit the relationship: `parent = "adayaml"`, for example, and keep declarations in separate toml files.
1. Something more fancy that allows the declaration of child projects inside a same toml file

Without much thought given, option 2 seems low effort and enough to recover the most interesting advantages.

On a related note, I used the '.' separator because it's more Ada-like. I wonder if we may want to recover that, even if we do nothing else. That leads me to:

4. Use '.' as a especial character in project names that implies the parent-child relationship, and dispense with the `parent = "xxx"` field to achieve the same result as 2.

I'm liking 4 more right now.

@pmderodat, another one ;-)

mosteo:
Fixes #108.

With this PR, we allow "child crates" with `parent-child.toml` on-disk filename and `parent.child` in-memory crate name.

To keep things simple and explicit, there is no inheritance between parents and children, so there are no changes in index syntax. Lexicographical loading order is enforced to ensure that a parent crate is loaded before its children, although this isn't used for anything at the moment. This could be revisited if at some point we work on extension/overriding mechanisms (#175).

One of the reasons why this feature was introduced originally was to split checkouts into several crates for different project files and dependencies. An obvious offender is the Simple Components; some of its project files have no dependencies while others bring in databases and crypto libraries. For that reason, there is an expectation that same-version sibling crates come from the same repository and commit, although this is not enforced, nor should make any difference.

The side-effect of reusing the same checked-out folder is still there, although we could easily remove that if not wanted.

There is a corresponding PR to alire-index that renames what formerly where child crates, and re-index the Simple Components which went missing during index migration and mostly stressed this feature.


Fabien-Chouteau:
This mechanism and the potential consequences (extension/overriding) seem very complex.

What are the benefits of having child crates over a classic explicit dependency?

mosteo:
> over a classic explicit dependency?

Dependencies are still explicit. In the current form, the mechanism is mostly cosmetic. Three scenarios to show the subtleties:

- Scenario A, these crates exist:
  - smart_pointers
  - simple_components

- Scenario B:
  - smart_pointers
  - simple_components
  - simple_components_odbc

- Scenario C:
  - smart_pointers
  - simple_components
  - simple_components.odbc

In scenario A, `simple_components` provides two gpr files, one that has no dependencies and another one that depends on some other libraries. The crate as a whole needs to bring in these dependencies.

In scenario B (pre-PR), `simple_components` has no dependencies and `simple_components_odbc` depends on ODBC stuff. If a crate depends on both `simple_components` and `simple_components_odbc`, the same repository commit is checked out twice.

In scenario C (post-PR) a project using both `simple_component` and `.odbc` would reuse the checkout. (Also, depending on `.odbc` does not implicitly depends on `simple_components` -- unless we want to.)

All in all, it's a way of distinguishing `'_'` as an space and `'_'`/`'.'` as a related crate. 

Actually I'm now wondering what would happen in scenario B, since there will be duplicate GPR and Ada files in scope of the aggregate build project. I'll check this.

mosteo:
> Actually I'm now wondering what would happen in scenario B, since there will be duplicate GPR and Ada files in scope of the aggregate build project. I'll check this.

I just checked and `gprbuild` just ignores one of the two folders (is this normal?). I can even delete the one that's being used by `gprbuild` (checked with `-vm`) and on the next try it just sees and compiles the other. 

I feel we are in thin ice here: if, say, two such crates that share sources are resolved to different commits, there will be differences in source contents and one set of them will be ignored.

All in all, after this test I don't feel confortable having these split crates as things stand now. I would in the short term merge the current split crates and more calmly think of something more foolproof if we see a clear advantage in having it (I know that Ravenports sells something like this (flavors/subpackages) as a big deal.)

Fabien-Chouteau:
As far as I understand the main advantage is not doing multiple a checkouts of the same repo.

I would say that this too little advantage compared to the complexity that it will bring to the code.
And as you mentioned it will not work if the commits are different.

Projects like those big libraries (simple-components, gnatcoll) exists because there is no package manager for Ada (otherwise they would be split already), so adding a complex feature for that use case seems wrong.

> I would in the short term merge the current split crates and more calmly think of something more foolproof if we see a clear advantage in having it

I would avoid bringing back a feature if we are not 100% sure that we are going to keep it in the future.

mosteo:
> As far as I understand the main advantage is not doing multiple a checkouts of the same repo.

No, that was a collateral. The target use case is to be able to selectively use parts of a checkout that have different dependencies. I know I haven't explained this well in the past because it wasn't clear either in my head.

> Projects like those big libraries (simple-components, gnatcoll) exists because there is no package manager for Ada (otherwise they would be split already), 

I'm not 100% sure about that. With that premise, several GPR files with different dependencies in turn require starting a new repository for each one. I see source code management and dependency management as something orthogonal to some extent.

> so adding a complex feature for that use case seems wrong.

I think that this use case is valid (which is not motivated by saving space, but by avoiding unnecessary dependencies). Perhaps it's not urgent, but it may arise as a valid need, and at that time we can try to make it not complex in the context of the rest of alire.

> I would avoid bringing back a feature if we are not 100% sure that we are going to keep it in the future.

In this I agree because what we have is fundamentally broken. I will merge crates with duplicate sources for the time being.

Issue #180, opened 2019-09-12
Add requirement for lsb_release to docs


Lucretia:
On Gentoo Linux, ```sys-apps/lsb-release``` is not installed by default, this results in alr issuing "undetected distribution" warnings. 

```
emerge -av sys-apps/lsb-release
```

mosteo:
`lsb_release` is no longer required after #204

Issue #177, opened 2019-09-12
ERROR: alr get unsuccessful


Lucretia:
Just installed alire to try it out, any attempt to ```alr get hello``` or any other package results in the above error.

mosteo:
These issues were fixed by #204

Issue #169, opened 2019-08-20
Do not raise Program_Error on unsupported package manager


mosteo:
Report native packages as unavailable instead.

This was known per #65, but I got a new report by email about the same problem. Since #65 as evolved into a discussion about how to support native packages in general, I create this one to track the narrow simpler fix.

mosteo:
Fixed by #204

Issue #65, opened 2018-10-17
Handling of native packages


pmderodat:
As of today, the handling of native packages only works on Debian or Ubuntu distributions. For instance, on my Arch Linux box, I’m getting a lot of crashes such as:
```
$ bin/alr search --full --list
Searching... .                       
raised PROGRAM_ERROR : Unsupported package manager
[/home/pmderodat/_git/alr/bin/alr]
Alr.Origins.New_Origin at alr-origins.adb:45
Alr.Query.Is_Available.Localalias at alr-query.adb:119
Alr.Query.Resolve.Expand.Expand_Value.Check at alr-query.adb:326
Alr.Query.Resolve.Expand.Expand_Value at alr-query.adb:353
Alr.Query.Resolve.Expand at alr-query.adb:408
Alr.Query.Resolve at alr-query.adb:428
Alr.Query.Is_Resolvable at alr-query.adb:126
Alr.Commands.Search.Execute.List_Release at alr-commands-search.adb:49
Alr.Commands.Search.Execute at alr-commands-search.adb:96
Alr.Commands.Execute_By_Name at alr-commands.adb:555
Alr.Commands.Execute at alr-commands.adb:533
Alr.Main at alr-main.adb:24
Main at b__alr-main.adb:1636
[/usr/lib/libc.so.6]
0x7f90deee4221
[/home/pmderodat/_git/alr/bin/alr]
0x404ecc _start at ???
0xfffffffffffffffe
```
I don’t think that requiring users to go through all packages in the index to complete package names each time they use a new distributions is a good position. Besides, the current formalism will not apply on package manager-less systems such as Windows. And this forces users to install software through their package managers, while some may want to do manual install from sources (to have recent versions, for instance).

For these reasons, I think the current mechanism to handle native package dependencies should be replaced. What about instead: assume the dependence is available then and check it is available only when trying to install a package that depends on it, or when looking for packages that are available?

Instead of using a never-exhaustive database of distribution-specific package names (which can vary over time, by the way!), maybe we could use the same trick as configure scripts: try to build a very simple program using the dependency, demonstrating that it is installed (no matter how). We could still have, as an optional feature, alr using the incomplete database to suggest which packages to install given their distribution, though…

What do you think?

mosteo:
Adding two links for future reference:

https://github.com/rust-lang/cargo/issues/1240
Brief discussion about the same problem in cargo.

https://crates.io/crates/libudev
[...]

PR #204, opened 2019-09-19
Fix errors about native packages in unknown platforms.

### LINKED ISSUE #181: Don't depend on OS specific tools
I'm on Gentoo, I don't have apt-cache:

```
$ alr show libsdl2
libsdl2=2.0.0: Simple DirectMedia Layer development files
Origin (native package):
Available when: True
Properties:
   Maintainer: alejandro@mosteo.com
ERROR: apt-cache policy exited with code: 1
ERROR: alr encountered an unexpected error, re-run with -d for details.
```

mosteo:
As part of the effort to better deal with native packages, this is a minimal patch to avoid general exceptions in unsupported distributions, that fixes #181, #180, #177, #169 (in the sense that clearer errors are emitted instead of the general internal failure one).

My plan is to rework the native situation as seamlessly as possible pre/post beta along these lines:

- Do not assume the platform is known (this patch) so people don't get bizarre errors.
- Allow resolution/retrieval of crates that are only missing native packages (the "permissive mode" we talked in #65).
  - In that case (missing native dependencies), use current native crates to give the user a hint of what they should provide manually (and to warn that the build will likely fail otherwise).

That would be my minimum target for the beta, so pure Ada packages work everywhere and ones with native dependencies can be also retrieved/tested everywhere. For the medium/long term, I'd like to:

- Remove current native crates, which have some fundamental problems.
  - At that time (or before) introduce a simple way of tracking missing external dependencies so we can continue operating with the hinting system.
- Introduce a new native description system that can accommodate all the issues identified with the current one, minimizing maintenance burden and making support from the community more amenable. I will prepare a draft proposal about this that we can discuss when the time comes.

### DIFF
--- a/src/alire/alire-origins-deployers.adb
+++ b/src/alire/alire-origins-deployers.adb
@@ -92,10 +92,16 @@ package body Alire.Origins.Deployers is
    -- Install_Native --
    --------------------
 
-   function Install_Native (From : Origin) return Outcome is
+   function Install_Native (Release : Releases.Release) return Outcome is
+      From : constant Origin         := Release.Origin;
       Orig : constant Deployer'Class := New_Deployer (From);
    begin
-      if Orig.Already_Installed then
+      if From.All_Native_Names (Platform.Distribution) = Origins.Unavailable
+      then
+         return Outcome_Failure
+           ("No native package known in current platform for "
+            & Release.Milestone.Image);
+      elsif Orig.Already_Installed then
          Trace.Detail (From.Package_Name (Platform.Distribution) &
                          " already installed");
          return Outcome_Success;
@@ -114,10 +120,19 @@ package body Alire.Origins.Deployers is
    -- Deploy --
    ------------
 
-   function Deploy (From : Origin; Folder : String := "") return Outcome is
+   function Deploy (Release : Releases.Release;
+                    Folder  : String := "") return Outcome
+   is
+      From : constant Origin := Release.Origin;
    begin
       if From.Is_Native then
-         return Install_Native (From);
+         if Platform.Distribution_Is_Known then
+            return Install_Native (Release);
+         else
+            return Outcome_Failure
+              ("Unknown distribution: cannot provide native package for "
+               & Release.Milestone.Image);
+         end if;
       else
          return Deploy_Not_Native (From, Folder);
       end if;
--- a/src/alr/alr-bootstrap.adb
+++ b/src/alr/alr-bootstrap.adb
@@ -19,15 +19,16 @@ package body Alr.Bootstrap is
    ---------------------
 
    procedure Check_Ada_Tools is
-      --  FIXME mini-leak (once per run)
+      procedure Check_Tool (Exec : String) is
+      begin
+         if not OS_Lib.Exists_In_Path (Exec) then
+            Trace.Error ("Required tool not detected: " & Exec);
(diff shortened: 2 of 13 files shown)


src/alire/alire-origins-deployers.adb
```diff
@@ -117,7 +117,12 @@ package body Alire.Origins.Deployers is
    function Deploy (From : Origin; Folder : String := "") return Outcome is
    begin
       if From.Is_Native then
-         return Install_Native (From);
+         if Platform.Distribution_Is_Known then
+            return Install_Native (From);
+         else
+            return Outcome_Failure
+              ("Native packages are unavailable in unknown distributions");
```
Fabien-Chouteau: What about this error message: `Cannot provide native package for unknown distribution`?
I think it explain better that it is because the distribution is unknown that we cannot provide the package.
Or: `Unknown distribution: Cannot provide native package`

Also print some kind of name for the native dependency.

src/alr/alr-bootstrap.adb
```diff
@@ -19,15 +19,17 @@ package body Alr.Bootstrap is
...
    begin
-      if not OS_Lib.Exists_In_Path ("gprbuild")
-        or else
-         not OS_Lib.Exists_In_Path ("gnatmake")
-      then
-         Trace.Error ("Ada tools not detected, alr cannot proceed");
-         OS_Lib.Bailout (1);
-      end if;
+      Check_Tool ("gprbuild");
+      Check_Tool ("gnatmake");
```
Fabien-Chouteau: Is `gnatmake` absolutely required?

We don't use it for embedded.

src/alire/alire-origins-deployers.adb
mosteo: Yup, better.

src/alr/alr-bootstrap.adb
mosteo: Ummm you got me there, I have little experience in embedded. I just presumed that it was always needed for binding. I will leave only gprbuild.

mosteo:
> I think it is worth having a test for this. We can run it on Windows because there isno native package manager.

I was just preparing a docker for CentOS as a stand-in for currently unsupported managers. I have manually tested that the testsuite passes. 

I can start preparing the Windows CI box, but that will be more involved; I expect at least some issues with paths.

Fabien-Chouteau:
> I was just preparing a docker for CentOS as a stand-in for currently unsupported managers. I have manually tested that the testsuite passes.

I mean a test specifically to check that we provide the right error message on unsupported managers.

mosteo:
> > I was just preparing a docker for CentOS as a stand-in for currently unsupported managers. I have manually tested that the testsuite passes.
> 
> I mean a test specifically to check that we provide the right error message on unsupported managers.

OK, understood.

src/alire/alire-origins-deployers.adb
mosteo: Fixed in 3f38dfc

src/alr/alr-bootstrap.adb
mosteo: Fixed in 3fec04c

mosteo:
Test implemented in e6100ea

mosteo:
Fixed the test header in 5446f16

Issue #209, opened 2019-09-26
Add in-alr way of computing integrity hashes for releases


mosteo:
At present hashes must be computed by hand with `git archive | shaXsum`, but since this is hidden knowledge for users, a way of computing a hash for a repository commit from `alr` should be provided. Furthermore, this probably should be easily integrated in the workflow for submitting a new crate release (both `cargo` and `opam` have a `publish` command).

Fabien-Chouteau:
I agree.

I see that as a command that takes a valid origin (online archive, git repo) and produce the hash.

`alr checksum https://github.com/alire-project/alire@master`

mosteo:
This is available after #215.

Issue #208, opened 2019-09-26
--trust switch to enforce integrity levels


mosteo:
Right now hashes are mandatory for source archives and are used when available for git sources, but not required. Since giving warnings would pollute most releases, I suggest the following mechanism (or other that we discuss here) to select desired behavior.

Add a `--trust={off,warn,full}` switch that, on retrieval of a release, works as follows:
- off: no warnings for VCS origins without hashes (current status)
- warn: emit a warning for VCS origins without hashes (might become the default if at some point most crates have hashes).
- full: fail for any release without integrity hashes

Fabien-Chouteau:
I think hash of the source is the bare minimum that we should accept. So I don't like the idea of disabling the checks.

mosteo:
Then we don't need a switch at all, once we fill in the hashes for all existing crates, and require it for all new submissions. It's certainly the safest approach.

Fabien-Chouteau:
> Then we don't need a switch at all, once we fill in the hashes for all existing crates, and require it for all new submissions. It's certainly the safest approach.

Yes, and we can automatically check the hash field on pull-requests on the index repo.

mosteo:
Closing since we don't need it.

Issue #198, opened 2019-09-17
Integrity field for downloaded repositories


mosteo:
See last conclusions in #66. The idea is to use `git archive | sha512sum` or equivalent.

Open questions:
- [x] What happens with ends of line in Linux vs Windows?
  - Verified that different hashes are obtained with the default git configuration
    - Can be overriden with `git -c core.autocrlf=false`
- [x] Implement for git
  - [ ] Study using https://stackoverflow.com/questions/10118302/how-to-fix-the-line-endings-in-checked-out-files-after-changing-the-autoclrf-set
- [ ] Implement for Mercurial (`hg --config extensions.eol=! checkout -C`)
- [ ] Implement for SVN (using `.svn/pristine`)

mosteo:
I'm sorry for being this obtuse, but after a night's sleep:
- If the index repo is trusted --> VCS will check crate integrity for us
- If the index repo is untrusted --> What is the point in checking integrity if the hash in index is already suspect.

By this reasoning, the whitelist is the only thing that is useful.

Fabien-Chouteau:
Here's a scenario I have in mind: someone hacks a GitHub account and changes the content of a repository used in a crate. If we don't have a hash of the content, no one will be able to detect that the sources have been tempered with. In this case the whitelist is not enough.

Another scenario: I write a crate and host it in a git repo on my private server with my domain name.
After a couple year, I forget about this project, I forget about the server and domain name. Someone can hack this server, or maybe acquires the domain name if I stop paying for it. This someone is now able to replace the code in the repo or provide another repo with different sources.
Again, no one will be able to detect that the sources have been tempered with.

The hash is here to guaranty that the sources that have been checked out are really the ones specified by the author of the crate.

Which makes me also think that it should be forbidden to change the hash of a version already published in a crate.

mosteo:
Thanks for these examples. They help me focus the argument.

> Here's a scenario I have in mind: someone hacks a GitHub account and changes the content of a repository used in a crate. If we don't have a hash of the content, no one will be able to detect that the sources have been tempered with. In this case the whitelist is not enough.

It's my understanding that the commit hash we use as origin is enough to detect this during cloning.

> Another scenario: I write a crate and host it in a git repo on my private server with my domain name.
> After a couple year, I forget about this project, I forget about the server and domain name. Someone can hack this server, or maybe acquires the domain name if I stop paying for it. This someone is now able to replace the code in the repo or provide another repo with different sources.
[...]

PR #210, opened 2019-09-26
Generalize integrity checking and implement it for git origins

mosteo:
This PR generalizes the `archive-hash` into `origin-hashes` and implements integrity checking for git origins. Functional changes are:

- Use `origin-hashes`, that applies to all origins, instead of `archive-hash`, which only applied to archive sources.
- Use a vector instead of a single value to allow supplying more than one hash (and verify with all that are given for a release).
- Use the `git archive | shaXsum` trick for platform-independent verification of git remotes.
- Use coreutils shaXsum (since we rely on it for git) also for source archives.

Unless mistaken, this covers the basic integrity verification that we wanted for the beta. I added hg/svn to #198 to track their progress. I also created #208 and #209 to track related improvements.


doc/catalog-format-spec.rst
```diff
@@ -358,9 +358,9 @@ following entries:
    [origin.'case(distribution)']
    'debian|ubuntu' = "native:make"
 
-* ``archive-hash``: mandatory string for source archives. A "kind:digest" field
-  that specifies a hash kind and its value. The only accepted kind is SHA512 at
-  this time.
+* ``origin-hashes``: mandatory string for source archives, optional for git
```
Fabien-Chouteau: I don't think it should be optional for git.

doc/catalog-format-spec.rst
```diff
@@ -358,9 +358,9 @@ following entries:
...
+  origins. An array of "kind:digest" fields that specify a hash kind and its
+  value. Kinds accepted are sha256, sha384, sha512.
```
Fabien-Chouteau: Why accept multiple kinds? Let's take the most secure.

src/alire/alire-hashes-compute.adb
```diff
@@ -0,0 +1,41 @@
...
+   is
+      --  Determine the coreutils tool to use for hashing
+      Hasher : constant String := Utils.To_Lower_Case (Kind'Img) & "sum";
+      --  E.g.; sha512sum
+
+      Output    : Utils.String_Vector;
+      Exit_Code : Integer;
+   begin
+      OS_Lib.Subprocess.Raw_Spawn
+        (Program    => "sh",
```
Fabien-Chouteau: It would be better to use the hash functions provided in libGNAT. There would be less dependency on external tools.

src/alire/alire-origins-deployers-git.adb
```diff
@@ -1,15 +1,68 @@
...
+
+      --  Determine the coreutils tool to use for hashing
+      Hasher : constant String := Utils.To_Lower_Case (Kind'Img) & "sum";
+      --  E.g.; sha512sum
+
+      --  Let git do its magic
+      Output    : Utils.String_Vector;
+      Exit_Code : Integer;
+   begin
+      Raw_Spawn (Program    => "sh",
```
Fabien-Chouteau: Why use `sh` here? Is it possible to spawn `git` directly?

src/alire/alire-origins-deployers-git.adb
```diff
@@ -1,15 +1,68 @@
...
+                 Arguments  => Utils
+                   .Empty_Vector
+                   .Append ("-c")
+                   .Append ("git -c core.autocrlf=false archive HEAD | "
+                            & Hasher & " | cut -f1 -d\ "),
```
Fabien-Chouteau: Like above, we should use libGNAT to produce the hashes.

src/alire/alire-origins-deployers-source_archive.adb
```diff
@@ -64,4 +48,18 @@ package body Alire.Origins.Deployers.Source_Archive is
...
+   -------------------
+
+   overriding
+   function Compute_Hash (This   : Deployer;
+                          Folder : String;
+                          Kind   : Hashes.Kinds) return Hashes.Any_Digest is
+      Archive_Name : constant String := This.Base.Archive_Name;
+      Archive_File : constant String := Dirs.Compose (Folder, Archive_Name);
+   begin
+      return Hashes.Digest (Hashes.Compute.Hash_File (Kind, Archive_File));
```
Fabien-Chouteau: Just to be sure, is this doing a hash of the archive before it is extracted?

src/alire/alire-origins-deployers.ads
```diff
@@ -63,6 +63,16 @@ package Alire.Origins.Deployers is
    --  case, it is polite not to uninstall anything installed in the user
    --  system.
 
+   function Compute_Hash (This   : Deployer;
+                          Folder : String;
+                          Kind   : Hashes.Kinds) return Hashes.Any_Digest with
+     Pre'Class =>
+       Supports_Hashing (This.Base.Kind) or else raise Program_Error;
```
Fabien-Chouteau: I don't thin you have to raise a Program_Error here, if the precondition is false an exception will be raised anyway.

doc/catalog-format-spec.rst
mosteo: In accordance with the comment in #210. OK.

doc/catalog-format-spec.rst
mosteo: So only sha512 it is.

src/alire/alire-hashes-compute.adb
mosteo: I thought we wanted this for simplicity. Will revert the change.

src/alire/alire-origins-deployers-git.adb
mosteo: I don't know a way to pipe stdout of a subprocess to stdin with GNAT.Os_Lib, but I might have missed it. Do you know if GNAT has portable support for some kind of pipes?

Alternatively, we can use a temporary file.

src/alire/alire-origins-deployers-source_archive.adb
mosteo: Yes

src/alire/alire-origins-deployers.ads
mosteo: It was to ensure it is a Program_Error and not an assertion failed.

doc/catalog-format-spec.rst
mosteo: I have made it a warning for now, with the idea of turning it into an error once the current index is completely updated.

doc/catalog-format-spec.rst
mosteo: Fixed in latest push

src/alire/alire-hashes-compute.adb
mosteo: Fixed in latest push.

src/alire/alire-origins-deployers-git.adb
mosteo: Done with temporary file (commit 7f10d7c).

src/alire/alire-origins-deployers-git.adb
mosteo: Reverted.

src/alire/alire-origins-deployers.adb
```diff
@@ -145,4 +159,55 @@ package body Alire.Origins.Deployers is
...
+            Trace.Warning ("No integrity hashes provided for "
+                           & This.Base.Image);
+            --  TODO: make this an error once all crates are updated with
+            --  their hashes.
+         else
+            Trace.Detail ("Verifying integrity...");
+         end if;
+
+         --  Compute hashes from downloaded release and verify:
+         for Index_Hash of This.Base.Data.Hashes loop
```
Fabien-Chouteau: So if multiple hashes are provided, all of them must be correct.

mosteo: Yes. We only support one now, but I think it may come in handy for future-proofing.

PR #212, opened 2019-10-04
Hash directory contents for reproducible hashes

mosteo:
Thinking a bit more about #210 I was uneasy about using `git archive` for an unintended purpose, so I asked about this in SO: https://stackoverflow.com/questions/58130300. Given the reply, I think we need something totally under our control so we don't find suddenly that a new git/tar version has invalidated all of our hashes.

This patch proposes a solution that should work for all the VCS origins we currently support. We check out a copy with server-side EOLs (with the same git trick we were using). We then recursively hash the folder contents, using utf8 names to avoid FS encoding issues.

The details of how hashing is performed are documented in alire-hashes.ads:37. I believe I've removed any possibility of ambiguity in the information that's being hashed, but since this is a critical point, I hope you can take a particularly careful look at the implementation, function `Hash_Directory`, in alire-hashes-common.adb:23. Both are in the first commit of the PR.

The test added in the last commit may also serve to see the sequence of information that's being hashed.

I've verified that utf8 names returned by GNATCOLL.VFS are identical in linux/ext4 (where utf-8 encoding is the norm) and in Windows 10/NTFS (where utf-16 is used AFAIK).


mosteo:
Also pinging @senier for feedback since he was in #198

senier:
I took a brief look at the recursive hashing. What makes me feel a bit uneasy is the assumption that everything that every `x` with the property `not Is_Regular_File(x)` we assume that `x` is a directory. Is that really the case? At least git can handle symlinks. What happens if we encounter one - are we just ignoring it or treating it as a directory?

src/alire/alire-hashes.ads
```diff
@@ -30,6 +30,31 @@ package Alire.Hashes with Preelaborate is
...
+   --  Deterministic recursive hashing of a folder contents. The Except
+   --  folder, if given, is not included in the hashing nor recursed into. The
+   --  deterministic procedure followed is:
+   --  * All operations use utf8-encoded filenames.
+   --  * Enumerate current folder contents, and sort them alphabetically.
+   --    (Note that ordinary String comparison is used, so not utf8-order but
+   --    plain Character order is used.)
+   --    * ".", "..", Except entries are omitted from hashing.
+   --  * For each dir entry hash the following:
+   --    * UTF8-encoded name, with full route from "root" (Path).
```
Fabien-Chouteau: How is this handling Windows paths root?

testsuite/tests/publish/hash-internals/test.py
```diff
@@ -0,0 +1,79 @@
...
+
+from drivers.alr import run_alr
+from drivers.asserts import assert_match
+
+import re
+
+# Get the release and check hashing log messages
+p = run_alr('get', '-vv', 'libfoo',
+            complain_on_error=True, quiet=False)
+assert_match(
```
Fabien-Chouteau: It seems a bit dangerous to base the test on debug output.

testsuite/fixtures/crates/libfoo_git/config
```diff
@@ -0,0 +1,6 @@
+[core]
+	repositoryformatversion = 0
+	filemode = true
+	bare = true
+[remote "origin"]
+	url = /home/jano/prog/alire/testsuite/fixtures/crates/libfoo/
```
Fabien-Chouteau: How is this setup specific path will be used?

src/alr/alr-commands-dev.adb
```diff
@@ -89,6 +106,11 @@ package body Alr.Commands.Dev is
                      "", "--filter",
                      "Used by scope filtering test");
 
+      Define_Switch (Config,
+                     Cmd.Hash'Access,
+                     "", "--hash",
```
Fabien-Chouteau: This is going to be very useful.

testsuite/tests/publish/hash-internals/test.py
```diff
@@ -0,0 +1,79 @@
+# -*- coding: utf-8 -*-
```
Fabien-Chouteau: I think it would be interesting to test the `alr dev --hash` here as well.

We can test using a tar hosted somewhere and a fake github repo where we put all the corner cases such as symlinks.

Fabien-Chouteau:
>  At least git can handle symlinks. 

How are they handled on Windows?

> What happens if we encounter one - are we just ignoring it or treating it as a directory?

I would say don't follow the links.

testsuite/tests/publish/hash-internals/test.py
pmderodat: I haven’t looked at the testcase, but generally: IMO it’s better to avoid relying on debug output as much as possible. That being said, when it’s not possible, it’s better to rely on debug output to write a test compared to not writing a test at all. :-) Worst case is that the debug output changes and the testcase needs to be updated, creating maintainance burden. To arbitrate with the importance of the test…

pmderodat:
> > What happens if we encounter one - are we just ignoring it or treating it as a directory?
> 
> I would say don't follow the links.

The symbolic link can affect the compilation process, so I think it’s important to make sure the hash covers it (so that messing with the symlink will invalidate the hash). Maybe we should include the name of the symlink + the path associated to it (the symlink target) in the hash computation?

testsuite/fixtures/crates/libfoo_git/config
mosteo: Uuups. I did a clone of a test repo so that origin got stored but should not have been checked in. The libfoo_git is itself a bare repository that is cloned, so its origins aren't really used. I will clean this up.

senier:
> The symbolic link can affect the compilation process, so I think it’s important to make sure the hash covers it (so that messing with the symlink will invalidate the hash). Maybe we should include the name of the symlink + the path associated to it (the symlink target) in the hash computation?

Fully agreed! We should not ignore anything we come across (TBH, I've no idea what else could show up when checking out a repo). All object types should get their own [marker](https://github.com/alire-project/alire/blob/feat/hash/src/alire/alire-hashes-common.adb#L92) to make sure we do not end up with a hash collision for different kind of objects (with the same name). IMHO it would also be good to raise an exception for file types we do not expect to be present in the hash function.

testsuite/tests/publish/hash-internals/test.py
mosteo: The constraint is that we wanted to avoid dependencies on external resources for the test suite. Not sure how hard this constraint is, since we need to check out alr subrepos anyway (although all of them come from github).

src/alire/alire-hashes.ads
mosteo: If I understand what you are asking, the fake root is our check out folder so it should not include anything platform-specific. E.g., if the checkout is of ```C:\Workspace\Crate```, then the hashing starts with ```/``` that maps to that folder.

senier:
> How are they handled on Windows?

[Looks like](https://github.com/git-for-windows/git/issues/117) they are converted to a regular (text) file containing the link. This will be a problem unless we can tell them apart from other regular files on Windows.

If symlinks cannot be distinguished from regular files on Windows, we will need to hash them as regular files (using the same marker) and implement a compatible hashing on Linux as @pmderodat suggested. In this case, an attacker could replace a symlink with a regular file containing the symlink target as text. Not sure what the implications are.

mosteo:
> I took a brief look at the recursive hashing. What makes me feel a bit uneasy is the assumption that everything that every `x` with the property `not Is_Regular_File(x)` we assume that `x` is a directory. Is that really the case? At least git can handle symlinks. What happens if we encounter one - are we just ignoring it or treating it as a directory?

Right now anything not a regular file or directory causes an error. Symlink omission was an oversight on my part, but at the time I would have thought that windows indeed supported symlinks! All the issues you have raised however are not trivial...

> How are they handled on Windows?

Summarizing the info all of you provide: with admin rights, they /might/ work similarly. We would depend on git support (seems to be there -- but what about other VCSs?) and on GNATCOLL.VFS support (can we cross-platform-ly retrieve symlink targets?)

There is this comment in my version of GNATCOLL (might be obsolete):
```
   procedure Symbolic_Links_Support (Active : Boolean);
   --  Whether this package should do extra system calls to handle symbolic
   --  links.
   --  This is automatically False on platforms like Windows where this notion
   --  does not exist,
```
Frankly, at this point, dealing with all the cross-platform kirks x VCS kirks seems something that we should not be investing too much time in. So I'm contemplating a couple of radical ways out:

1. Don't allow anything not a regular file/folder
   1. If windows symlink support is problematic anyway, why allow them? Even unzipping an archive that contains them there will likewise be problematic?
   1. Are there known instances of symlinks in source packages in some lib we know? (As opposed to using them during the build process at some point) I guess this is very hard to ascertain.
1. Abandon support for VCSs and rely only on distribution through tar/zip files.
   1. Hashing files is easy, but does this gains us anything for symlinks on windows? (See above).
1. Support only git and forget about hg/svn, and deal with only one cross-platform/VCS combo.
   1. There is one SVN crate and no Hg crates at present in the catalog.
   1. The issue of windows support for symlinks in GNATCOLL remains.
1. Support symlinks on linux only, presuming windows-portable libs aren't going to use them for the reasons above, so no problem about cross-platform consistent hashing.
   1. Too tempting/easy to be true?

Right now I would go for the last option, but I need some distance from this subject right now (wasn't expecting this can of worms). Your further thoughts welcome!

mosteo:
An update with some minimal tests:

* Zipped files in linux with symlinks (`-y` option) become plain files that contain their path when unzipped in windows. By default (without `-y`), zip will follow links and store these as regular files.
* `git with -c core.symlinks=false` (the default) will do the same upon clone from windows (create a plain file with the path inside).
   * Cloning in linux respects the symlinks in the repo.
* `git with -c core.symlinks=true` gives the error `error: unable to create symlink link: Function not implemented`

So I would minimize our pain and just deal with symlinks in linux as @pmderodat and @seiner suggested (kind, name, and relative path), which is VCS agnostic. We don't support cross-platform symlinks, since it doesn't seem to be portable anyway. Anyone doing cross-platform won't be using them, for these reasons. Worst case and they do, they will have to publish using source archives instead of repository commits.

Fabien-Chouteau:
> Frankly, at this point, dealing with all the cross-platform kirks x VCS kirks seems something that we should not be investing too much time in. So I'm contemplating a couple of radical ways out:

I agree with you, this is getting way to complex and error prone.

 - 1. Are we sure we won't find other corner cases like this in the future?
 - 2. is ok in my opinion. Making tar release on github is actually pretty easy. We only have to hash the tarball so I don't see why symlinks would be a problem.
 - 3. And use the commit hash?
 - 4. I don't like the idea of having different behavior on different platorms

testsuite/tests/publish/hash-internals/test.py
mosteo: About the fake github repo, this is what the test does currently (it clones the libfoo_git folder, which is a bare repo).

mosteo:
I see we crossed messages. Tarballs do bypass all the problems, which is a major selling point. It's a major change, though, with some ugly implications.

I've tested the zip archives that github generates. For example, for
```
https://github.com/alire-project/alire/archive/9317b914aebba4398313bf05e34ea12faf498d88.zip
```
you get a file that unpacks as
```
alire-9317b914aebba4398313bf05e34ea12faf498d88/<contents>
```
With our current index syntax, that means prepending project files in the index with that prefix. Alternatively, we can introduce some way of saying that the release contents are all relative to some prefix dir.

Of course that can be avoided by crafting zip files manually without the prefix path, but that would prevent using the tarball service of github and similar sites.

mosteo:
> I was more thinking about the release mechanism: https://github.com/Fabien-Chouteau/hello-github-actions/releases

Same thing happens, although now I think that's only a very minor hurdle.

OTOH, the actual question is to be sure that these tarballs are pre-generated vs on-demand, or we would be back to the original reason for this PR. My understanding is that, at least for releases, they will be pre-generated. The one I posted from an arbitrary commit is likely on-demand. Just to be sure I asked [here](https://github.community/t5/How-to-use-Git-and-GitHub/Ensuring-identical-tarballs-from-releases/m-p/34196/highlight/true#M8997).

Fabien-Chouteau:
> Same thing happens,

Right, I didn't notice the release tag at the end of the directory. Still makes is more manageable than the commit hash.

> although now I think that's only a very minor hurdle.

Because of the  `project-files` property?

e.g.: `project-files = ["asis-gpl-2017-src/asis.gpr"]`

Should we have a `root-dir` property for releases that will be the base for all path related properties?

mosteo:
I think we can safely assume that either

1. the tarball unzips in-place, and `project-files` will be where they are expected (without any prefix).
1. the tarball unzips to a single dir, within which the `project-files` are where they are expected (and the prefix can be deduced).

We can easily check which of the two is happening after unzipping and in case 2 either prepend the prefix internally or simply move everything out in place after unzipping. This way we don't need to add a `root-dir` to be filled which in essence is noise (a side-effect of packaging).

Any other situation would be unsupported and reported as an error.

This should be documented somewhere in the publishing workflow though.

senier:
I sometimes [put my project files into a dedicated subdirectory](https://github.com/Componolit/libsparkcrypto), especially when different variants exist. This would then be unsupported, right? I'd opt for an optional `root-dir` property then. Also, I have the suspicion that people will shoot themselves in the foot with the automatic deduction of the prefix.

mosteo:
If it's unsupported it won't be because of this... As of today, alr requires project files to be in a known-in-advance place (or at least, known per environment case). I need to check the build process of the project you link to.

About the foot-shooting, that's true, but if we semi-automate the publishing process it could be a non-issue for users. We could also fill the `root-dir` at that time ourselves, though, if we really want it to be explicit. To me it's overkill, but I'll go with the majority sentiment.

mosteo:
@senier , I've checked your library. After make I see the following project files:
```
./out/libsparkcrypto/libsparkcrypto.gpr
./build/libsparkcrypto.gpr
./build/build_libsparkcrypto.gpr
./tests/build.gpr
```
The first one is produced by make. I'm assuming that clients of the lib have to use either the first or second one. Since `project-files` in `alr` expects relative paths (not simply basenames), there should be no ambiguity: `project-files = ["out/libsparkcrypto/libsparkcrypto.gpr"]`. 

The problem is different, in that the project file doesn't exist at the time of unzipping, so it can't be used for validation. 

I just checked opam, which uses tarballs, and I see no mention of prefixes or assumptions.

Bottom line may be to just mandate that the tarball must be a single folder with the crate contents (so we forbid in-place tarballs), or allow both and have the prefix.

mosteo:
I'm unsure if we reached a firm consensus, so I'm going to summarize my views on this so we can take action:

1. Allow both tarballs that contain a single root folder and tarballs where files are directly at the top level. Explicitly indicate with a new property in the index what is the root inside the tarball for the former ones.
1. Require that tarballs have always a single folder with the contents. Err during publishing/retrieval if the tarball is malformed. Out of curiosity I've checked github, gitlab and sourceforge, and all of them generate this kind of files.

My vote goes to 2 because this requires only minimal modifications in `alr` and I see no advantage in allowing the flexibility, which imposes a distributed overhead on all packagers of crates for no clear gain or use case.

senier:
I'm also in favor of the simpler option 2.

Fabien-Chouteau:
Let's go with option 2 and we can always make option 1 if necessary in the future.

mosteo:
Then I'm closing this one for the time being, and will test run migrating a few crates from git to archives.

PR #214, opened 2019-10-07
Updated readme to remove obsolete information

mosteo:
- Mention that we are ramping up towards public beta so changes to be expected.
- Hide the installer until we have something minimally stable.
- Updated installation instructions.
- Remove the link to the paper that explains the self-compiling version.


mosteo:
This is quite minimal but I'm pressed for time right now. At least it shouldn't direct people to the highly obsolete self-compiling version.

PR #217, opened 2019-10-15
Add local tarballs as a valid filesystem origin

mosteo:
We already had the possibility to use a local folder as the source for a crate; this use case was intended for testing without requiring a network and to enable developers to use an in-progress local crate as a dependency.

With this PR, a filesystem crate may also point to a file with one of the known tarball formats. On retrieval, the file will be unpacked as if it had been just downloaded. Local tarballs require a hash, unlike local folders.


src/alire/alire-origins-deployers-source_archive.adb
```diff
@@ -61,4 +54,86 @@ package body Alire.Origins.Deployers.Source_Archive is
...
+      -- Check_And_Move_Up --
+      -----------------------
+
+      procedure Check_And_Move_Up is
+         Contents : constant VFS.Virtual_File_Vector :=
+                      VFS.Read_Dir
+                        (VFS.New_Virtual_File (VFS.From_FS (Dst_Dir)));
+         Success  : Boolean;
+      begin
+         if Natural (Contents.Length) /= 1 or else
```
Fabien-Chouteau: So this is doing the heuristic for archive subdir/rootdir?

src/alire/alire-origins.ads
```diff
@@ -39,10 +39,10 @@ package Alire.Origins with Preelaborate is
      );
 
    Supports_Hashing : constant array (Kinds) of Boolean :=
-                        (Git |
+                        (Filesystem |
+                         Git |
```
Fabien-Chouteau: What do we do with git origin? Only allowed in for local crates?

mosteo: Currently it is using the "obsolete" `git archive` way. We should deprecate that, I guess.

mosteo:
It seems this was auto-closed when the base branch was closed, reopening for merge...

PR #228, opened 2019-10-27
Initial work for Windows support

Fabien-Chouteau:
This is enough to have `alr get` and `alr build` working on the `eagle_lander` crate.

### DIFF
--- a/src/alire/alire.ads
+++ b/src/alire/alire.ads
@@ -102,32 +102,35 @@ package Alire with Preelaborate is
           'a' .. 'z' | 'A' .. 'Z' | '0' .. '9' | '_' | Extension_Separator);
    --  Used for cross-platform folder names
 
-   subtype Platform_Independent_Path is String;
-   --  This type is used to ensure that folder separators are externally always
-   --  '/', and internally properly converted to the platform one
+   subtype Any_Path is String;
+   --  Base type for paths in Alire
 
    --  To clarify constants/functions declared herein:
 
-   function Check_Absolute_Path (Path : String) return Boolean;
+   function Check_Absolute_Path (Path : Any_Path) return Boolean;
    --  Return True if the string Path represent an absolute path on the
    --  platform.
 
-   subtype Absolute_File is Platform_Independent_Path
-     with
-       Dynamic_Predicate =>
-         Check_Absolute_Path (Absolute_File)
-          and then
-         Absolute_File (Absolute_File'Last) /= GNAT.OS_Lib.Directory_Separator;
+   subtype Directory_Path is Any_Path;
+
+   subtype File_Path is Any_Path
+     with Dynamic_Predicate =>
+       File_Path (File_Path'Last) /= GNAT.OS_Lib.Directory_Separator;
+
+   subtype Absolute_File is Any_Path
+     with Dynamic_Predicate =>
+       Check_Absolute_Path (Absolute_File)
+        and then
+       Absolute_File (Absolute_File'Last) /= GNAT.OS_Lib.Directory_Separator;
    --  Filenames with full path
 
-   subtype Absolute_Path is Platform_Independent_Path
-     with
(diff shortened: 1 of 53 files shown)


mosteo:
Super nice!

Any ideas on why e3.testsuite isn't found? It seems to install properly.

Fabien-Chouteau:
> Super nice!

Right :)

> Any ideas on why e3.testsuite isn't found? It seems to install properly.

No, at first I thought it was a python2 vs python3 thing, but after a test I don't think so anymore.

Fabien-Chouteau:
It looks like the version of `tar` in the Windows CI setup doesn't want to untar in a directory that doesn't exists:
```
stderr: Command ["tar", "--force-local", "-xf", "D:\a\alire\alire\testsuite\fixtures\crates\libhello_1.0.0.tgz", "-C", "C:\Users\RUNNER~1\AppData\Local\Temp\tmpx_ymxo\get__unpack-in-place\libhello_1.0.0_filesystem"]
exited with code 2 and
output: /usr/bin/tar: C\:\\Users\\RUNNER~1\\AppData\\Local\\Temp\tmpx_ymxo\\get__unpack-in-place\\libhello_1.0.0_filesystem: Cannot open: No such file or directory 
/usr/bin/tar: Error is not recoverable: exiting now
```

@mosteo do we have an API to create directory?

mosteo:
> do we have an API to create directory?

We don't. I've checked in a few places and it is either `Ada.Directories.Create_Directory` or `GNATCOLL.VFS.Make_Dir`. The latter will create intermediate directories.

Fabien-Chouteau:
> `tar` in the Windows CI setup doesn't want to untar in a directory that doesn't exists

Turns out it was not the problem...

Fabien-Chouteau:
Tests have failed, do not merge.

Fabien-Chouteau:
So the solution was to move into the destination directory before running `tar`.

The tests pass but I want to cleanup before this is merged.

Fabien-Chouteau:
@mosteo @pmderodat The PR is ready for review.

I know there is a lot of commits but the changes are relatively limited.
The most impacting is maybe the refactoring of directory and file path types.

Thanks in advance.

src/alire/os_windows/alire-platform.adb
```diff
@@ -0,0 +1,24 @@
+with Alire.OS_Lib;
+
+package body Alire.Platform is
+
+   --  Linux implementation
```
mosteo: Leftover comment

src/alire/alire.ads
```diff
@@ -102,32 +102,35 @@ package Alire with Preelaborate is
           'a' .. 'z' | 'A' .. 'Z' | '0' .. '9' | '_' | Extension_Separator);
    --  Used for cross-platform folder names
 
-   subtype Platform_Independent_Path is String;
-   --  This type is used to ensure that folder separators are externally always
-   --  '/', and internally properly converted to the platform one
+   subtype Base_Path is String;
```
mosteo: Things are much better with your introduced distinctions, thanks! Still, I find `Base_Path` a bit confusing because it evokes `basename` to me, which is entirely a different thing. What do you think about using `Any_Path` instead?

mosteo:
Thanks for the outstanding effort with this one! I see you caught a lot of bugs here and there too. 

I have left only one comment which is in essence a matter of taste, so to me the PR is fundamentally OK.

src/alire/alire.ads
Fabien-Chouteau: I'm ok with `Any_Path`.

src/alire/os_windows/alire-platform.adb
Fabien-Chouteau: Thanks :)

mosteo:
@Fabien-Chouteau , since you told me this is ready, I assume you checked with @pmderodat internally for his pending review. Thanks for your hard work on this one!

PR #229, opened 2019-10-29
Reorganize retrieval/verification steps for origins

mosteo:
This patch adds an extra Fetch step to `Origins.Deployers`. This way, instead of (Fetch & Unpack) -> Verify, we Fetch -> Verify -> Unpack. This is more in line with the workflow for source archives.

### DIFF
--- a/src/alire/alire-origins-deployers.adb
+++ b/src/alire/alire-origins-deployers.adb
@@ -59,13 +59,23 @@ package body Alire.Origins.Deployers is
                                Folder : String) return Outcome
    is
       The_Deployer  : constant Deployer'Class := New_Deployer (From);
-      Deploy_Result : constant Outcome := The_Deployer.Deploy (Folder);
+      Result        : Outcome;
    begin
-      if Deploy_Result.Success then
-         return The_Deployer.Verify_Hashes (Folder);
-      else
-         return Deploy_Result;
+
+      --  1. Fetch sources
+      Result := The_Deployer.Fetch (Folder);
+      if not Result.Success then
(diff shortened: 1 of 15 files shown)


src/alire/alire-origins-deployers.adb
```diff
@@ -59,13 +59,23 @@ package body Alire.Origins.Deployers is
...
-      Deploy_Result : constant Outcome := The_Deployer.Deploy (Folder);
+      Result        : Outcome;
    begin
-      if Deploy_Result.Success then
-         return The_Deployer.Verify_Hashes (Folder);
-      else
-         return Deploy_Result;
+
+      --  1. Fetch sources
+      Result := The_Deployer.Fetch (Folder);
```
Fabien-Chouteau: Shouldn't we fetch and deploy in different directories?

mosteo: Do you see some advantage? Right now the tarball is deleted after unpacking so it doesn't mix with the actual sources.

Fabien-Chouteau: > Right now the tarball is deleted after unpacking so it doesn't mix with the actual sources.

That is the problem I had when trying to switch eagle_lander to archive distribution this weekend :)

So maybe the only argument left will be caching or the archives, not a huge deal at the moment...

mosteo: Yup, I hit that too when migrating the index. BTW, I have eagle_lander unmigrated for you to play with ;-) (in alire-project/alire-index#60)

Issue #191, opened 2019-04-05
Use SPDX license identifiers in the index


pmderodat:
As suggested in https://github.com/alire-project/alire/pull/41#discussion_r264729939 it would be nice to use a stardardize way to refer to existing licenses. Opening this issue to track investigation/developments on this.

mosteo:
Summarizing the discussion in alire-project/alire-index/pull/62: the current implementation checks licenses against the list in [alire-licensing.ads](https://github.com/alire-project/alire/blob/master/src/alire/alire-licensing.ads). This list currently comes from https://github.com/github/choosealicense.com and the SPDX ones are at https://spdx.org/licenses/

The option to explicitly use a custom license is already implemented (unknown licenses are not automatically treated as custom).

I see that neither of the above online lists include the GMGPL variants.

Fabien-Chouteau:
To refine what i meant in alire-project/alire-index/pull/62, I think that the license field should be strictly limited to identifiers. So having the text of a custom license in the field should be forbidden.

In other words, https://github.com/alire-project/alire/blob/545d49995bb25bc4166e8bf0470110b27ff8a67c/src/alire/alire-licensing.adb#L18 should not return `Unknown`, but report and error.

mosteo:
Ah, I see. Not the most intuitive code, but the function that loads from the index does the check:
https://github.com/alire-project/alire/blob/545d49995bb25bc4166e8bf0470110b27ff8a67c/src/alire/alire-properties-licenses.ads#L57

To have a custom license text you have to precede it with `custom:` to avoid the error.

From another POV: In Alire.Licensing we have just an enumeration (because originally it was intended to be perhaps auto-generated), in Alire.Properties.Licenses the possibility of a custom license is added.

Fabien-Chouteau:
I don't really see the requirement for full custom license text field, and it will be difficult to present on the website.

I will open a PR with a proposal.

Fabien-Chouteau:
I looked at the SPDX stuff and it is much more complex than I expected with the operators `WITH`, `AND`, `OR`.

PR #244, opened 2019-11-26
Licenses: remove the possibility of custom license text

Fabien-Chouteau:
As discussed in #191, it is enough to limit the license field to one of
the possible identifier supported.


pmderodat:
Thanks for doing this! Could you please update the index format specification accordingly?

Issue #178, opened 2019-09-12
"path" property didn't make to new index


mosteo:
A `Path` property was used in the old index gnat crate to force the use of the system compiler in crates that depend on native libraries (e.g., eagle_lander). That property no longer exists in the new index syntax.

I am not sure reintroducing it is the best course of action yet (it might depend on the rework of native packages); but I'm making a note here not to forget about this issue. The current situation is that crates that depend on the native compiler will fail if attempted to be build with a non-native one (although all its native dependencies will be installed), but will work normally otherwise.

mosteo:
I don't believe this is an issue anymore as it hasn't been a problem in over 5 years (crossing fingers).

Issue #94, opened 2019-04-22
Perform `gprinstall` to create unified environment


mosteo:
Currently, compiled projects stay where they are checked out. Running `gprinstall` will unify the location of libraries and executables. This might clarify the build products of a session folder.

This would make obsolete the `path` attributes (see #178) of projects managed with `gprbuild` (all of them, currently).

mosteo:
I think we now lean more to not do this, at least not for all dependencies, as that might not be easily doable for some crates depending on how they use project files, so I'm closing it.

We might have a use for explicit `alr install` instead (#482).

PR #245, opened 2019-11-27
Don't run a default executable if more than one declared by the release

### LINKED ISSUE #239: Handling of multiple binaries
I have a simple project file:

```sh
$ cat FOO.gpr 
project FOO is
   for Source_Dirs use (".");
   for Object_Dir use "obj";
   for Main use ("bar.adb", "baz.adb");
end FOO;
```

The toml file has the following additions to the default template:

```sh
project-files = ["FOO.gpr"]
executables = ["bar", "baz"]
```

Both executables are found after compiling:

```sh
$ alr run --list
Project foo builds these executables:
   bar (found at alire/build/obj/bar)
   baz (found at alire/build/obj/baz)
```

But running yields:

```sh
$ alr run
Warning: Executable "foo" not found
ERROR: alr run unsuccessful
```

When removing the second binary, bar is executed. Seems like `alr` uses the configured binary when only one is present, but falls back to the project name when multiple binaries are present.

* Is there a way to set the default binary to run on `alr run`?
* I wonder why binaries need to be configured in the catalog at all - couldn't this be derived from the project file's `for Main use (...)`?

mosteo:
Fixes #239.

`alr run` without executable won't proceed if the release declares more than one executable. If exactly one is declared, it will be run. If none is declared but an executable file with the name of the crate exists, it will be run.


Fabien-Chouteau:
I'm sorry to react so late to this question, but I wonder if we need that much complexity in the handling of executables. In particular I think that  the part that searches for the executables is not required.

In my opinion the `executables` field should only contain paths to the executables (relative to the build dir), and then either the executable is here and we execute it or it is not and we return an error. So I would remove the call to `Locate_File_Under`.

mosteo:
I'm OK either way. I think I wrote this somewhere else, but the original reason for not giving executable paths was that (at least in certain toolsets) it's usual that the executable will be produced in a path that depends on the build config (e.g. debug, release).

~~This is moot if we don't want to support that case. Or we could support that, by specifying all possible locations of an executable. This however introduces an ambiguity about what executables are going to be actually built vs ones that might be possibly built.~~

I just realized that if we add the `gprinstall` step (#94), we will know the location of all executables and we could keep the non-prefixed list. Thoughts?

mosteo:
@Fabien-Chouteau, since this is a fix for the statu quo, what do you think if we merge this one and tackle the larger point in a separate PR?

Fabien-Chouteau:
> @Fabien-Chouteau, since this is a fix for the statu quo, what do you think if we merge this one and tackle the larger point in a separate PR?

Sorry for the delay.

I agree with the current solution, I just have one question because I tried this patch on a project with  just one executable and got this output:
```
-> alr build is (PROJECT) (0 releases indexed) (loaded in 0.001s)
-> Using candidate project root: /home/chouteau/src/github/ada-voxel-space
-> run:
-> Using candidate project root: /home/chouteau/src/github/ada-voxel-space
-> Generating alr buildfile: /home/chouteau/src/github/ada-voxel-space/alire/alr_build.gpr
-> Dependency resolution failed
ERROR: alr run unsuccessful
``` 
So the `Dependency resolution failed` error comes from the call to `Commands.Compile.Execute` ([alr-commands-run.adb:132](https://github.com/alire-project/alire/blob/457f2400a3f45f6ccbfc249ccb4de80f4f9cdce8/src/alr/alr-commands-run.adb#L132)).

So my question is what does the `compile` command does compared to the `build`.

mosteo:
> So my question is what does the compile command does compared to the build

Sorry for the long reply, I tend to digress:

Build is `update` followed by `compile`. The idea (taken from `cargo` IIRC) is that you use `build` when you want to fetch new releases of dependencies and recompile, or after you add a dependency, so it is fetched before attempting compilation.

Update solves dependencies and regenerates the metadata files in the `alire` folder (toml + gpr). It is kind of a misnomer, since `update` doesn't refresh the index (that's in `alr index --update-all`). `update` and `build` used to have an `--online` flag to request updating of the index prior to dependency resolution; it was lost at some point during the index reworking. Perhaps we could think of a better name for `update`.

The interrelations between these commands are shown in Fig.2 of the [original paper](https://github.com/alire-project/alire/blob/master/doc/2018-03.alr-draft.pdf).

Compile shouldn't really try to resolve dependencies (it does though, as you see in the error you get). This happens now in order to recompute the closure of actual dependencies in use, for a feature that is no longer in the new index (#178). I don't think we need that feature back, if we go the way of `gprinstall` (#94). Either way, compile shouldn't require to resolve dependencies (I'll open an issue for that).

I guess your error comes from the resolution attempted by `compile`. So `update` should fail too, for the same reason.

Fabien-Chouteau:
> I guess your error comes from the resolution attempted by `compile`. So `update` should fail too, for the same reason.

I tried again on a fresh setup, and `alr run` works...

We are good to go, sorry for the noise.

mosteo:
Still, you reminded me of the `compile` issue, which is good :)

Issue #242, opened 2019-11-24
Reporting invalid crate name


Fabien-Chouteau:
I was trying to create an Alire crate and got confused by the error message:
```
$ alr init --bin GESTE-examples
ERROR: A project/version string was invalid
ERROR: alr init unsuccessful
$ alr init --bin GESTE_examples
ERROR: A project/version string was invalid
ERROR: alr init unsuccessful
```
The problem was actually the uppercase letters.

It would be nice to have more explicit error message here, including the crate naming rules.

In particular since this is likely one of the first commands beginners will try.

PR #247, opened 2019-11-27
Extend `alr help` to support non-command topics

mosteo:
We have several instances (#242, probably more) where errors are detected but no further information is provided. Besides providing better information on the spot, with this patch we can define more general topics that the user can check for help.

Also the help support becomes a bit better structured, with some parts being moved from Alr.Commands to Alr.Commands.Help.

A minimal new topic is included: `identifiers` describes the rules for crate/index names.

### DIFF
--- a/src/alr/alr-commands-help.adb
+++ b/src/alr/alr-commands-help.adb
@@ -1,0 +1,129 @@
+with AAA.Enum_Tools;
+with AAA.Table_IO;
+with AAA.Text_IO;
+
+package body Alr.Commands.Help is
+
+   type Help_Topics is (Identifiers);
+   --  Enumeration used to index available help topics.
+
+   -----------------------
+   -- One_Liner_Summary --
+   -----------------------
+
+   function One_Liner_Summary (Topic : Help_Topics) return String
+   is (case Topic is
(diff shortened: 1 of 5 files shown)


Fabien-Chouteau:
Good idea Alex, this very much needed. Do you think it is also possible to display naming rules at the point of failure (in the error message reporting the naming error)?

Fabien-Chouteau:
Now that I think of it, the list of commands is also displayed when `alr` is run without arguments: https://github.com/alire-project/alire/blob/feat/help/src/alr/alr-commands.adb#L230

If there is a single function that prints commands and topics it should be used here as well.

mosteo:
> Now that I think of it, the list of commands is also displayed when `alr` is run without arguments: https://github.com/alire-project/alire/blob/feat/help/src/alr/alr-commands.adb#L230
> 
> If there is a single function that prints commands and topics it should be used here as well.

I'll replace this with the new Help.Display_Valid_Keywords before merging.

> Do you think it is also possible to display naming rules at the point of failure (in the error message reporting the naming error)?

Sure, I'll see how to patch in the information when addressing #242

Issue #253, opened 2019-12-02
Spawn: use array of strings for arguments


Fabien-Chouteau:
Since the different `Spawn` functions in `Alire.Os_Lib.Subprocess`  take the command line as a string and then split it with `System.Os_Lib.Argument_String_To_List` we have to be careful with white-spaces in paths. Having a space in my Windows user directory, I notice a few problems when spawning `tar` and `git`.

It seems like directly using an array of string to pass the arguments to `Spawn` would make this easier to handle in the code.

mosteo:
So, with #252, we should converge to a single `Checked_Spawn` that takes an array of strings, right?

Fabien-Chouteau:
I think so.

The problem is that arrays of strings are a pain in the back in Ada. Maybe String vector is better.

Also it looks like `Alr.Os_Lib` also has subprocess function that should be replaced.

mosteo:
Yep, that's why `Alire.Utils.String_Vector` has chained ops so it can be used in place: `Empty_Vector.Append ("foo").Append ("bar")`. Still a bit verbose...

Fabien-Chouteau:
The Ada 202x container aggregates will be nice in that situation. But let's not wait for that to refactor.

mosteo:
Fixed by #255.

Issue #252, opened 2019-12-02
Unified behavior for Alire.Subprocess


Fabien-Chouteau:
Using `alr` on Windows I noticed strange behavior of `git clone` depending on the `-v` option was passed to `alr` or not.

I realized that `Spawn`  is called in has many different ways depending on verbose, quite and debug level.

I also think that `Alire.Os_Lib.Subprocess.Spawn` should be remove and replaced by `Checked_Spawn` everywhere.

PR #255, opened 2019-12-03
Alire.Os_Lib.Subprocess refactoring

Fabien-Chouteau:
The main goal of this patch is to move all the sub-process APIs to use
arguments in the form of a vector of strings rather than a single string
that is later sliced into different arguments. The reason for this
change is to better handle white-spaces in arguments, in particular
file-system paths.

The patch also removes unused or duplicate sub-programs in
Alire.Os_Lib.Subprocess and Alr.Os_Lib.

Because of the change of actions.command format in the manifests, this
patch requires an update of some crates in the index.

Part of #253 and #252.


Fabien-Chouteau:
Unfortunately I think that the testsuite does not cover all the changes made in this patch, in particular `hg` and `svn` VCSs .

Before merging it would be better to add at least a test on the `action.command` of the manifest.

mosteo:
> Unfortunately I think that the testsuite does not cover all the changes made in this patch, in particular `hg` and `svn` VCSs .

We aren't using `hg` or `svn` anywhere anymore, I think...

src/alire/alire-origins-deployers-apt.adb
```diff
@@ -42,19 +43,21 @@ package body Alire.Origins.Deployers.APT is
...
-                            This.Base.Package_Name (Platform.Distribution));
-
-      if Exit_Code /= 0 then
-         return Outcome_Failure ("apt-get install exited with code:"
-                                 & Exit_Code'Img);
-      end if;
+      Subprocess.Checked_Spawn
+        ("sudo", Empty_Vector &
+           "apt-get" &
+           "instal" &
```
mosteo: install typo

mosteo:
> Before merging it would be better to add at least a test on the `action.command` of the manifest.

Waiting for this before merging, then.

src/alire/alire-origins-deployers-apt.adb
Fabien-Chouteau: Good  catch!

Fabien-Chouteau:
I changed the `actions.command` definition to only accept array of string(s). In the previous version single string was accepted but I realized that it could be error prone to accept `command = "make something"`

mosteo:
Great, it's surely more robust. Also, there are no ambiguities accepted in any other place in the index format either (in the sense of accepting a single value where an array is expected).

Fabien-Chouteau:
> We aren't using `hg` or `svn` anywhere anymore, I think...

Should we remove them?

mosteo:
> Should we remove them?

I'd just leave them there, if they're not giving you extra work. Otherwise, I don't foresee a clear need.

Fabien-Chouteau:
> I'd just leave them there, if they're not giving you extra work. Otherwise, I don't foresee a clear need for them.

The problem is that this is more or less dead code.

mosteo:
Out with it, then.

Fabien-Chouteau:
The code is not that big and it will be possible to bring it back if need be.

Let's remove that in a separate PR though.

mosteo:
OK! I will merge this one shortly.

mosteo:
To be merged with https://github.com/alire-project/alire-index/pull/72

PR #257, opened 2019-12-04
Better feedback on bad crate/index names

mosteo:
With this patch, instead of forcing us to code explicit validation or unexpectedly failing an assertion, a more precise `Checked_Error` is generated that better informs the user of what is going wrong.

I was unaware of the `Predicate_Failure` aspect until recently, and I see big potential in it to automatically enforce predicates while simultaneously giving good feedback to users.

@Fabien-Chouteau , I see we don't have a maximum length limit on names. I guess this can bite us at some point. I just checked and found some limits in other PMs are: [cargo, 64](https://github.com/rust-lang/crates.io/issues/696), [npm, 50](https://github.com/npm/validate-npm-package-name/issues/4)


mosteo:
I just went ahead and enforced a Max length of 64 (we already had the unused Max length (?)), let me know if you have any concern about it.

Fabien-Chouteau:
Sorry I forgot to reply. 

We definitely need a upper limit, and 64 sounds good.

PR #258, opened 2019-12-04
Rename `compile` to `build`

mosteo:
As discussed in Gitter, `build` is currently not doing anything extra over `compile`, and seems a better term for the general process of generating a library/executable.


Fabien-Chouteau:
Looks good, the `alr get --compile` switch should also be renamed.

mosteo:
Good point...

PR #264, opened 2019-12-04
Fix sneaky space that broke git pulls

mosteo:


mosteo:
I test would be handy but I will get this one in first so we don't have a broken feature in master

Issue #160, opened 2019-07-18
Remove global states


mosteo:
Some globals exist from where recompilation existed which no longer make sense:

- [ ] Config
- [ ] Index
- [ ] Project descriptions

PR #266, opened 2019-12-08
Fix duplicate Not_Interactive config

Fabien-Chouteau:


mosteo:
This is another global to be addressed at some point (#160)

PR #268, opened 2019-12-09
Check index version during loading

### LINKED ISSUE #234: Check index version during loading
We currently have a metadata file at the root of indexes that contains a version for the whole index. We check the existence of this file but we do nothing with the version.

Steps in increasing order of work:
1. Stop on version mismatch.
1. Accept indices with same major number, ignore unknown fields in that case.
    1. This would allow us to add fields to the index relatively painlessly in a non-breaking way.

mosteo:
Turns out we had already minimal version checking (#234) but inoperative due to a bug. This patch fixes the bug and adds better feedback in case of version mismatch, and a couple of tests.

Fixes #234


mosteo:
> The current version should probably not be 1.0 since we are still in beta.

We can move to `0.x` when we prepare the index branching?

> Also what will be the version number for the development version?

We could use the `0.x` indefinitely for those, although it'll be strange to use a prior version to that in stable. Other idea is to prefix the branch names so there's no possible confusion, and continue numbering normally: `devel-0.1`, `stable-1.0`, etc.

PR #269, opened 2019-12-10
Remove VCS origins

mosteo:
Those were deprecated in favor of only source archives. This is a cleanup of the now unused code.


Fabien-Chouteau:
I don't think agree with this one. Maybe I should read the discussion again, but git origin seems like a useful feature for local indexes.

Fabien-Chouteau:
Provided we can test it.

mosteo:
No objection from me. We might then enforce that those must be local? All the issue stems that we didn't find an easy way of checking the integrity of this kind of remotes.

A separate issue would be if we want to keep hg/svn too.

> Provided we can test it.

In the ultimately dropped #198 there were local tests using local repos, IIRC.

Fabien-Chouteau:
> We might then enforce that those must be local?

Yes, forbid them in the community index with a check.

PR #270, opened 2019-12-10
Remove all support for SVN/Hg

mosteo:
~~Note: with this we lose the ability to use indexes stored in these kinds of remotes. This is something we didn't discuss in [#255](https://github.com/alire-project/alire/pull/255#issuecomment-561721869) (I missed this possibility). So it might still be worth keeping this code.~~

The above is wrong; there's no support in any case.


mosteo:
Kept for now for local origins (#272)

PR #271, opened 2019-12-11
Add missing Subprocess.Checked_Spawn_And_Capture

mosteo:
To complete #252


src/alire/alire-os_lib-subprocess.ads
```diff
@@ -2,17 +2,25 @@ with Alire.Utils;
 
 package Alire.OS_Lib.Subprocess is
 
-   --  Separate from Alire.OS_Lib because not preelaborable
+   --  Separate from Alire.OS_Lib because not preelaborable.
```
Fabien-Chouteau: We usually don't add dots (`.`) for single line comments.

mosteo: Understood

mosteo:
> So you removed the progress version.

But it wasn't used anywhere? I thought in fact that you had removed it in your rework :confused:

mosteo:
I think I know what's happened, so indeed I may have removed it involuntarily. Let me do some checks...

Fabien-Chouteau:
When you do `alr build` I think the progress animation is used.

Note that I am personally not a huge fan of this kind of animation, I prefer to see the output.

mosteo:
OK, my preliminary inspection shows that the progress version is no longer used (at least in `alr build`) after the Windows PR. I presume as part of your rework to the Spawn versions (that was my original impression).

:later:

I bisected it to commit 49e217298 in the Windows PR.

That said, the output to console is something that for some time I've thought we might want to rework, so I'm ok with losing this now. The only thing is that right now, for long compilations, alr is eerily quiet... (I was trying with the GNATCOLL crate).

Fabien-Chouteau:
> I bisected it to commit [49e2172](https://github.com/alire-project/alire/commit/49e217298a481850c853fcd70525b49fb76d082f) in the Windows PR.

It was not my intention but OK.

> That said, the output to console is something that for some time I've thought we might want to rework, so I'm ok with losing this now. The only thing is that right now, for long compilations, alr is eerily quiet... (I was trying with the GNATCOLL crate).

Probably for build and actions command we should let the output pass through.

PR #277, opened 2019-12-14
MacOS support

simonjwright:
This pull supports macOS.
Building on macOS requires the user to define the environment/scenario variable `OS` to `macOS`.
The changes that may have a wider impact are
* the use of `curl` instead of `wget`, which isn’t available on macOS
* the way that the tar option --force-local is only passed on Windows
* the action workflows run on push as well as pull_request.


alr.gpr
```diff
@@ -12,8 +12,9 @@ project Alr is
    Src_Dirs := ("src/alr");
 
    case Alire_Common.Host_Os is
-      when "Windows_NT" => Src_Dirs := Src_Dirs & ("src/alr/os_windows");
-      when others       => Src_Dirs := Src_Dirs & ("src/alr/os_linux");
+      when "windows" => Src_Dirs := Src_Dirs & ("src/alr/os_windows");
+      when "osx"     => Src_Dirs := Src_Dirs & ("src/alr/os_macos");
```
Fabien-Chouteau: Why `osx` and not `macos` here?

Fabien-Chouteau (approved):

Fabien-Chouteau:
Thank you @simonjwright that's great!

alr.gpr
simonjwright: I do see what you mean: `alr_env.gpr` sets `GNATCOLL_OS` to `osx`, & I just followed that for `ALIRE_OS`.

mosteo:
Thanks a lot, @simonjwright.

PR #282, opened 2019-12-17
Restore progress bar with `curl` in default log level

mosteo:
We had this with `wget` so long downloads don't make `alr` appear to do nothing. It can still be silenced with `-q`.


simonjwright:
Looks good! Not sure why `curl` has to tell us twice ...
```
$ rm -rf hello_1.0.1_dcc36a2f/
$ alr get hello
######################################################################## 100.0%
######################################################################## 100.0%
$ rm -rf hello_1.0.1_dcc36a2f/
$ alr get -q hello
$
```

PR #288, opened 2019-12-19
[RFC] Filesystem deployer: allow the use symlinks for local crates

Fabien-Chouteau:
This makes it easier to work on multiple crates at the same time by using symlink to local repositories.


Fabien-Chouteau:
I know this will fail on Windows. I will look at the available options on that platform.

Fabien-Chouteau:
The problem that this patch is trying to solve will be better implemented in the lockfile.

PR #294, opened 2019-12-29
Explicitly use python2 for test suite

mosteo:
Some distros are already moving to Python 3 (e.g. CentOS) and this is likely to continue as support for Python 2 ends. Until the e3 testsuite is migrated we ensure we are using/finding the Python 2 interpreter.


mosteo:
@Fabien-Chouteau I'm merging this one right away since CI is broken otherwise (long story short: CentOS 7 -> 8 exposed our assumptions on Python 2).

PR #295, opened 2020-01-01
[RFC] Introducing Externals to fix the Native situation (and hinting in general)

### LINKED ISSUE #65: Handling of native packages

mosteo:
(Note: this PR builds upon still-unmerged PRs so I don't recommend to look at the code just yet.)

This, with some upcoming PRs, would fix #65 and future-proof Alire for the use of externally provided releases. I'm going to present a tentatively complete solution, but now would be a good moment to discuss conceptual shortcomings.

Recap: native releases were introduced as a way to jumpstart the availability of some complex dependencies. In its current form they have, at least, two main problems: they do not serve to represent other kind of "externally provided" dependencies, and there is no way to know in advance the exact version used by some native packages, hence they cannot be accurately represented in Alire, which requires a known version for every release.

This patch introduces a new class of index entities, which I've called Externals. Crates may have externals in addition to regular releases. The main point and difference is that externals can generate releases at run time, when we have the opportunity to prospect the current platform. Those "detected" releases then can be used normally during dependency resolution, or we may choose to simply use the existence of externals as a hint that a dependency cannot be provided and the user must fulfill it (just as we are already doing with native releases).

Externals are classwide, so we may expand the suite of available features as needs arise, and hence are represented in the index as top-level array entries:
```
[[external]]
# data specific to a external kind
```

Externals work in two steps: pre-detection they are used for hinting and general information (in `alr show`, `alr search`...); post-detection they add any detected releases to the index and those releases are used as dependencies normally. They can be detected on demand, only when needed (e.g. when a dependency for their crate appears), to not slow down resolution innecessarily.

I have been testing three kinds of externals that cover every shortcoming I had seen identified till date. This PR introduces only the first of them, the others I will open at a later time if we decide the idea is sound.

1. Unindexed external: plainly, a crate we know exists, but is not yet in Alire (e.g., `aws`, needed by alire-project/alire#73). This enables the indexing of dependent crates without hidden dependencies, and a hint is emitted on resolution. These don't really detect releases and are intended as an interim measure until the dependency is properly added to Alire.

2. Executable external: a tool (e.g., `gnat` itself, `make`...) whose version can be ascertained by invoking it (e.g., `make --version`, `gnat --version`). These generate releases with filesystem origin (that we already have). The external is defined by the command to run and the regular expression to apply to extract the version. This also serves to improve the situation with the compiler versions (related: #241).

3. Native external: our current native releases, but properly working. Instead of having the candidate package names in the release, this information is moved to the external. If the external detects an available native package, it creates a native release which has the proper version and an origin with the package to install.

Although the explanation is long, the core idea is simple and implementation also is quite straightforward. You can play with my current prototype by compiling my [external-native](https://github.com/mosteo/alire/tree/feat/external-native) branch and running, e.g., `alr search --list --external` and `alr search --list --external-detect` to see the different externals in action (look for releases with NOTES text), and their definitions in e.g. the [aws](https://github.com/mosteo/alire-index/blob/feat/external-0.4/index/aw/aws.toml), [make](https://github.com/mosteo/alire-index/blob/feat/external-0.4/index/ma/make.toml), [libsdl2](https://github.com/mosteo/alire-index/blob/feat/external-0.4/index/li/libsdl2.toml), [ncursesada](https://github.com/mosteo/alire-index/blob/feat/external-0.4/index/nc/ncursesada.toml) crates.

Looking forward to your feedback.


Fabien-Chouteau:
It does sound like an improvement on the current situation.

Although I am a bit worried about the growth of complexity, it could be a solution for detecting and specifying dependency on cross tool-chains.

mosteo:
I understand your concern (as always when I submit a largish PR :-)). I'd say, as interested party, that by separating releases from externals the situation (and the code) is in practice simpler.

When time permits I will prepare the follow-up PRs so you can better look at the changes. The final result is what the branch in the last paragraph points to.

Fabien-Chouteau:
BTW `gpconfig` provides a way to detect compilers and run-times:
```
$ gprconfig --target=all --mi-show-compilers
   1 target:x86_64-linux-gnu
   1 normalized_target:x86_64-linux-gnu
   1 executable:gnatls
   1 path:/usr/bin/
   1 lang:Ada
   1 name:GNAT
   1 version:8
   1 runtime:default
   1 native:TRUE
   2 target:arm-none-eabi
   2 normalized_target:arm-none-eabi
   2 executable:arm-none-eabi-gcc
   2 path:/usr/bin/
   2 lang:Asm
   2 name:GCC-ASM
   2 version:7.3.1
   2 runtime:
   2 native:FALSE
[...]
```
 I have a piece of code that parses that if need be.

mosteo:
Ahhh that was the piece I was missing... This for sure will come in handy

Fabien-Chouteau:
Here is the `gprconfig` "parser" I played with.

`gprconfig` outputs a different entry for each compiler+runtime combination, so that is what generate here. Maybe it would be better to have the run-times as a list for each compiler in our representation.

```ada
with Ada.Containers.Vectors;

with Semantic_Versioning;

package Alire.Compilers is

   type Compiler_Info is record
      Name    : Ada.Strings.Unbounded.Unbounded_String;
      Target  : Ada.Strings.Unbounded.Unbounded_String;
      Version : Semantic_Versioning.Version;
      Path    : Ada.Strings.Unbounded.Unbounded_String;
      Lang    : Ada.Strings.Unbounded.Unbounded_String;
      Runtime : Ada.Strings.Unbounded.Unbounded_String;
   end record;

   package Compiler_Info_Vectors is new Ada.Containers.Vectors
     (Index_Type   => Positive,
      Element_Type => Compiler_Info);

   subtype Compiler_Info_Vector is Compiler_Info_Vectors.Vector;

   function All_Compilers return Compiler_Info_Vector;

   function Ada_Compilers return Compiler_Info_Vector;

   function Filter_By_Lang (Vect : Compiler_Info_Vector;
                            Lang : String)
                            return Compiler_Info_Vector;

   function Filter_By_Target (Vect   : Compiler_Info_Vector;
                              Target : String)
                              return Compiler_Info_Vector;

   function Filter_By_Runtime (Vect    : Compiler_Info_Vector;
                               Runtime : String)
                               return Compiler_Info_Vector;

   procedure Print (Compilers : Compiler_Info_Vector;
                    Level     : Trace.Levels := Trace.Always);

end Alire.Compilers;
```
```ada
with Ada.Strings.Unbounded;
with GNAT.Regpat;

with Alire.Utils;
with Alire.OS_Lib.Subprocess;

package body Alire.Compilers is

   package Semver renames Semantic_Versioning;

   Cached_Compilers : Compiler_Info_Vector;

   -----------
   -- Print --
   -----------

   procedure Print (Compilers : Compiler_Info_Vector;
                    Level     : Trace.Levels := Trace.Always)
   is
      use Ada.Strings.Unbounded;
   begin
      for Comp of Compilers loop
         Trace.Log ("---- " & To_String (Comp.Name), Level);
         Trace.Log ("  version: " & Semver.Image (Comp.Version), Level);
         Trace.Log ("  Runtime: " & To_String (Comp.Runtime), Level);
         Trace.Log ("  Lang   : " & To_String (Comp.Lang), Level);
         Trace.Log ("  Path   : " & To_String (Comp.Path), Level);
         Trace.Log ("  Target : " & To_String (Comp.Target), Level);
      end loop;
   end Print;

   -------------------
   -- All_Compilers --
   -------------------

   function All_Compilers return Compiler_Info_Vector is
      use Ada.Containers;
      use Ada.Strings.Unbounded;
      use Alire.Utils;
      use GNAT.Regpat;

      Result : Compiler_Info_Vector renames Cached_Compilers;
      Output : String_Vector;

      Grps : Match_Array (0 .. 3);

      Compiled_Expression : constant Pattern_Matcher :=
        Compile ("^[ ]+([0-9]+) ([a-z_]+):(.*)$");

   begin
      if Result.Length /= 0 then
         return Cached_Compilers;
      end if;

      Output := Alire.OS_Lib.Subprocess.Checked_Spawn_And_Capture
        ("gprconfig",
         Alire.Utils.Empty_Vector &
           "--target=all" &
           "--mi-show-compilers",
         Err_To_Out => True);

      for Str of Output loop
         Match (Compiled_Expression, Str, Grps);

         if Grps (0) /= No_Match then
            declare
               Id  : constant Positive :=
                 Positive'Value (Str (Grps (1).First .. Grps (1).Last));
               Key : constant String := Str (Grps (2).First .. Grps (2).Last);
               Val : constant String := Str (Grps (3).First .. Grps (3).Last);
            begin
               if Id > Result.Last_Index then
                  Result.Set_Length (Ada.Containers.Count_Type (Id));
               end if;

               if Key = "name" then
                  Result.Reference (Id).Name := To_Unbounded_String (Val);
               elsif Key = "target" then
                  Result.Reference (Id).Target := To_Unbounded_String (Val);
               elsif Key = "version" then
                  Result.Reference (Id).Version := Semver.Parse (Val);
               elsif Key = "runtime" then
                  Result.Reference (Id).Runtime := To_Unbounded_String (Val);
               elsif Key = "lang" then
                  Result.Reference (Id).Lang := To_Unbounded_String (Val);
               elsif Key = "path" then
                  Result.Reference (Id).Path := To_Unbounded_String (Val);
               end if;
            end;
         end if;
      end loop;
      return Result;
   end All_Compilers;

   --------------------
   -- Filter_By_Lang --
   --------------------

   function Filter_By_Lang (Vect : Compiler_Info_Vector;
                            Lang : String)
                            return Compiler_Info_Vector
   is
      use Ada.Strings.Unbounded;

      Result : Compiler_Info_Vector;
   begin
      for Elt of Vect loop
         if Elt.Lang = Lang then
            Result.Append (Elt);
         end if;
      end loop;
      return Result;
   end Filter_By_Lang;

   ----------------------
   -- Filter_By_Target --
   ----------------------

   function Filter_By_Target (Vect   : Compiler_Info_Vector;
                              Target : String)
                              return Compiler_Info_Vector
   is
      use Ada.Strings.Unbounded;

      Result : Compiler_Info_Vector;
   begin
      for Elt of Vect loop
         if Elt.Target = Target then
            Result.Append (Elt);
         end if;
      end loop;
      return Result;
   end Filter_By_Target;

   -----------------------
   -- Filter_By_Runtime --
   -----------------------

   function Filter_By_Runtime (Vect    : Compiler_Info_Vector;
                               Runtime : String)
                               return Compiler_Info_Vector
   is
      use Ada.Strings.Unbounded;

      Result : Compiler_Info_Vector;
   begin
      for Elt of Vect loop
         if Elt.Runtime = Runtime then
            Result.Append (Elt);
         end if;
      end loop;
      return Result;
   end Filter_By_Runtime;

   -------------------
   -- Ada_Compilers --
   -------------------

   function Ada_Compilers return Compiler_Info_Vector is
   begin
      return Filter_By_Lang (All_Compilers, "Ada");
   end Ada_Compilers;

end Alire.Compilers;
```

mosteo:
Thanks for the snippet! I need to learn more about targets/runtimes.

mosteo:
@Fabien-Chouteau , I have restructured this PR into four new ones (#302 to #305) so ideas are not intermixed.

The first one contains common changes to support the other three. Each of these latter three defines a new kind of external, corresponding to the ones described in this PR. These three are independent of the rest, so each can be discussed/merged/discarded independently.

Fabien-Chouteau:
> The first one contains common changes to support the other three.

I was going to say, there is a lot of shared code between those PRs :)

mosteo:
> I was going to say, there is a lot of shared code between those PRs :)

This is something I haven't found how to do; when the PRs come from a fork, you cannot use as base another PR in the original repo :S 

This is a reason for using branches in the main repo, but then I'm growing wary of polluting the main repo with several in-progress branches.

mosteo:
Closing, since this is superseded by #302-#305

PR #296, opened 2020-01-03
Refactor Project to Crate

### LINKED ISSUE #113: Refactoring Project --> Crate
We adopt 'crate' for the bundles of GPR project + Ada code downloaded by alr. This was previously called Project through alire, with the risk of confusion with GPR projects or abstract projects.

This entails:
1. Plain refactoring of Project --> Crate through code
2. A possible opportunity to enhance `Alire.Projects` to better store all crate-exclusive information which cannot be overriden by releases.

mosteo:
Fixes #113.


src/alire/alire-containers.ads
```diff
@@ -31,12 +31,12 @@ package Alire.Containers with Preelaborate is
                                              Releases."=");
    subtype Release_H is Release_Holders.Holder;
 
-   package Project_Release_Maps is new Ada.Containers.Indefinite_Ordered_Maps
-     (Alire.Project, Releases.Release, "<", Releases."=");
-   type Release_Map is new Project_Release_Maps.Map with null record;
+   package Name_Release_Maps is new Ada.Containers.Indefinite_Ordered_Maps
```
Fabien-Chouteau: Shouldn't this be `Crate_Release_Maps`?

src/alire/alire.ads
```diff
@@ -51,8 +51,8 @@ package Alire with Preelaborate is
    function "+" (S : UString) return String
    renames UStrings.To_String;
 
-   subtype Project_Character is Character
-      with Static_Predicate => Project_Character in
+   subtype Crate_Character is Character
+      with Static_Predicate => Crate_Character in
          'a' .. 'z' | '0' .. '9' | '_' | Extension_Separator;
```
Fabien-Chouteau, line 56: Completely unrelated: Why hyphen is not allowed in crate name?

src/alr/alr-commands-search.adb
```diff
@@ -104,14 +104,14 @@ package body Alr.Commands.Search is
...
-           (Crate : Alire.Projects.With_Releases.Crate) is
+           (Crate : Alire.Crates.With_Releases.Crate) is
          begin
             if Cmd.Full then
                for Release of Crate.Releases loop
                   List_Release (Release);
                   Busy.Step;
                end loop;
-            else
+            elsif not Crate.Releases.Is_Empty then
```
Fabien-Chouteau, line 114: This is more than renaming. What is the reason for the change?

src/alr/alr-paths.ads
```diff
@@ -92,7 +92,7 @@ private
    function Alr_Source_Folder return String
    is (OS_Lib.Getenv (Alire.Environment.Source, Alr_Config_Folder / "alire"));
 
-   function Projects_Folder      return String
-   is (Alr_Working_Cache_Folder / "projects");
+   function Releases_Folder      return String
+   is (Alr_Working_Cache_Folder / "releases");
```
Fabien-Chouteau: Why not `"crates"` here?

src/alire/alire-containers.ads
mosteo: Sometimes I have difficulties choosing between name and crate depending on context. I change it if you find it more consistent.

src/alire/alire.ads
mosteo: When the index was Ada-based I tried to keep things more closer to Ada syntax (indeed the hyphens in the toml syntax ruffle me the wrong way, but since it's a matter of taste...) So no actual reason now I'd say.

src/alr/alr-commands-search.adb
mosteo: Ah, good eye. It's a bug fix needed after #293, for the case where we had a crate without releases (there aren't in the current index). This PR was initially on top of my externals experiments in #295. I caught the bug then, and when rebasing this PR I didn't realize this was a functional change (even if it's a bug fix).

src/alr/alr-paths.ads
mosteo: I've realized that many times when we talk about crates we are talking about a particular release of a crate. In this case, we are downloading releases, so that's it. Although it's kind of splitting hairs probably.

mosteo: On further thought, this should probably be better "dependencies"

src/alire/alire.ads
Fabien-Chouteau: I'm mostly curious, I do like hyphens in project names better than underscores (`lvgl-ada` vs `lvgl_ada`) but since allowing both would be a source of confusion let's keep it how it is.

src/alire/alire-containers.ads
Fabien-Chouteau: From what I know of the code it makes more sense to me.

mosteo: Done in e07df2a

src/alr/alr-paths.ads
mosteo: Done in e07df2a

Fabien-Chouteau (approved):

PR #301, opened 2020-01-16
[RFC] Replace the alr_build.gpr aggregate project by environment variables

Fabien-Chouteau:
Instead of creating an aggregate project that contains the path to
dependencies' gpr files, and the list of gpr files to build for the
crate. This patch introduce a procedure that sets the GPR_PROJECT_PATH
environment variable and runs gprbuild one time for each of of gpr file
of the crate. Same goes for the gprclean.

The keys/values of external scenario variable are also set in the
environment.

This patch also introduces a setenv command that prints the
GPR_PROJECT_PATH and external scenario variable. This can be used to set
a correct build environment before starting an IDE for instance.


src/alr/alr-build_env.adb
```diff
@@ -0,0 +1,175 @@
...
+with Alr.Commands;
+
+package body Alr.Build_Env is
+
+   type Env_Var_Action_Callback is access procedure (Key, Val : String);
+
+   Path_Separator : constant Character := GNAT.OS_Lib.Path_Separator;
+
+   -----------------
+   -- Projet_Path --
```
mosteo: Typo

Fabien-Chouteau: Indeed, thank you :)

src/alr/alr-commands-build.adb
```diff
@@ -15,13 +17,24 @@ package body Alr.Commands.Build is
 
    procedure Do_Compile is
    begin
+      Requires_Full_Index;
```
Fabien-Chouteau, line 20: @mosteo I am not sure about the consequences of calling this procedure here.
I guess this will not fetch the index from the web, but what are the consequences in terms of performances?

mosteo: That loads all the local indexes. It does not reload unless explicitly forced, so it is safe to call at any point where we want the crates to be already loaded.

In the case of `alr build`, it is there so post-compile actions are known and run. Obviously we could make things more efficient by recursively loading dependencies only, or caching actions after retrieval... I'm actually unsure if any crate uses `post-compile` actions.

I guess as the index grows bigger, loading time may become noticeable and we may need to do something about it.

src/alire/alire-origins-deployers-source_archive.adb
```diff
@@ -175,20 +174,9 @@ package body Alire.Origins.Deployers.Source_Archive is
...
-               Force_Local_Option : constant String_Vector :=
-                 (case GNATCOLL.OS.Constants.OS is
-                     when GNATCOLL.OS.Windows =>
-                        Empty_Vector & "--force-local",
-                     when others =>
-                        Empty_Vector);
             begin
                Subprocess.Checked_Spawn
                  ("tar", Empty_Vector &
-                    Force_Local_Option &
```
mosteo: Is this intended for this PR?

src/alr/alr-build_env.adb
```diff
@@ -0,0 +1,175 @@
...
+with Alire.GPR;
+with Alire.Directories;
+
+with Alr.Platform;
+with Alr.Paths;
+with Alr.OS_Lib;
+with Alr.Query;
+with Alr.Commands;
+
+package body Alr.Build_Env is
```
mosteo, line 17: I think this is one of these changes, as you're creating new files, where you can take advantage to move core functionality from `alr` the tool into `alire` the lib. What do you think?

src/alr/alr-build_env.adb
```diff
@@ -0,0 +1,175 @@
...
+
+      --  Sort and remove duplicates in paths (may come from extension
+      --  projects).
+      for Path of All_Paths loop
+         Sorted_Paths.Include (Path);
+      end loop;
+
+      if not Sorted_Paths.Is_Empty then
+         for Path of Sorted_Paths loop
+            Result := Result & Path & Path_Separator;
```
mosteo: This leaves a trailing separator, although I guess it's harmless.

src/alr/alr-commands-setenv.adb
```diff
@@ -0,0 +1,55 @@
...
+
+   ----------------------
+   -- Long_Description --
+   ----------------------
+
+   overriding
+   function Long_Description (Cmd : Command)
+                              return Alire.Utils.String_Vector is
+     (Alire.Utils.Empty_Vector
+      .Append ("Print the environment variables used to build the crate." &
```
mosteo, line 38: Missing space after this sentence's period.

mosteo (commented):
Thanks, @Fabien-Chouteau . It looks good to me. Do you plan on adding a test for the `setenv` command?

Fabien-Chouteau:
> Do you plan on adding a test for the `setenv` command?

Yes I do :)

src/alire/alire-origins-deployers-source_archive.adb
Fabien-Chouteau: Looks like something change in the GitHub Windows environment and `tar` doesn't require `--force-local` anymore. This change doesn't belong in this PR, I just wanted to have clean test results before review.

src/alr/alr-build_env.adb
Fabien-Chouteau: I was going to add this in `Alire.` but there are dependencies in `Alr.`. Do you think I should try to move everything to `Alire.`?

src/alr/alr-build_env.adb
Fabien-Chouteau: Harmless indeed, I can still try to remove it.

src/alr/alr-build_env.adb
mosteo: You're right, this might not be easy. I'd say, don't lose time over it.

src/alire/alire-origins-deployers-source_archive.adb
mosteo: Cool

mosteo (approved):
Fabien, should we wait for @pmderodat review or have you discussed this one internally?

pmderodat:
Hello

We haven’t, but I don’t have enough time to have a real look at this. Looks good to me on the principle. :-)

Issue #106, opened 2019-04-24
Substitutable dependencies


mosteo:
Alire has the ability to express dependencies that can be fulfilled by one of several projects (like OPAM does with the pipe operator). This supports at least two use cases:
- We need one of two libraries, no matter which one (e.g. some crypto, some sockets...).
- We do not want to bring in heavy dependencies if we can do with lighter ones.
  - In turn this can simplify finding solutions, since less libs are depended upon.

The first case I think is pretty simple. It appears, e.g., when there are no other conditions (like a lib is only available in windows and another on linux) on which to select one of several possible candidates.

The second case is a bit more tricky. In the old index it was tested with [gnatcoll](https://github.com/alire-project/alire/blob/be788521894b275a9361a188ff6d22ffedd16c08/src/oldindex/alire-index-gnatcoll.ads). It came in three flavors:
- gnatcoll, which depends in turn on gpr and maybe other libs
- gnatcoll.slim, which has no dependencies (and loses anything that depended on that)
- gnatcoll.xstrings, which is only the XString type from gnatcoll.

To make proper use of this, the old index had two features:
- "renaming" field, that was used by gnatcoll.slim and gnatcoll.xstrings to express that, at the time of resolution, they are equivalent to gnatcoll (i.e., once one of the three is included in a solution, the "gnatcoll" dependency is fulfilled.
- "or" operator in dependencies, that was used to express an ordered list of dependencies, only one of which was needed (in that preferred order): `gnatcoll.xstrings="1.0" or gnatcoll.slim="1.0" or gnatcoll="1.0".

Now, suppose there are projects B and C. Project B, in turn, depends on full gnatcoll. Project C, in turn, depends only on gnatcoll.xstrings. With the previous arrangement, a third project A that depends on B and C, would get a solution that can only be fulfilled by plain "gnatcoll". OTOH, a project D that depends only on C, would get the lighter "gnatcoll.xstrings" dependency:
```
A --> B --> gnatcoll
  \-> C -/

D --> C --> gnatcoll.xstrings
```

To make thinks even simpler for clients, the old project [xstrings](https://github.com/alire-project/alire/blob/be788521894b275a9361a188ff6d22ffedd16c08/src/oldindex/alire-index-xstrings.ads) encapsulated the three dependencies above, so a client interested on xstrings did not need to specify the triple OR dependency: 
```C --> xstrings --> gnatcoll.xstrings```

[...]

PR #304, opened 2020-01-20
Externals 2/4: command-line tools

mosteo:
This implements detection of version from strings emitted by an executable in the env path. Currently, this serves to fix the make and gnat crates.

In regard to the gnat crate, this improves the situation of the growing compiler enumeration in that now the compiler is a normal dependency, so the compiler version environment variable is gone (see e.g. [gnatcoll.toml](https://github.com/mosteo/alire-index/blob/5e2cccdf3a0f8bcd0d7ad5a87bd4f30e3e4657aa/index/gn/gnatcoll.toml#L49)). It still does not use gprconfig to detect runtimes/targets. That is an issue at another level for which I don't have a clear idea yet.


Fabien-Chouteau:
What I find interesting with this is that we will be able to declare dependencies on cross compilers:

`riscv32_elf_gnat.toml` or `gnat_riscv32_elf.toml`:
```
[...]
[[external]]
kind = "output-version"
version-command = [ "riscv32-gnat", "--version" ]
version-regexp = "^GNAT ([\\d\\.]+).*|^GNAT Community ([\\d]{4}).*"
```

mosteo:
Yup. I was thinking however how would we like to do with a crate available for several targets. I guess you will have better ideas than me about this:

- individual releases per target, each depending on the concrete cross-gnat. The build string of versions could fit the bill here: "1.0+native", "1.0+riscv32", and so on.
- a way to express alternative dependencies, so a single release may depend on any of several gnat_* crates (related to #106).
- ...

mosteo:
Indeed I just saw that your latest crates for the index have a list of project files, which I guess only one can be used, for each appropriate target.

Fabien-Chouteau:
I think almost all library crates will be cross-platform, the real question will be run-time profile.

The bin/app crates will have a fixed set of boards/targets they can be compiled for, so a fixed list of compiler dependency.

Issue #298, opened 2020-01-07
[RFC] Policy on unavailable dependencies


mosteo:
The issue of the dependency on AWS in alire-project/alire-index#73 prompts me to think about the stand to take on unavailable dependencies. A few thoughts:

We have a somewhat indirect way of indexing something as unavailable (e.g. [aws](https://github.com/alire-project/alire-index/blob/40d66cb7598ed25226fcf0ca8a5b968cf210fd68/index/aw/aws.toml)). A more explicit way could be #295.

Advantages of referring to these unavailable crates in dependents: 
- The user gets a hint about the need for the crate to be provided externally.
- Once the crate gets packaged, dependents will start to use it without changes.
- Already available.

Drawbacks:
- Once a dependency gets packaged, a packaged dependent might break (although arguably it was never working? Yet it's different to say "go look for this" than "take this although, surprise, it's incompatible").
- Who gets to create these missing crates? Ourselves? The contributor adding a crate that needs it for the first time? It's another barrier to entry.

Another possibility would be to define the missing dependencies explicitly in dependents (e.g. a new `externally-depends-on`, used only to keep track of those until available). 

Advantages:
- Still documents the missing dependencies and provides hinting.
- May be simpler to sell to contributors, it's just a list of dependencies.
- No breaking, since dependent crates would need explicit edition to move dependencies from `externally-depends-on` to normal `depends-on` -- at which point the testing should occur.

Drawbacks:
- Each dependent crate would require updating down the road.
- Not readily available, needs being implemented.

Of course we could simply do nothing, and document missing dependencies in e.g. comments in the crate file. Situation similar as previous one, without hinting, but with higher risk of "contaminating" the index with hidden dependencies. I would prefer to do this only as a last resort until something else is in place.

...

mosteo:
I think this is solved after #305 and later dealings with incomplete solutions in #451. 

For the record, the situation now is that we can use external hints or system packages, or even give a dependency that doesn't even exist in the index, and incomplete (complete if system packages are available) solutions will be provided by `alr`, that users can accept at their own risk (i.e., by making the missing dependencies available somehow on their own).

PR #305, opened 2020-01-20
Externals 3/4: unindexed placeholders

mosteo:
Just an empty external that still serves to get a hint about a missing dependency.

Related discussion in #298


Fabien-Chouteau:
Can an external be native on Linux but unindexed on Windows?

The example I have in mind is SDL. It is trivial to install on Ubuntu, but on Windows requires a manual installation that will not be handled by `Alire`.

Follow-up question, is it possible to display a message to the users when a dependency is not available. For instance: `On Windows ,SDL2 requires manual installation. Please follow installation instructions at https://wiki.libsdl.org/Installation#Windows_XP.2FVista.2F7.2F8.2F10`.

mosteo:
> Can an external be native on Linux but unindexed on Windows?

In the current draft we can have an array of externals, and all of them would be tried everywhere, with only the relevant ones succeeding. But, probably better, we can narrow down which ones are attempted by having the `available` field also added to the base external class. This would reduce noise in the hints emitted.

> The example I have in mind is SDL. It is trivial to install on Ubuntu, but on Windows requires a manual installation that will not be handled by `Alire`.

I see. See below how this could look.

> Follow-up question, is it possible to display a message to the users when a dependency is not available. For instance: `On Windows ,SDL2 requires manual installation. Please follow installation instructions at https://wiki.libsdl.org/Installation#Windows_XP.2FVista.2F7.2F8.2F10`.

Good idea. We add a `hint` field that gets displayed on platforms where the external is available.

So `libsdl.toml` could be something like:

```toml
[[external]] # Unindexed, fallback for Windows
hint = "On Windows, follow these instructions..."
   [external.available.'case(os)']
   windows = true
   '...' = false

[[external]] # Native, fallback for unknown distros
kind = "native"
hint = "Check available SDL packages for your distribution"
    [external.origin.'case(distribution)']
    # known packages, etc
```

After resolution, since we get the list of missing crates resolved as hints, we can print the hints of any external that is available but hasn't detected anything.

mosteo:
I'm implementing the hint idea as part of the common #302.

PR #322, opened 2020-02-18
Update documentation with external changes

mosteo:
The system crates rework done by ae81cbd introduced a few changes that were not updated in the docs. Namely, the concept of system compiler which was lost during the TOML index transition is brought back so system crates are properly [un]available depending on the compiler in use.

### DIFF
--- a/doc/catalog-format-spec.rst
+++ b/doc/catalog-format-spec.rst
@@ -500,6 +500,6 @@ Parameters
   different OS. Currently supported values are: ``debian``, ``ubuntu``.
 * ``toolchain``: takes ``system`` value in distributions with the system Ada
   compiler first in PATH (GNAT FSF in Debian/Ubuntu), ``user`` otherwise
-  (GNAT CE editions, other cross-target toolchains).
+  (GNAT Community editions, other cross-target toolchains).
 * ``word-size``: architecture word size. Currently supported values are:
   ``bits-32``, ``bits-64``, ``bits-unknown``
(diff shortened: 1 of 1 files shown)


doc/catalog-format-spec.rst
```diff
@@ -482,14 +482,24 @@ can accept dynamic expressions in this context):
...
   and ``windows``.
 * ``distribution``: name of the Linux distribution, or ``none`` if running on a
   different OS. Currently supported values are: ``debian``, ``ubuntu``.
-* ``compiler``: name of the current compiler. Currently supported values are:
-  ``gnat-unknown``, ``gnat-fsf-old``, ``gnat-fsf-7.2``, ``gnat-fsf-7.3``,
-  ``gnat-gpl-old``, ``gnat-gpl-2017``, ``gnat-community-2018``.
-* ``word-size``: architecture word size. Currently supported values are: ``bits-32``, ``bits-64``, ``bits-unknown``
+* ``toolchain``: takes ``system`` value in distributions with the system Ada
+  compiler first in PATH (GNAT FSF in Debian/Ubuntu), ``user`` otherwise
+  (GNAT CE editions, other cross-target toolchains).
```
Fabien-Chouteau: It is better to spell out GNAT Community, I don't think everyone know what CE means.

Fabien-Chouteau (changes requested):

doc/catalog-format-spec.rst
mosteo: Right, done. Thanks for the reviews!

PR #324, opened 2020-02-19
Fix bugs in `alr test`

mosteo:
* Child output was not logged on Checked_Error
* Some regular child failures were reported as errors in `alr test` itself
* Crates without regular releases broke `alr test`

### DIFF
--- a/src/alire/alire-os_lib-subprocess.ads
+++ b/src/alire/alire-os_lib-subprocess.ads
@@ -19,6 +19,15 @@ package Alire.OS_Lib.Subprocess is
       Understands_Verbose : Boolean := False;
       Err_To_Out          : Boolean := False) return Utils.String_Vector;
    --  Either suceeds or raises Checked_Error with the code and output as info.
-   --  Output is captured and returned.
+   --  Output is captured and returned on success.
+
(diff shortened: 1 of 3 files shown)


src/alire/alire-os_lib-subprocess.ads
```diff
@@ -19,6 +19,15 @@ package Alire.OS_Lib.Subprocess is
       Understands_Verbose : Boolean := False;
       Err_To_Out          : Boolean := False) return Utils.String_Vector;
    --  Either suceeds or raises Checked_Error with the code and output as info.
-   --  Output is captured and returned.
+   --  Output is captured and returned on success.
+
+   procedure Checked_Spawn_And_Capture
```
Fabien-Chouteau, line 24: I am not sure why you are making a procedure version of this function.

Fabien-Chouteau (approved):

src/alire/alire-os_lib-subprocess.ads
mosteo: This is to be able to log the partial output of the spawned command in case of failure.

Fabien-Chouteau: ok I understand

PR #332, opened 2020-03-08
Fix spurious warnings on MacOS

### LINKED ISSUE #329: Spurious "cat: /etc/os-release: No such file or directory" on MacOS
When running e.g. `alr run --list`

mosteo:
For some unclear reason the body of `alr-platforms-linux.adb` seemed to be in use when compiling for MacOS instead of the proper one, causing #329.

We relied on both OS and ALIRE_OS to select the host OS. Each of these used different names for each OS. Furthermore, we needed to explicitly set OS in MacOS but not in other OS. I think the bug was caused by some mix-up with these two variables.

With this patch, one must explicitly set ALIRE_OS only, which defaults to `linux` (as before). The project files now use a typed enumeration instead of a free-form value. The build scripts have been updated to match.

Fixes #329.

- [x] Test pending

### DIFF
--- a/alire.gpr
+++ b/alire.gpr
@@ -15,8 +15,8 @@ library project Alire is
 
    case Alire_Common.Host_Os is
       when "windows" => Src_Dirs := Src_Dirs & ("src/alire/os_windows");
-      when "osx"     => Src_Dirs := Src_Dirs & ("src/alire/os_macos");
-      when others    => Src_Dirs := Src_Dirs & ("src/alire/os_linux");
+      when "macos"   => Src_Dirs := Src_Dirs & ("src/alire/os_macos");
+      when "linux"   => Src_Dirs := Src_Dirs & ("src/alire/os_linux");
    end case;
 
    for Source_Dirs use Src_Dirs;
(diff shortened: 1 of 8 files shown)


mosteo:
Hello @simonjwright, since you worked out the MacOS port, maybe you want to take a look at this one.

simonjwright:
> Hello @simonjwright, since you worked out the MacOS port, maybe you want to take a look at this one.

Looks good to me (mind, I couldn’t reproduce #329)

README.md needs updating too.

mosteo:
> Looks good to me (mind, I couldn’t reproduce #329)

And rightly so, it was a problem in my setup script. 

> README.md needs updating too.

Good catch.

Since there's no actual need for this patch, now I wonder if the removal of the `OS` setting step is worth the merge. Thoughts?

simonjwright:
> Since there's no actual need for this patch, now I wonder if the removal of the `OS` setting step is worth the merge. Thoughts?

I’d have thought not. Two (much more popular) OS’s work without user involvement.

mosteo:
OK, let's not stir things needlessly then.

PR #336, opened 2020-03-24
Windows msys2 support

Fabien-Chouteau:
Msys2 is a software distribution platform for Windows (system package manager).

It can provide easy install of external dependencies like we would see on Linux distribution.

With this change msys2 is automatically installed in the `.cache/alire` directory.


mosteo:
Hi @Fabien-Chouteau , I'm seeing changes about Pacman. Should these be here?

Fabien-Chouteau:
> I'm seeing changes about Pacman. Should these be here?

msys2 uses the Pacman project from Arch Linux to install packages.
The code should be reusable on Arch BTW.

mosteo:
Ah, that's great then. Thanks for the clarification.

src/alire/alire-origins-deployers-system-pacman.adb
```diff
@@ -0,0 +1,118 @@
...
+            Trace.Detail
+              ("Candidate version string: "
+               & Package_Line (Matches (1).First .. Matches (1).Last));
+            return
+              Version_Outcomes.New_Result
+                (Semantic_Versioning.Parse
+                   (Package_Line (Matches (1).First .. Matches (1).Last),
+                    Relaxed => True)); -- Relaxed because some pacman
+                                       --  versions have extra version dash,
+                                       --  e.g. 1.2.3-4
```
mosteo: If the extra part after the dash does not conform to semantic versioning interpretation we should ignore it instead of parsing it. The thing is that in Semver a version with dash is a pre-release version, so `1.2.3-4 < 1.2.3`

testsuite/tests/index/external-msys2/my_index/index/di/dialog.toml
```diff
@@ -0,0 +1,10 @@
+[general]
+description = "Just for the test"
+maintainers = ["alejandro@mosteo.com"]
+maintainers-logins = ["mosteo"]
```
mosteo, line 4: Could this confuse us down the road if I didn't write this test?

mosteo (commented):
Thanks Fabien, this will surely be a major step. I have only a couple of observations.

testsuite/tests/index/external-msys2/my_index/index/di/dialog.toml
Fabien-Chouteau: I just copied from another test. We can use `example@example.com` and `example` instead.

src/alire/alire-origins-deployers-system-pacman.adb
Fabien-Chouteau: Indeed that's not what the hyphen part means in Pacman. I will change that.

Issue #227, opened 2019-10-24
Define a publishing workflow, possibly with `alr publish` support.


mosteo:
For reference, the opam docs:
- https://opam.ocaml.org/doc/Packaging.html
- https://github.com/ocaml/opam-publish

I've received [here]() a reply about using tagged releases which is not entirely reassuring. Still, opam is using these releases. Let's hope the reply comes from an extra- person.

mosteo:
Now I got a clear [answer](https://github.community/t5/How-to-use-Git-and-GitHub/Ensuring-identical-tarballs-from-releases/m-p/35808/highlight/true#M9217). For whatever reason they're keeping a escape clause open, so we should at least make a recommendation to generate your own tarball, and whether possible automate it.

Fabien-Chouteau:
Should we consider hosting the archives in the index repo or another repo?

I don't know what the size limit is github.

mosteo:
> Should we consider generating the archives with alr publish and hosting them in the index repo or another repo?

I don't think we can allow users to do so; at most we could help with creating the release/uploading the tarball to the user's repo (if it's in github). And we should be careful to avoid that we are spammed with automated PRs.

> I don't know what the size limit is github.

According to https://help.github.com/en/github/administering-a-repository/about-releases there's no limit on the global size of releases; only 2GB per file.

Also for reference, here's the release API: https://developer.github.com/v3/repos/releases/

Fabien-Chouteau:
> I don't think we can allow users to do so;

Why not?

> at most we could help with creating the release/uploading the tarball to the user's repo (if it's in github).

Maybe with this indeed: https://developer.github.com/v3/repos/releases/

> And we should be careful to avoid that we are spammed with automated PRs.

I am not sure this is a problem considering that creating a PR requires a GitHub account. I think GitHub already has an anti spam account protection.

> > I don't know what the size limit is github.
> 
> According to https://help.github.com/en/github/administering-a-repository/about-releases there's no limit on the global size of releases; only 2GB per file.

I mean if we host the archives in a GitHub repo, as checked-in files. How big the repo can be? How many archives can we host?

mosteo:
[...]

PR #349, opened 2020-04-17
Split single upstream source into several crates

mosteo:
This is a new attempt at child crates, with well-defined semantics and much less invasive changes. It addresses only one use case, that manifested e.g. in alire-project/alire-index#85, which is when we want to split a single source archive or checkout into several crates (possibly because they have different dependencies, or it makes logical sense in a large project).

In the Matreshka case the problem was that there is a single upstream repository. Splitting it into several crates required preparing extra artificial tarballs, which furthermore use fake project files to bring in projects from the main crate, with a name that really should not be exposed to users.

This PR introduces a new origin syntax, which points the crate source to another crate:
```
origin = "crate:parent_crate"
```
A release using such an origin has an implicit dependency on parent_crate at the same version as itself. This dependency must not be provided again by the packager.

What we gain:
- No need to prepare separate tarballs/repos for each child crate.
- No need to have proxy project files with a workaround name in the child crates.
- Ability to have different dependencies for project files within a single upstream source.

This PR relies on a temporary index branch where the Simple Components have been indexed this way, as an example. See, e.g., [simple_components_openssl.toml](https://github.com/mosteo/alire-index/blob/devel-0.3/index/si/simple_components_openssl.toml)

Internally, what changes is that the project paths of a child crate are not relative to its checkout folder (which is empty) but to the parent crate (see the [test](https://github.com/mosteo/alire/blob/031667045dacd356a2faa5e3a7aef3b99c1bda79/testsuite/tests/get/child-crate/test.py)).


mosteo:
@reznikmm, if you have the time to see if this one address your needs... Thanks.

mosteo:
@stcarrez , from what I read in alire-project/alire-index#73, this might also interest you.

mosteo:
My doubt is if imposing the matching parent-child version makes sense in all cases, or if it could be relaxed to any arbitrary explicit version, e.g.:
```
origin = "crate:parent_crate=2.0"
```

Fabien-Chouteau:
I have doubts about this feature. I think I understand the benefits of the ability to have different dependencies for child crates (even if right now `simple_components` and `matreshka` don't have any dependencies that I can see).

But I feel like the existence of those those big projects that do so many different things is a consequence of not having a package manager. So my feeling is that the package manager should not encourage such big heterogeneous projects.

On the other hand we cannot ignore those libraries in the ecosystem.

I wonder if it is not better to have the big libraries as a single crate anyway, even if it means that you will pull more dependencies than you really need.

mosteo:
I share the same concerns.

> simple_components and matreshka don't have any dependencies that I can see

simple_components has, depending on the parts you use, dependencies on odbc, gnutls, libssl, apq, gnade. If you mean non-external dependencies, only apq and gnade are likely to eventually make into alire.

I have not checked the criterion followed to split Matreshka, but I suspect there are unstated dependencies (which is a different problem).

> I wonder if it is not better to have the big libraries as a single crate anyway, even if it means that you will pull more dependencies than you really need.

This seems the simplest no-op solution. I'm just wary of non-obvious situations, like a post-fetch build process that fails for parts that you don't need (because of missing externals, for example). Not a problem for simple_components, which has no build process.

@stcarrez said in alire-project/alire-index#73: 
> don't put constraints on upstream projects. There are many reasons why such upstream separation may not be possible.

I wonder what reasons he had in mind, besides the obvious inconvenience.

stcarrez:
> I have doubts about this feature. I think I understand the benefits of the ability to have different dependencies for child crates (even if right now `simple_components` and `matreshka` don't have any dependencies that I can see).
> 
> But I feel like the existence of those those big projects that do so many different things is a consequence of not having a package manager. So my feeling is that the package manager should not encourage such big heterogeneous projects.
> 
> On the other hand we cannot ignore those libraries in the ecosystem.
> 
> I wonder if it is not better to have the big libraries as a single crate anyway, even if it means that you will pull more dependencies than you really need.

Again, don't put constraints on upstream projects.

There are many reasons why a project may be managed as a single source project tree
and still produce several libraries. This applies to `simple_components`, `matreshka`,
`ada-util`, `ada-ado`, `ada-servlet`, `ada-asf`, and all the Ada Web Application framework.
In the Debian packaging world, a single source tree can produce several Debian packages.
Likewise for FreeBSD. I see nothing wrong in the way these authors manage their project. It is
their own choice.

As far as I'm concerned, some reasons I will not split further some of my projects:

* Spliting increases the complexity of a project. Sometimes it is not worth to increase the complexity of a project to isolate a small subset of it. Still the mainstream project allows their users to do it by a good organization of the their GNAT projects.
* Sometimes this is impossible without breaking the build of the mainstream project or without loosing some functionality.
* A typical example in my case is a unit testing framework addon library that is built on top of the component itself and used for the unit testing of the component. And this small unit testing framework is provided as a separate Alire package because it can be (and is) used by other projects. You can't move it to another project easily because this would break the unit tests of the library itself. There are other examples like that.

Instead of adding such crate origin, would it be possible to make alire more intelligent
by identifying the common origins and extract the mainstream project only once?

mosteo:
Thanks for the further feedback, Stephane. There's good food for thought there, and several pointers to follow.

> Instead of adding such crate origin, would it be possible to make alire more intelligent
by identifying the common origins and extract the mainstream project only once?

Perhaps we could do something at the time of publishing with support from `alr publish`. It can be more complex for maintainers (for complex projects) I guess, but it might be the best-rounded solution.

For example, identify sources needed by a project file, to create an initial manifest that can be hand-tweaked. Of course scenarios will make things less ideal. But I see how this can work. Especially if we already had an eye on autopackaging and uploading tarballs to ensure binary immutability as part of the publish workflow (#227).

reznikmm:
Matreshka has different dependencies for some "sub-libraries". The `spikedog` server requires `AWS`, while core library doesn't. SQL database drivers (not in alire yet) requires corresponding native development package (for instance postgresql).

So, I think, current PR should work with this, but I haven't tried it yet.

mosteo:
I'm going to close this one since after reading all the comments I'm convinced it's better to act at packaging time. Since one of the priorities is to improve the publishing workflow I hope we will revisit this topic sooner that later.

reznikmm:
So, for Matreshka I still need to provide dummy archives with trivial project files, right?

mosteo:
Yes. A possibility would be to prepare custom scripts to split the sources+projects into different crates so the archives aren't dummy. I'm unsure how much work that would entail compared to the dummy archives. A problem with the dummy archives is that the dummy project file names are exposed to users, and that can cause confusion and future incompatibility if they use them mistakenly.

I still have to research how Debian does this, but I expect our solution down the road may be with helping in this direction (detailing in the toml manifests the parts that go into the crate).

PR #350, opened 2020-04-19
Allow dots in crate names

mosteo:
Although currently on top of #349 (because I reuse the Simple Components crates), this PR is totally independent.

This PR is a new attempt at reintroducing dots in crate names:
```
$ alr list simple
simple_components          Simple Components - A collection of support packages
simple_components.apq      Simple Components - APQ                             
simple_components.gnade    Simple Components - GNADE                           
simple_components.odbc     Simple Components - ODBC                            
simple_components.openssl  Simple Components - SSL                             
simple_components.secure   Simple Components - Secure connections              
simple_logging             Simple logging to console
```
The purpose is to have unambiguous space (`_`) and "specialization" separators (`.`). In this proposal, this is purely cosmetic and has no implications about crate definition reuse or dependencies. Each crate is, as currently, its own entity. Obviously, it could make sense to use #349 with this one, if one wants.

To mimic Ada hierarchies and GNAT naming convention, the `simple_components.apq` crate is described in file `simple_components-apq.toml`, and so on.


Fabien-Chouteau:
In my opinion having dots would only be useful if it specifies child crates. Just for cosmetic I don't think this is required, and even in that case I would prefer dash (`-`) over dot (`.`).

mosteo:
> I would prefer dash (`-`) over dot (`.`).

Isn't that a bit strange in the context of Ada?

stcarrez:
> > I would prefer dash (`-`) over dot (`.`).
> 
> Isn't that a bit strange in the context of Ada?

Dash being often used for package separation within file names, I also prefer it too.

Fabien-Chouteau:
> Isn't that a bit strange in the context of Ada?
I don't think it is necessary that Alire package name follow the Ada identifier syntax.

mosteo:
No, of course. My reasoning is that Ada programmers may expect Ada-like syntax, but clearly I'm in the minority already :-)

Fabien-Chouteau:
There is a problem with the dash and dot anyway, they have a meaning for gprbuild. And it's nice to use the same name for the crate and the project file. So I think we should stick to only underscore to keep it simple.

mosteo:
I had to work around that for this PR. All right, closing.

Issue #251, opened 2019-11-29
[RFC] Do not update dependency versions unless requested


mosteo:
Arising from my original misunderstanding of some cargo/opam features, our current implementation defaults to updating dependencies by default, any time a dependency changes (addition/removal), or when `build` is called (but see #250).

Both opam/cargo stay on the initial dependency versions until the user request an upgrade of dependencies (re-resolving them with new index contents):
- In opam: `opam update` (to update index) followed by `opam upgrade`
- In cargo: index update is automatic on `rust update`, `rust update` updates dependencies

I think from a principle of least surprise, we should operate likewise (and so the `alr update` makes more sense). Our two step update process would be:
1. `alr index --update-all`, to update the index (already implemented)
1. `alr update` inside a working copy, to update its dependencies (unless pinned).
    - We need to know the current solution in use, via a lockfile or other method. This is not currently available, hence the use of the solver and the automatic update when this info is needed.   

Related notes:
- Cargo stores the current solution in `Cargo.lock`. This file is not intended to be edited by users. The allowed version sets come from `Cargo.toml`, which can be edited manually. 
   - In alr we likewise have the allowed version sets in the toml file. We do not have yet a lockfile or equivalent, although the information is present in the downloaded dependencies toml files.
   - A lockfile has the advantage that it can be published with binary crates, to ensure reproducible builds with the same dependency solution (like in cargo), while allowing update of dependencies on demand.
- Opam uses `pin` to freeze versions even across updates. It also uses `pin` to force use of a local directory instead of a downloaded source. 
   - `alr pin` currently freezes dependencies by modifying the toml file. This implies the local loss of the originally allowed version sets, precluding future updates. We should probably act like rust in this regard, keeping the toml file as a mirror of what is in the index, and only modifying it if the modification should go back to the upstream index at publishing time.
   - To use a local dir for a crate, we can now add an index where the origin is the local folder. We could automate this somehow through `alr pin`, a-la opam, just like `alr with` avoids manual edition of the toml file.
      - The information in the lockfile for pinned releases might make the need of an extra index obsolete for this use case.

Issue #240, opened 2019-11-19
Package development workflow


jklmnn:
I started developing a package for Alire and initialized the repo with `alr init --lib --in-place <name>`. I noticed that adding a dependency requires a version to be set, yet a version doesn't exist yet in this project. So I created the entry for the next version but it required an origin. Since I didn't create a release yet on GitHub I could only set the origin to something, e.g. the master tarball. But even then there is a circular dependency as I can never have the version in the toml file in the repo and the actual release on GitHub in a consistent state as I would have to guess the hash for the release tarball before crafting the release.
I then looked into other project and noticed that they didn't incorporate the alire directory into the source tree. I wonder what the desired workflow is then as the only developer provided information seems to be the toml file in the index?
I somehow miss the functionality to develop (and in this process specify dependencies) and then to publish (an already created) release independently.

mosteo:
You're right that this is an area that is sorely lacking information (and automation). Even so, I think it should be doable without circularities. I will prepare some instructions ASAP.

pmderodat:
Any news on this? I stumbled on the same issue. :-) Thanks!

mosteo:
Ooops this had been buried in my to-do stack. Thanks for the heads-up.

mosteo:
I think I'm missing some piece of the puzzle. @jklmnn, if I understand correctly you want to develop in an unreleased project, while using dependencies from alire. At a later time you might want to publish this development with a proper version. This is what I did to tests this use case:

1. `alr init --bin xxx` (This should be equivalent to your in-place initialization)
1. `alr with libhello`
   - This adds a dependency on libhello. The `alire/xxx.toml` file is modified accordingly by `alr` (you could do this manually if you prefer). 

Your working release (`xxx` here) has version 0.0.0 when initialized. This does not really matter (yet). 

I may be missing something obvious. The *"I noticed that adding a dependency requires a version to be set, yet a version doesn't exist yet in this project"* is throwing me away. The version is needed once you want to publish your release, but not yet. 

If I'm following correctly, after the above steps let's say you want to publish xxx v1.0. At that time, in essence, you need to take the `alire/xxx.toml` file and use it as the basis for the PR you'd submit to the index. It's then when you can replace the 0.0.0 version with the proper one, and the filesystem origin with the online URL of the zipped `xxx` (without the `alire` folder, which is recreated on download by `alr`) (we may need a toml pretty-printer at some point to help with this).

Please let me know if this helps.

[...]

Issue #237, opened 2019-11-14
Recursive upgrading project dependencies


dmitry-kazakov:
The objective is walking the dependencies list and switching to the latest available version of each keeping the list consistent. Example. Let be 4 projects. A, B, C, D. Bv1 depends on Av1. Cv1 depends on Av1. Dv1 depends on Bv1 and Cv1. Now there is Av2 and Bv4 dependent on Av2 etc. The objective is to upgrade D to the latest versions of A, B, C. Options would be upgrading C to Av2 or rejecting upgrade of D since there is no version of C working with Av2.

mosteo:
If I understand what you mean correctly, currently there is no management of individual versions: whenever the list of dependencies is needed (`alr get`, `alr build`, probably in some other cases I'm missing right now) the solver is invoked and it will find a consistent solution, if there is one. The current solver is exhaustive and will attempt all combinations in the worst case, so a dependency might be resolved as a lower version than available if another dependent requires it (as in your example IIUC). 

Dependencies are attempted by default from newest to oldest and the objective is to find a consistent solution, not individually managing the dependencies (e.g. I want to only upgrade C is not directly supported right now). The latter would be a new use case.

Currently the solver is invoked every time the list of dependencies is needed. I had in the radar the idea of caching this resolution so the solver is only called when explicit upgrades are requested. This cached list would make easy to upgrade only a subset of dependencies, considering the rest as "pinned".

But, I'm not sure if that's what you want. Perhaps what you want is already happening with:

1. `alr index --update-all`
1. `alr update`

The first retrieves new crate versions, and the second call will re-resolve dependencies, upgrading whatever can be upgraded. Right now, to have finer control of versions, one has to `alr pin`, which will freeze all versions in place, and manually edit `<crate>/alire/crate_name.toml` to allow different version sets therein.

dmitry-kazakov:
Then it is another issue. There should be possible to pull a certain version of a package with all dependencies corresponding to that version. Pulling the latest versions is inconsistent. A version of a package must always refer to a definite version of the packages it depends on. It cannot be open-ended. Only version equality should be allowed. So, let the package B depend on A. Then it is always like Bv1->Av2, Bv2->Av5, Bv3- Av5, Bv4->Av7 etc. When Bv2 is pulled, then Av5 pulled with it, not Av7.

mosteo:
This is doable by specifying exact versions in the release dependencies. The option to use version sets as dependencies is up to the publisher of a release.

Issue #149, opened 2019-07-09
alr pin --remove option to allow unpinning


mosteo:
It seems polite to have a way of unpinning what can be pinned.

PR #355, opened 2020-04-26
Cache dependency solution in a lockfile

mosteo:
I'm using the term "lockfile" to follow the Cargo nomenclature. In practice, this means having the dependency solution exported to disk so we don't resolve every time the closure is needed.

A good explanation of the purpose of this file for Cargo is [here](https://doc.rust-lang.org/cargo/guide/cargo-toml-vs-cargo-lock.html). Summarizing:

- The `crate.toml` file stores the crate description, with unresolved dependencies. This file is modified manually or through alire non-destructively, and eventually makes into the index as a new release description. This is already implemented.
- The `crate.lock` file stores the actual closure of releases that fulfill all transitive dependencies. It's not edited manually, but created on first solving of dependencies and recreated when dependencies are updated, pinned, etc.

This PR implements the exporting of the solution we already use to the lockfile. There is no new functionality other than that. Low-hanging fruit that can be built on top of this:

- Do on-demand updating of dependencies (#251, #237)
- Pin dependencies properly (without losing the original dependency; related: #149)
- Pin dependencies to local folders (to use WIP crate dependencies, #240)

@Fabien-Chouteau, asking for your opinion on this one. You can test it by retrieving any crate and the lockfile will be in the alire folder.


Fabien-Chouteau (approved):
My only comment is that loading and saving of solutions might me implemented in a child package to avoid having a big `Alire.Solver` package.

Fabien-Chouteau:
How difficult do you think loading the solution will be?

mosteo:
I should be trivial. We already do it for releases and dependencies, which are the two things appearing in the file.

mosteo:
> My only comment is that loading and saving of solutions might me implemented in a child package to avoid having a big `Alire.Solver` package.

OK. I will add loading after the refactor so we have the complete save/load cycle.

mosteo:
I just pushed the loading of the lockfile, and as proof of concept this is used when showing a local crate. I've tested it with ada_voxel_space_demo:

    alr get ada_voxel_space_demo
    cd ada_voxel*
    alr show --solve # This uses the solution stored in the lockfile

Next steps would be to use it too in build/setenv and any other places that currently use the solver to get a complete solution, but I'll stop here until we agree this serves us properly.

mosteo:
I have structured changes related to lockfiles in a few PRs. This one now contains the core implementation, but nothing is done with the lockfile yet.

There are no new tests, but some are already updated to check the existence of the lockfile. PRs that build on top of this one will similarly exercise the contents of the lockfile.

mosteo:
I just realized that the PR was already approved. I'll wait to merge it for the review of the dependent ones, so we don't have a do-nothing feature in.

PR #356, opened 2020-04-26
Minor fixes

### LINKED ISSUE #328: Emphasizing --recursive in the README instructions
This is really small stuff, nevertheless...

This morning I cloned alire and I did not pay attention to the --recursive option ("Oh, yes, I need to clone it... usual stuff..")  in the git clone command.  The result was a sequence of "project not found."

I modified the README adding a warning about the necessity of --recursive.  I also created pull request #327  with the change.

mosteo:
* Eliminate the Depends/Dependencies duplicity in Alire.Releases
* ~Downgrade a debug message that was at Info level and polluting normal operation~
* Add a warning about `--recursive` in README (fixes #328)

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,7 @@ To install run the following as user in a terminal, or see below for more detail
 
 1. If running on macOS, `export OS=macOS`
 1. `git clone --recursive https://github.com/alire-project/alire.git`
+    1. Note the **--recursive** flag; otherwise compilation will fail.
 1. `cd alire`
 1. `gprbuild -j0 -p -P alr_env`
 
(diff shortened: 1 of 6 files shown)


src/alire/alire-externals-lists.adb
```diff
@@ -13,7 +13,7 @@ package body Alire.Externals.Lists is
                     return Containers.Release_Set
    is
    begin
-      Trace.Info ("Looking for external crate: " & (+Name));
+      Trace.Detail ("Looking for external crate: " & (+Name));
```
Fabien-Chouteau: I actually changed that on purpose when working on `msys2` because the operation for external crate was taking a long time and it felt as a user that `alr` was stuck. I am talking several minutes here to download and install dependencies in `msys2`.

As a user I prefer to have some information that something is going on.

mosteo: Ah, I see. I wondered how I had forgotten about that making into the tests. All right then.

mosteo:
Updated without the debug log downgrade.

Fabien-Chouteau (approved):

Issue #313, opened 2020-02-05
Automatically add `with` line in the project file when adding a dependency with `alr with`


Fabien-Chouteau:
This is a very interesting suggestion that we had a FOSDEM and I think we should do it.

The idea is that currently when the users do `alr with ada_toml`, they will also have to edit their gpr file to add `with "ada_toml.gpr";`. It would be way more user friendly to automate this.

I think we have all the information required because we know the project file of the current/local `Alire` project that needs to be edited, and we also have the list of project files from a dependency that have to be added. 

For project that needs to be extended like `GESTE`, we can potentially offer an option to deactivate this feature from the toml file.

And for users that don't want their project file to be modified, we can have command line option to deactivate as well.

stcarrez:
It would be nice if the with command prints the GNAT projects that the new library provides.

Fabien-Chouteau:
@mosteo Do you have an idea where I could start to implement this?
How do I get the list of project file dependencies from a crate that is "withed"?

mosteo:
You need to first solve the new dependency to get the recursive solution. This is done with https://github.com/alire-project/alire/blob/123fd7ca303ddb16ed6d2a6f3a5c0d118b814afa/src/alr/alr-query.ads#L110

Then you can iterate over the releases in the solution to get their project files.

Fabien-Chouteau:
Seems like it would be easier to have all the `with` between special comments in the project file and rewrite all of them every time.

Fabien-Chouteau:
Here are some of the results of my investigations on the subject:

This is a little bit more difficult to implement than I expected. I couldn't find a way to get just the list of project files for the new dependency.

The main problem I see, is the handling of all the possibilities of project files. For instance:
 - Some project are not supposed to be withed, but extended (e.g. GESTE). So an automatic `with` will result in an error.
 - Some crates will provide more than one project file, and those project files are potentially not supposed to be used together. E.g. ada-lvgl provides a project file for each size of pixel possible (8bit, 16bit, 32bit). Again, an automatic `with` will result in an error.

In my opinion, the options are: 
 1. Delay this feature for a future release, and rely on tutorial to explain that `alr with` is not enough to use a dependency.
 2. Add a key in the manifest for maintainers to declare if the project files should be automatically added or not (probably with a default to yes).
 3. Consider that the the corner cases will not cause more harm than good. And implement the feature anyway.

mosteo:
> I couldn't find a way to get just the list of project files for the new dependency.

[...]

PR #365, opened 2020-04-29
Customizable crate tests

mosteo:
From the latest comments in #313 it seems that not all crates will be able to be tested by simply building their declared project files. While this remains the default test, this PR introduces a new test action that can be used to execute alternative custom commands.

The idea behind `alr test` is not to so much to test the crates themselves (which can have more comprehensive test suites Alire doesn't care about) but to have a way to ensure that a crate builds within the Alire ecosystem and is not unknowingly broken/unresolvable in some of our testing platforms.


src/alr/alr-commands-test.adb
```diff
@@ -102,56 +106,158 @@ package body Alr.Commands.Test is
...
-                         Empty_Vector &
-                         "get" &
-                         "--build" &
-                         "-d" &
-                         "-n" &
-                         R.Milestone.Image;
+            Docker_Prefix : constant String_Vector :=
+                              Empty_Vector
+                              & "sudo"
+                              & "docker"
```
Fabien-Chouteau, line 121: Why do we need docker here?

Fabien-Chouteau (approved):

src/alr/alr-commands-test.adb
mosteo: It is a way of testing locally in the same environment used by the Linux CI github workflow, instead of in whatever local environment one has. Convenience, mostly. If only Windows/macOS could be run cross-host...

PR #388, opened 2020-05-05
Convert catalog format spec to markdown

Fabien-Chouteau:
So it can be used in the website.


doc/catalog-format-spec.md
```diff
@@ -0,0 +1,495 @@
...
+To create atomic values in an environment-dependent way, use the following
+construct (here to create a boolean):
+
+```toml
+{'case(distribution)' = {
+    'debian|ubuntu': true,
+    '...': false
+}}
+
+# Or in a more idiomatic TOML syntay
```
mosteo: errata

Issue #345, opened 2020-04-09
System externals to specify extra paths (C_INCLUDE_PATH, LIBRARY_PATH, PATH)


Fabien-Chouteau:
In the context of using `sldada` with `SDL2` from `msys2` on Windows I got into a problem that is not very easy to describe, it deals with include paths and `sdl2-config`.

At the moment, `sdlada` has three C source files that include headers from `SDL2`. The includes look like this: `#include <SDL.h>`. The path to the include is supposed to be provided by `sdl2-config --cflags` which returns options like `-I/usr/include/SDL2`.

With `msys2` the `sdl2-config` return a UNIX path relative to the "context"/"world" of `msys2`: `/mingw64/include/SDL2`. This is hardcoded and doesn't depend at all on where `msys2` is installed.

The problem is that a compiler not running in the `msys2` context will not be able to understand this path, so the includes cannot be resolved.

One potential solution I see for this problem is to allow external crate to specify include paths depending on the distribution. For instance:
```
[[external]]
kind = "system"
    [external.origin.'case(distribution)']
    'msys2' = ["mingw-w64-x86_64-SDL2"]
    [external.c_include_path.'case(distribution)']
    'msys2' = ["$MSYS_ROOT/mingw64/include/SDL2"]
```
Alire would then add this path to the C_INCLUDE_PATH environment variable before build a project.

(The `$MSYS_ROOT` keyword can be optional and all paths considered as relative to the root dir of the distribution)

@mosteo, @pmderodat what do you think of this?

mosteo:
Two points you can confirm for me, to see if I'm not confusing something:

- Do I understand correctly that `sdl2-config` is not really called with this solution? Could that bite us in the future with some other library that requires a `-config` script? My gut feeling is not a problem, but I'm not sure.
- IIUC, `$MSYS_ROOT` will be replaced by `alr`, which detects where it is installed?

With these caveats in my understanding:

Would we be more future-proof by having something general that sets/appends/prepends to environment variables? (which reminds me: there is an undocumented property that prepends items to PATH (#178). IIRC it's broken after the index migration. Perhaps we could build on that?)

If you see that's overkill, I trust your better understanding of the issue.

Either way, loading dynamic properties is a somewhat convoluted part right now, so I could tackle it if you prefer and simplify the internals in the process (#323).

Fabien-Chouteau:
[...]

PR #389, opened 2020-05-05
Load `environment` properties

mosteo:
Per #345, this PR implements the loading part of these properties. They can be inspected as any other property with `alr show`.

Properties can be retrieved in code with the new `Alire.Releases.Environment` function.


doc/catalog-format-spec.rst
```diff
@@ -285,6 +285,32 @@ entries:
...
 
+* ``environment``: optional table used to modify environment variables that
+  will apply at build time. Variables and values are specified with the form
+  ``VARIABLE.<action> = "value"``, where ``<action>`` is one of ``append``,
+  ``prepend``, or ``set``. For instance:
+
+  .. code-block:: toml
+
+   [environment]
+   C_INCLUDE_PATH.append = "/usr/include/something"
```
Fabien-Chouteau, line 296: Is the case maintained here? `PATH` vs `path`.

doc/catalog-format-spec.rst
```diff
@@ -285,6 +285,32 @@ entries:
...
+   PATH.prepend = "$CRATE_ROOT/bin"
+   MSYS.set = "$MSYS_ROOT"
+
+  Other environment variables may appear in the value, and will be replaced.
+  Furthermore, a few predefined variables are provided by Alire:
+
+  * ``$CRATE_ROOT`` is the deployment directory of the crate.
```
Fabien-Chouteau: We should discuss the predefined value format. I think this `${CRATE}` will be easier to handle and less error prone.

doc/catalog-format-spec.rst
mosteo: Yes, TOML keys are case sensitive and I did no modification on these.

doc/catalog-format-spec.rst
mosteo: Sure. At the moment there's no code about that anyway, so it's a matter of expressing what we want in the documentation. I too like better with `{}`.

Fabien-Chouteau (approved):

PR #393, opened 2020-05-08
Rely on lockfile for `alr show` of local crate

mosteo:
This way we report the actual solution on disk and not the one calculated on the fly from root dependencies.

Only the last commit is new (this was originally part of PR #355, so it has already been reviewed).

### DIFF
--- a/src/alr/alr-commands-search.adb
+++ b/src/alr/alr-commands-search.adb
@@ -29,7 +29,7 @@ package body Alr.Commands.Search is
       ------------------
 
       procedure List_Release (R : Alire.Releases.Release) is
-         package Solver renames Alire.Solver; use Solver;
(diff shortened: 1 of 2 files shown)


src/alr/alr-commands-search.adb
```diff
@@ -29,7 +29,7 @@ package body Alr.Commands.Search is
       ------------------
 
       procedure List_Release (R : Alire.Releases.Release) is
-         use Alr.Query;
+         package Solver renames Alire.Solver; use Solver;
```
Fabien-Chouteau: `use` should be on a separate line.

mosteo: Corrected.

Fabien-Chouteau (approved):

PR #394, opened 2020-05-08
Rely on lockfile for `alr setenv`

mosteo:
This way externals are not attempted to be detected every time.

Only the last commit is new (this was originally part of PR #355, so it has already been reviewed).


mosteo:
I've moved the `workflow/setenv` test to `setenv/basic`, and added a `setenv/with-external` that tests the output when there are external dependencies.

src/alr/alr-commands-setenv.adb
```diff
@@ -14,16 +14,7 @@ package body Alr.Commands.Setenv is
...
-      --  would prevent directly sourcing the output. TODO: remove this once
-      --  the lockfile is used, that will make this unnecessary.
-      declare
-         Old_Level : constant Simple_Logging.Levels := Alire.Log_Level;
-      begin
-         Alire.Log_Level := Simple_Logging.Always;
-         Alr.Build_Env.Print (Alr.Root.Current);
-         Alire.Log_Level := Old_Level;
-      end;
+      Alr.Build_Env.Print (Alr.Root.Current);
```
Fabien-Chouteau, line 17: Why are you removing the temporary change of output level?

mosteo: After #385 the problem with the logging no longer happens. The detection status either goes away on regular runs or doesn't happen on redirections.

Fabien-Chouteau: ok :+1:

Fabien-Chouteau (approved):

PR #396, opened 2020-05-11
Types to identify changes in dependencies/solutions

mosteo:
Depends on #355.

This PR is a preparatory one for an upcoming PR that improves the information shown when performing commands that change the crates needed in a workspace. For example:
```
$ alr init --bin xxx && cd xxx
$ alr with ada_voxel_space_demo 
Requested changes:                               

   ✓ ada_voxel_space_demo * (added)

   ✓ ada_voxel_space_demo 1.0.1  (new)
   ✓ libsdl2              2.0.8  (new)
   ✓ libsdl2_image        2.0.3  (new)
   ✓ libsdl2_ttf          2.0.14 (new)
   ✓ sdlada               2.3.1  (new)

Do you want to proceed?
```

This PR implements packages `Alire.Dependencies.Diffs` and `Alire.Solutions.Diffs` which are used to simplify and reuse the identification of changes between the current status and the one post `update/with`.

### DIFF
--- a/src/alire/alire-solutions-diffs.adb
+++ b/src/alire/alire-solutions-diffs.adb
@@ -1,0 +1,221 @@
+with Alire.Containers;
+with Alire.Utils.Tables;
+with Alire.Utils.User_Input;
+
+package body Alire.Solutions.Diffs is
+
+   use type Semantic_Versioning.Version;
+
+   ------------------
+   -- Best_Version --
+   ------------------
+
+   function Best_Version (Status : Crate_Status) return String is
+     (case Status.Status is
+         when Needed   => Semantic_Versioning.Image (Status.Version),
+         when Hinted   => Status.Versions.Image,
+         when Unneeded => "unneeded",
+         when Unsolved => "unsolved");
+
+   -------------
+   -- Between --
+   -------------
+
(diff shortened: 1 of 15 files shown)


src/alire/alire-solutions-diffs.adb
```diff
@@ -0,0 +1,221 @@
...
+         Trace.Info ("Changes to dependency solution:");
+         Trace.Info ("");
+         This.Print (Changed_Only => Changed_Only);
+         Trace.Info ("");
+      else
+         Trace.Info
+           ("There are no changes between the former and new solution.");
+      end if;
+
+      return Utils.User_Input.Query
```
Fabien-Chouteau: I don't think the query belongs here. We might want to print the diff without asking this specific question.

Fabien-Chouteau (commented):
Very nice feature, I like the output a lot.

src/alire/alire-solutions-diffs.adb
mosteo: Right, there is a `Print` that does only that for these cases. This is to avoid repetition of the same prompt in `alr with`, `pin` and `update`. Still this one can be moved elsewhere to keep this class free from specifics.

PR #399, opened 2020-05-11
Generate lockfile on the fly when missing

mosteo:
This PR is intended to reduce manual intervention for users with workspaces initialized pre PR #355. With PR #355 lockfiles are created but never read. However, the queued PRs that build on top of it will cause errors when looking for a lockfile that should be there but isn't.

With this PR the lockfile will be generated transparently when first needed.

### DIFF
--- a/src/alire/alire-lockfiles.adb
+++ b/src/alire/alire-lockfiles.adb
@@ -1,3 +1,4 @@
+with Ada.Directories;
 with Ada.Text_IO;
 
 with Alire.Directories;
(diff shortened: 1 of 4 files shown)


Fabien-Chouteau:
I thought lockfile would only be needed after a `alr pin`.

Fabien-Chouteau (approved):

mosteo:
> I thought lockfile would only be needed after a alr pin.

Strictly speaking we were able to live without it but with the problems outlined in #251; namely that every time we needed the solution to dependencies we recomputed it on the fly. This means that index updates might change the solution in use at any moment. Plus the problems with pinning.

PR #402, opened 2020-05-12
Colorize output

mosteo:
I'm aware of GNATCOLL.Terminal and NCursesAda, but I decided to go for ANSI codes for simplicity. GNATCOLL.Terminal is stateful, which doesn't blend well with our current logging setup, and curses is a heavier dependency with the same problem. Any of these would have required radical changes, whereas this is a minor cumulative change.

This means that colors don't work in Windows (although recent Windows terminal versions can be configured to recognize ANSI).

Colors are palette-based so custom palettes should work properly. Still, I've tried to use colors readable over dark/light backgrounds.

A new global switch `--no-color` allows to force disable colors.

Commands where the effects can be seen, for example: `alr search aw --external-detect`, `alr get ada_voxel_space_demo`

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -23,3 +23,6 @@
 [submodule "testsuite/fixtures/crates/libhello_git"]
 	path = testsuite/fixtures/crates/libhello_git
 	url = https://github.com/alire-project/libhello.git
+[submodule "deps/ansi"]
+	path = deps/ansi
+	url = https://github.com/mosteo/ansi-ada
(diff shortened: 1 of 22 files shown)


Fabien-Chouteau (approved):

Fabien-Chouteau:
@mosteo This doesn't work on Windows cmd and Powershell. I don't know if we have a way to detect the shell and disable the feature.

mosteo:
I just checked on a Windows VM and color is disabled. Are you seeing otherwise?

PR #406, opened 2020-05-13
Fix `pin` command to preserve dependencies across pin/unpin

### LINKED ISSUE #149: alr pin --remove option to allow unpinning

mosteo:
After this PR, `pin` works as intended, allowing the pinning and unpinning of regular releases without affecting dependencies. Previously, the `.toml` file was overwritten, losing the original dependencies in favor of the exact versions at the time of pinning.

New functionality:

* Running `alr pin` without arguments will list existing pins.
* Selective pinning when specifying a crate as argument.
* New `--all` switch to pin/unpin the complete solution (original behavior)
* New `--unpin` switch used for unpinning.

Internal notable changes that support this new implementation and the future dir/repo pins are:

* The `Solution` type is now private and has subprograms to help manage pins in the solution.
* The solver now accepts a previous solution to reuse its pin information to solve pinned dependencies with the pinned version and conserve the pin in the new solution.

Fixes #149. I used `--unpin` instead of `--remove` since it's a less generic verb.


src/alire/alire-releases.ads
```diff
@@ -97,6 +97,9 @@ package Alire.Releases with Preelaborate is
    function Replacing (Base   : Release;
                        Origin : Origins.Origin) return Release;
 
+   function Replacing (Base   : Release;
```
Fabien-Chouteau: I don't understand the name of this subprogram. My understanding of the body is that it marks a release pinned.

Fabien-Chouteau (commented):
The changes look good.

I think we should have at least one functional testing for pinning. As far as I can see the tests so far are only looking at the output of `alr pin`.

Something like:
```
alr with libhello
alr pin libhello=oldversion
alr update
```
Check that libhello was not updated.

mosteo:
Agreed about the test. Also that downgrading a pin correctly downloads the new dependency.

src/alire/alire-releases.ads
mosteo: Yes, it was for keeping it in tune with the other Replacing functions that modify properties. I'll think of something more explicit.

Fabien-Chouteau: Never mind then if it's a pattern.

mosteo:
Added three new tests and a bugfix uncovered by them

PR #409, opened 2020-05-14
minor fixes and adding links

yannickmoy:


doc/catalog-format-spec.md
```diff
@@ -58,8 +58,8 @@ libbar = "^2.0 & /=2.1.3" # Excluding a known bad version
 In some contexts, information can be dynamic: special encodings can be used to
 make data vary depending on the environment (OS, architecture, ...). The
 environment is represented as a set of specific variables which can have a
-specific set of values: see the Parameters section below for a comprehensive
-list.
+specific set of values: see the [Parameters](#-parameters) section below for a
+comprehensive list.
```
mosteo: This link does not work for me, I think because of the '-' after the '#' which doesn't seem to be removed in the rendered file.

doc/publishing.md
```diff
@@ -95,8 +95,8 @@ be edited instead of created).
        through the GitHub web interface (or the
        [`hub`](https://github.com/github/hub) tool).
         1. The base branch you select for the pull request will determine where
-           your changes will become available; see the section on index
-           branches for details.
+           your changes will become available; see the section on [index
+           branches](#-index-branches) for details.
```
mosteo: Likewise, as the other link.

mosteo (changes requested):

mosteo:
Hi Yannick, thanks for the review. Nice seeing you around here :)

doc/catalog-format-spec.md
yannickmoy: I see it online too. I tried to follow advice found online. Will try without the space as I cannot test locally (unless you have a way!)

PR #413, opened 2020-05-15
Testing the spellchecker action

Fabien-Chouteau:


Fabien-Chouteau:
Well... that's not working :)

Fabien-Chouteau:
Actually it does work, as seen in #446.

PR #415, opened 2020-05-18
`alr update` improvements

### LINKED ISSUE #237: Recursive upgrading project dependencies

### LINKED ISSUE #251: [RFC] Do not update dependency versions unless requested

mosteo:
* Add back the `--online` option to update index and solution in a single step
* New ability to say in the command line the crates to try to update: `alr update [crate1]...` When none given, all crates in current solution are valid update candidates.
   * Refactored update logic into `Alire` while implementing this change.

Up to now, updating a workspace dependencies was an all-or-nothing affair.

Fixes #237, fixes #251


src/alr/alr-commands-update.adb
```diff
@@ -1,42 +1,49 @@
...
+                  Root.Current.Solution;
    begin
-      Requires_Full_Index;
 
-      Requires_Valid_Session;
+      --  Ensure requested crates are in solution first
+
+      for Crate of Allowed loop
+         if not Old.Releases.Contains (Crate) then
+            Reportaise_Wrong_Arguments ("Requested crate is not in solution: "
```
Fabien-Chouteau: I feel like it would be more meaningful for the user to use `Requested crate is not a dependency` here. `Solution` is a bit internal vocabulary.

src/alr/alr-commands-update.adb
```diff
@@ -113,9 +147,23 @@ package body Alr.Commands.Update is
    function Long_Description (Cmd : Command)
                               return Alire.Utils.String_Vector is
      (Alire.Utils.Empty_Vector
-      .Append ("Resolves dependencies using the loaded indexes, and"
-               & " regenerates the aggregate project building file found in"
-               & " <crate>" & GNAT.OS_Lib.Directory_Separator
-               & Alire.Paths.Working_Folder_Inside_Root));
+      .Append ("Resolves unpinned dependencies using available indexes"));
```
Fabien-Chouteau: Is the "selective" per crate update explain in the help message?

src/alr/alr-commands-update.adb
mosteo: D'accord

src/alr/alr-commands-update.adb
mosteo: It is now

mosteo:
Added a check that requested crates be not pinned, since no update would be found for them anyway.

Fabien-Chouteau (approved):

PR #420, opened 2020-05-18
Fix missing checks for invalid solution

mosteo:
These omissions were causing uncontained errors after an `alr get --only`


mosteo:
This one has the same problem with nomenclature as the one yesterday. I'm not sure how to refer to the situation in which there is no solution for the dependencies without using "solution". Suggestions welcome!

Fabien-Chouteau (approved):

Fabien-Chouteau:
Maybe I am wrong about the word solution. In this PR is makes more sens to me than the other I commented on (#415).

Issue #367, opened 2020-04-30
Persistent configuration


mosteo:
There is an increasing need to keep some information stored on disk:

- Msys2 choice
- Withing choices
- User identity for publishing
- ...

TODO list:
- [x] Global config
- [x] Local (per workspace) config
- [ ] Overriding via switch; defaults via env vars?

PR #424, opened 2020-05-26
Add support for persistent configuration options

Fabien-Chouteau:
The configuration options are either global or local (local overriding
the global definitions) and are stored in TOML format.

This patch also introduces a command to handle configuration options
from the command line.

Implements #367

### DIFF
--- a/src/alire/alire-config.adb
+++ b/src/alire/alire-config.adb
@@ -1,10 +1,86 @@
+with Ada.Containers.Hashed_Maps;
+with Ada.Strings.Unbounded;
+with Ada.Strings.Unbounded.Hash;
+
+with GNAT.Regexp;
+
 with Alire.Environment;
 with Alire.Platform;
+with Alire.Directories;
+
+with TOML.File_IO;
 
 package body Alire.Config is
 
+   use Ada.Strings.Unbounded;
+   use TOML;
+
+   type Config_Value is record
+      Source : Unbounded_String;
+      Value  : TOML.TOML_Value;
+      Lvl    : Level;
+   end record;
+
+   No_Config_Value : constant Config_Value := (Source => Null_Unbounded_String,
+                                               Value  => No_TOML_Value,
+                                               Lvl    => Global);
+
+   package Config_Maps is new Ada.Containers.Hashed_Maps
+     (Key_Type        => Unbounded_String,
+      Element_Type    => Config_Value,
+      Hash            => Hash,
+      Equivalent_Keys => "=");
+
+   Config_Map : Config_Maps.Map;
+
+   procedure Import (Table  : TOML.TOML_Value;
+                     Lvl    : Level;
+                     Source : String;
+                     Prefix : String := "");
+   --  Import TOML Table in the Config_Map global variable
+
+   function Get (Key : Config_Key) return Config_Value;
+
+   function Image (Val : TOML_Value) return String;
+
    type String_Access is access String;
    Config_Path : String_Access;
+
+   -----------
+   -- Image --
+   -----------
+
+   function Image (F : TOML.Any_Float) return String is
+   begin
+      case F.Kind is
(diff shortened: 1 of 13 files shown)


testsuite/tests/config/basics/test.py
```diff
@@ -0,0 +1,105 @@
+"""
+Check that 'alr -h <command>' produces same output as 'alr help command'
+"""
```
mosteo: Forgotten description?

src/alire/alire-config.adb
```diff
@@ -33,4 +103,297 @@ package body Alire.Config is
...
+      if Val.Value.Is_Null then
+         Trace.Detail ("Using default value for configuration '" & Key &
+                         "': '" & Image (Default) & "'");
+         return Default;
+
+      elsif Val.Value.Kind /= Expected_TOML_Kind then
+         Trace.Error ("Invalid type ('" & Val.Value.Kind'Img &
+                        "') for configuration '" & Key & "'");
+         Trace.Error ("in '" & To_String (Val.Source) & "'");
+         Trace.Error ("Boolean expected");
```
mosteo: I don't think this is intended to always be `Boolean`?

src/alire/alire-config.adb
```diff
@@ -33,4 +103,297 @@ package body Alire.Config is
...
+
+   --------------
+   -- Filepath --
+   --------------
+
+   function Filepath (Lvl : Level) return Absolute_Path is
+   begin
+      case Lvl is
+         when Global =>
+            return Alire.Platform.Default_Config_Folder / "config.toml";
```
mosteo: I suspect this should be `Config.Path`, or the ability to use different global configurations (`-c` switch) will be broken

Fabien-Chouteau:
@mosteo the patch is ready for review.

I don't know what those two macOS failures are...

mosteo:
> I don't know what those two macOS failures are...

It seems to be a problem with setting up `e3.testsuite`. Do you know how the migration to python 3 is going?

mosteo (approved):

Fabien-Chouteau:
There migration to Python3 is complete I think. But anyway we use the release in `pip` aren't we?

mosteo:
> But anyway we use the release in pip aren't we?

True, but we are forcing the use of python2/pip2. Not sure if that may have anything to do.

> The migration to Python3 is complete I think

OK, so we should migrate to it too. But first I'll try to find whatever is happening here (unless you're on it).

pmderodat:
Hello,

Both the latest version of e3-testsuite published on pip and the `master` branch are now Python3 only. We will publish a release based on the current `master` branch at some point today, and it contains minor breaking changes. I can help for the migration, if you want. :-) By the way, there is now a proper documentation for this framework: https://e3-testsuite.readthedocs.io/

mosteo:
> I can help for the migration, if you want

Than would be great, thanks :-) 

I've tried to create a python3 venv and install the e3-testsuite in it. The installation succeeds (after previously pipinstalling `wheel`, or I get a bunch of errors) for either the latest release in pip or using master directly.

I'm stuck at trying to run our `testsuite/run.py`, which errs with a `ERROR    cannot find driver for test 'tests__config__basics'` for every test. Since we have `testsuite/run.py` and `testsuite/drivers/python_script.py` that are not updated despite using the latest `e3-testsuite`, I guess the problem may lie there. `testsuite/run.py` mentions python2, for example.

I was going to try next to go through the tutorial in the docs, but I'll way for your feedback about the proper course of action.

pmderodat:
I have a commit ready to transition the testsuite to Python3 and the new e3.testsuite API. I’ll open a PR as soon as we have managed to publish a new e3.testsuite release on Pypi. :-)

pmderodat:
We just release e3-testsuite on Pypi, so I could create the PR #432

PR #428, opened 2020-06-01
Exit code fix for `get --build` with build failure

mosteo:
We were not erroring if the build failed, which is counterintuitive. We now require successful retrieval and build to exit with code 0. A test to verify this situation has been added.


Jellix:
https://github.com/alire-project/alire/blob/440a4c0b66e2e90c8eb1fa9a402eb21558d2a8cc/testsuite/tests/get/build/test.py#L8

After running alr to build the "good" build, shouldn't there be a similar assertion like for the "bad" build that the returned code was OK?

Fabien-Chouteau:
> After running alr to build the "good" build, shouldn't there be a similar assertion like for the "bad" build that the returned code was OK?

@Jellix the python function `run_alr` will report the error if the argument `complain_on_error` is not set to false.

Fabien-Chouteau (approved):

Issue #377, opened 2020-05-01
alr update fails: Dependency resolution failed


stcarrez:
Yet another useful improvement for Alire. Sometimes you get the following error when you run the `alr update` command:

```
-> Dependency resolution failed
ERROR: Update failed
ERROR: alr update unsuccessful
```

The `-v` option does not help. The `-vv` option requires deep understanding of how Alire works.

It would be nice to have an explanation about the dependency failure.

Of course, I made a mistake, but which crate is having a wrong dependency and for which other crate? That's is the question!

mosteo:
Yes, I think this is a duplicate of #220.

Fabien-Chouteau:
There are indeed many places where `alr` output can be more user-friendly.
Thank you for reporting those :+1: 

@mosteo maybe we should have a specific GitHub issue label for this kind of problem so that we can list them when we have time to work on usability.

mosteo:
:+1: Created the `usability` label for this

mosteo:
Implemented by #451. Now you get a summary of the dependency solution, where any problematic dependency will also be highlighted as missing.

PR #429, opened 2020-06-02
Internal enhancements to Solution and Dependency states

mosteo:
This is a largish PR so I've tried to have a logical sequence of commits, although a couple of them in the middle don't compile (the idea is to squash it after review).

While prototyping the new "softlinks" to provide dependencies I ran into some limitations of our current solution format. Small workarounds here and there started piling up and it became increasingly clear that a better foundation was advantageous for cleanliness and flexibility.

This PR expands the Solution information to always hold the state of all dependencies in the solution, instead of the partial information about solved releases/hints we had previously. To do so, this PR moves from valid/invalid to a complete/incomplete paradigm. A solution is complete when all dependencies have a fulfilling release, but is otherwise treated the same be it complete or not (no more `Is_Valid` checks), because dependency information is always available (via the new `Alire.Dependencies.States` type.

This has several benefits:

* Some expensive solver checks can be removed by using solutions directly to track search state. 
* The solver is now exhaustive about incomplete solutions. This is useful when no solution exists to report the "less incomplete" solution found (related: #377). From the users' POV, we no longer will need to leave them in the dark when dependency resolution doesn't find a complete solution (it no longer can fail; at worst, the solver reports missing dependencies).
* As a corollary, user changes that result in an incomplete solution won't imply losing all other solvable dependencies. This may simplify workflows in platforms without a native package manager, or when working with still unindexed crates.
* A few changes down the road are simpler: the use of directories to fulfill dependencies, more precise reporting in solution changes, clearer update logic...

This PR has no user-facing changes. For now, the same valid/invalid logic is emulated to keep this PR more focused. The only visible difference is the detail/debug output from the solver, which now reports more details about complete/incomplete solutions found. Summary of main changes:

* New `Alire.Dependencies.States` to store dependencies and how they're fulfilled.
* Reworking of `Alire.Solutions` to rely on these new states.
* Reworking of `Alire.Solver` to incrementally build solutions while keeping track of dependency states.
* Pinning info is moved from `Alire.Releases.Release` to `Alire.Dependencies.States`.

The user-facing improvements are being prepared in follow-up PRs.


src/alire/alire-solutions.adb
```diff
@@ -13,49 +15,267 @@ package body Alire.Solutions is
...
+      end return;
+   end Crates;
+
+   -----------------
+   -- Debug_Print --
+   -----------------
+
+   procedure Debug_Print (This : Solution) is
+   begin
+      Trace.Always ("SSSSSSSSSSSSSSS complete: " & This.Is_Complete'Img);
```
Fabien-Chouteau: I guess that should be removed :)

src/alire/alire-solutions.adb
```diff
@@ -13,49 +15,267 @@ package body Alire.Solutions is
...
+      for Rel of This.Releases loop
+         Trace.Always ("   " & Rel.Milestone.Image);
+      end loop;
+
+      Trace.Always ("Dependencies:" & This.Dependencies.Length'Img);
+      for Dep of This.Dependencies loop
+         Trace.Always ("   " & Dep.Image);
+      end loop;
+
+      Trace.Always ("sssssssssssssss");
```
Fabien-Chouteau: And this too.

src/alire/alire-milestones.ads
```diff
@@ -56,7 +72,7 @@ private
    function New_Milestone (Name    : Crate_Name;
                            Version : Semantic_Versioning.Version)
                            return Milestone
-     is (Name'Length, Name, Version);
+     is (Name.Name'Length, Name, Version);
```
Fabien-Chouteau: Should this be `Name.Length`?

Fabien-Chouteau (approved):

PR #432, opened 2020-06-04
Transition the testsuite for Python3 and the new e3.testsuite API

pmderodat:


pmderodat:
@mosteo @Fabien-Chouteau CI scripts for GitHub still install a Python2 interpreter. I’m not yet familiar with it, so would one of you be able to adjust them?

mosteo:
Thanks a lot for the migration, @pmderodat. I'll take a stab at it.

pmderodat:
Great, thank you! Please let me know if you have any issue with the new testsuite framework.

Fabien-Chouteau:
Thank you @pmderodat ,

For Windows and macOS you have to change the python version in the GitHub workflow:
```
    - name: Install Python 2.x (required for the testsuite)
      uses: actions/setup-python@v1
      with:
        python-version: '2.x'
```
to
```
    - name: Install Python 3.x (required for the testsuite)
      uses: actions/setup-python@v1
      with:
        python-version: '3.x'
```

For Linux I am not sure, this is probably part of the Docker that we use. @mosteo ?

And then the definitions of `run_python` and `run_pip` at https://github.com/pmderodat/alire/blob/8cebc34034a92f8eac884914cfbba65bca2ada5c/scripts/ci-github.sh#L54

testsuite/run.py
```diff
@@ -13,32 +13,31 @@
...
+    tests_subdir = 'tests'
+    test_driver_map = {'python-script': PythonScriptDriver}
 
-    @property
-    def default_driver(self):
-        return 'decoder'
+    def set_up(self):
+        super().set_up()
+
+        # Some tests rely on an initially empty GPR_PROJECT_PATH variable
```
Fabien-Chouteau, line 29: Is this something new? Which test relies on that?

Fabien-Chouteau (approved):

testsuite/run.py
mosteo: The `alr setenv` ones will have unexpected output when run with `GPR_PROJECT_PATH`, I had that once.

pmderodat: Probably new, as I found that only today. That was for the `setenv/with-external` testcase

mosteo:
> For Linux I am not sure, this is probably part of the Docker that we use. @mosteo ?

Right, it's possible some of them already include it by default, but we'll soon see.

mosteo:
Hi @Fabien-Chouteau and @pmderodat.

Only Windows remains to pass tests (surprise). I'm hitting a permissions error that seems to be a new thing. Every test on Windows fails with:
```
"rlimit: cannot spawn process (error 0x5)\r\n"
```
According to [this page](https://docs.microsoft.com/en-us/windows/win32/debug/system-error-codes--0-499-) (not 100% sure it applies) the error would be:
```ERROR_ACCESS_DENIED
5 (0x5)
Access is denied.
```
Does it ring any bells? I'm not finding much that seems related out there. I've read it could be several things, from permissions/incorrect paths to files already opened/read-only.

Here's the complete output of a test:
```
!e3.testsuite.result.TestResult
diff: null
env:
  driver: python-script
  indexes:
    checked_index: {}
  test_dir: D:\a\alire\alire\testsuite\tests\get\integrity-tarballs
  test_name: get__integrity-tarballs
  working_dir: C:\Users\RUNNER~1\AppData\Local\Temp\tmpxxycwx7o\get__integrity-tarballs
expected: null
failure_reasons: !!set {}
info: {}
log: "Running: 'C:\\hostedtoolcache\\windows\\Python\\3.8.3\\x64\\python.exe' test.py\
  \ (cwd=C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\tmpxxycwx7o\\get__integrity-tarballs)\n\
  Status code: 1\nOutput: \nrlimit: cannot spawn process (error 0x5)\r\n\n"
msg: non-zero status code
out: null
processes:
- cmd:
  - C:\hostedtoolcache\windows\Python\3.8.3\x64\python.exe
  - test.py
  cwd: C:\Users\RUNNER~1\AppData\Local\Temp\tmpxxycwx7o\get__integrity-tarballs
  output: "rlimit: cannot spawn process (error 0x5)\r\n"
  status: 1
status: !e3.testsuite.result.TestStatus |-
  2
test_name: get__integrity-tarballs
time: null
```

pmderodat:
Hello,

I think this is due to a problem in the e3-core Python package (the one that provides this rlimit program). I’ll get in touch with maintainers, but in the meantime you should be able to workaround this disabling the timeout feature (so that rlimit is no longer involved): add a  `default_process_timeout = 0` attribute in the PythonTestDriver subclass (`testsuite/drivers/python_script.py`)

mosteo:
Thanks, @pmderodat, your suggestion worked. I'm going to merge with the workaround until a proper solution is available.

PR #437, opened 2020-06-09
Refactor Alr.Checkout into Alire.Workspace

mosteo:
Move reusable logic for checkouts/updates from Alr into Alire. There are no user-facing changes. This refactor, besides the general goal of having Alire as a library, simplifies some upcoming PRs.

Goes on top of #431


testsuite/tests/get/external-tool-dependency/test.py
```diff
@@ -2,17 +2,31 @@
...
 # Should silently retrieve everything
 p = run_alr('get', 'main')
 assert_eq('', p.out)
 
 # Check folder contents
 compare(contents('main_1.0.0_filesystem/'),
         ['main_1.0.0_filesystem/alire',
+         'main_1.0.0_filesystem/alire/cache',
+         'main_1.0.0_filesystem/alire/cache/dependencies',
+         'main_1.0.0_filesystem/alire/cache/dependencies/make_' + version + '_external',
```
Fabien-Chouteau, line 29: Why is there a folder for make now?

Fabien-Chouteau (approved):

testsuite/tests/get/external-tool-dependency/test.py
mosteo: Sorry! I missed this comment at the time. While doing the refactor it seemed a good idea to have some trail of all dependencies in use in the cache directory. The folder is empty for non-source dependencies. It could be removed without further implications.

mosteo: Actually, reviewing the code, there's an additional reason which is to have a folder from which to run any associated actions. But this is at this time hypothetical because externals can't define actions.

PR #439, opened 2020-06-09
Ability to use a local folder to fulfill a dependency

mosteo:
This PR implements the pinning of dependencies to folders. These folders then act as a wildcard for any version; that is, the dependency will be always satisfied. I chose this approach instead of associating a version to the folder because both opam and cargo work like this, and for crates in progress the version is actually not a concern.

On terminology and internals: for Alire users, both pinning to a version or a folder is presented under the concept of pins (because the idea is to fix a dependency to something concrete). Internally, there are separate concepts, "pin to a version" (already implemented) and "linked to a folder" (new in this PR), and the term softlink/link is used for new types/subprograms that involve these new user pins.

Given the lack of version of these linked crates, internally we don't rely on placeholder releases. Instead, links are represented with a new `Alire.Externals.Softlinks` type and the solver is aware of them, which has the added benefit of the solver not needing to blindly try releases until finding the good one; instead, links are reused from the solution the solver receives as a basis.

There are two ways of pinning a dependency to a folder: right away at the time of adding the dependency (`alr with crate --url path`) or later with the pin command (`alr pin crate --url path`). I named the switch `--url` instead of `--dir` because in the foreseeable future we may want to support a remote repository or archive instead (as `opam` does).

When adding and linking simultaneously, the dependency needs not to exist in the index, which is convenient to rely on unindexed dependencies:
```alr with unindexed_crate --url /path/to/working/version```

See the new tests for examples of usage. In essence:
```
$ alr init --bin xxx
$ cd xxx
$ alr with libhello --url ../libhello
Requested changes:                                            

   ✓ libhello * (added)

   ⊙ libhello file:/tmp/a/libhello (pinned)

Do you want to proceed?
Using default: Yes
$ alr with --solve
Dependencies (direct):
   libhello*
Dependencies (external):
   libhello* (direct,linked,target=/tmp/a/libhello)
Dependencies (graph):
   xxx=0.0.0 --> libhello*
┌───────────┐     ┌───────────┐
│ xxx=0.0.0 │ ──> │ libhello* │
└───────────┘     └───────────┘
$ alr setenv
export GPR_PROJECT_PATH="/tmp/a/libhello"
export ALIRE="True"
```

Goes on top of #437, fixes parts of #240

### CODE REVIEW

File: src/alire/alire-pinning.ads [hunk 1]
```diff
@@ -19,13 +19,22 @@ package Alire.Pinning is
    --  must exist in the solution. Root dependencies are given, and a previous
    --  solution with possibly more pins. The resulting solution may be invalid.
 
+   function Pin_To (URL      : String;
+                    Solution : Solutions.Solution;
+                    Crate    : Crate_Name)
+                    return Solutions.Solution with
+     Pre => Solution.Depends_On (Crate) or else
+     raise Checked_Error with
+       "Cannot pin crate not in dependencies: " & (+Crate);
```
Fabien-Chouteau, line 28: Is this pre-condition checked before the call? Because if we compile without checks this error message will never be displayed to the user.

src/alr/alr-commands-pin.adb
```diff
@@ -214,14 +242,20 @@ package body Alr.Commands.Pin is
...
+               & " By default, the current solution version is used."
                & " A pinned release is not affected by automatic updates.")
       .New_Line
       .Append ("Without arguments, show existing pins.")
       .New_Line
       .Append ("Use --all to pin the whole current solution.")
       .New_Line
       .Append ("Specify a single crate to modify its pin.")
+      .New_Line
+      .Append ("Use the --url switch to "
```
Fabien-Chouteau: Even if it is just an alias, I would like to have a `--dir` option here. `--url` for a local directory doesn't feel right (even if I know it is correct).

And if the URL part is not implemented yet, maybe keep the `--url` option for later.

Fabien-Chouteau (changes requested):

Fabien-Chouteau:
So when pinning to a directory, only the GPR_PROJECT_PATH and current directory for actions change. How do you see the URL thing working? One time download or re-download for every `alr` command?

Fabien-Chouteau:
Awesome feature by the way :)

src/alire/alire-pinning.ads
mosteo: It is checked [here](https://github.com/mosteo/alire/blob/81deb84f2f0e9452225164903779901077977348/src/alr/alr-commands-pin.adb#L100). However your question puts the spot on something I've wondered a few times: should we treat `-gnata` as a requisite (hence you can rely on these checks -- incidentally we are doing this for `Alire.Crate_Name`) or not (so they're only useful for testing -- and you want the Pre/Post never to trigger).

I don't really like the second option unless one is using SPARK, or it's a explicit codebase policy. I have never seen this discussed from this perspective either, excepts some remarks from Kazakov at `c.l.a` that I couldn't really follow to conclusion. If checks are expected to be disabled at some point, certainly it makes no sense to use "expected" exceptions like `Checked_Error` in our case, since that changes the program from valid to erroneous.

Do you have some policy about this at AdaCore? (I.e., checks are always disabled for release builds).

src/alr/alr-commands-pin.adb
mosteo: OK, this was something that made me uneasy too.

mosteo:
> only the GPR_PROJECT_PATH and current directory for actions change

I'm not sure what you mean by actions here... A linked dir won't have actions.

> How do you see the URL thing working? One time download or re-download for every alr command?

My current idea is that a remote source archive would be downloaded once, when added. A repository could be pulled for changes via `alr update`.

src/alr/alr-commands-pin.adb
mosteo: Given that this will ripple through all tests and follow-up PRs... I was thinking to rename it to `--use` instead to avoid latter having two similar/aliased switches. What do you think?

src/alire/alire-pinning.ads
Fabien-Chouteau: I don't think we have a common policy for releases. It's all a question of performances.

For the first releases I think we should enable all checks, so that we catch as many problems as possible. But then we might want to remove them for performance.

src/alr/alr-commands-pin.adb
Fabien-Chouteau: `alr pin --use=~/src/plop`?

Yeah I think that's ok.

Fabien-Chouteau:
> I'm not sure what you mean by actions here... A linked dir won't have actions.

The actions that can be defined in the crate toml: 
```toml
[[general.actions.'case(os)'.linux]]
type = "post-fetch"
command = ["make"]
```

> My current idea is that a remote source archive would be downloaded once, when added. A repository could be pulled for changes via `alr update`.

Sounds good :+1:

src/alire/alire-pinning.ads
mosteo: Ok, so better not rely on Pre/Post checks as part of the "actual" code for the future.

mosteo:
> The actions that can be defined in the crate toml:

A linked folder does not go through `fetch` or `compile` steps, so no actions will be run for it. For the root release in the workspace, the `compile` actions will run normally from its own folder.

Fabien-Chouteau (approved):

PR #443, opened 2020-06-10
Alr.Platforms.Windows: use Alire.config in msys2 install

Fabien-Chouteau:


src/alr/os_windows/alr-platforms-windows.adb
```diff
@@ -149,29 +149,54 @@ package body Alr.Platforms.Windows is
             return Alire.Outcome_Failure ("Cannot setup msys2 environment");
       end;
 
+      if not Cfg.Defined ("msys2.install_dir") then
+         --  Save msys2 install dir in the global config
+         Cfg.Edit.Set (Path  => Cfg.Filepath (Cfg.Global),
+                       Key   => "msys2.install_dir",
+                       Value => Alire.Utils.YAML.YAML_Stringify (Install_Dir));
```
mosteo: I think that the `TOML` library escapes strings as needed. Just curious, has this any advantage?

mosteo (approved):

src/alr/os_windows/alr-platforms-windows.adb
Fabien-Chouteau: I actually did this because of an error I got. I will check what the TOML lib can do.

Fabien-Chouteau: Should be fixed. Thanks for pointing it out, I was not happy about this YAML call...

PR #447, opened 2020-06-11
Allow incomplete solutions in relevant commands

### LINKED ISSUE #377: alr update fails: Dependency resolution failed

mosteo:
Goes on top of #440, fixes #377.

This PR completes the work on pinning and solutions/solver internal rework by removing the valid/invalid solution dichotomy for the user (it was internally removed since #429). With these changes, solutions can be instead complete (when all dependencies have a corresponding release or linked folder) or incomplete. 

In practical terms, instead of stopping with "invalid solution" errors whenever the solution is incomplete, now `alr` still presents the solution to the user, identifying any missing dependencies. The user can still accept this (although the default is No) or use `--force` to default to Yes (so it can also be done non-interactively).

This PR applies these changes to `get`, `pin`, `update` and `with`. The user should gain flexibility when working with "fringe" crates that use missing dependencies/versions, or on unsupported distributions, and encounter less breaking situations when a solution becomes incomplete for any reason.

A side benefit for the publishing workflow is that, for unindexed dependencies, they can be included in the crate to be published without the need to also index the missing dependencies as externals. This way, no crate should be published with hidden dependencies (there's no cost to include them), nor we need to clutter the community index with "empty" externals.

Externals of course still are needed to allow detection of tools/system packages, and `hint` externals still are valuable when they include manual installation hints, but `hint` externals are no longer mandatory to be able to retrieve dependent crates as they were until now.

Because of this new flexibility, `Solution.Is_Complete` is now  `False` for crates depending on undetected external hints, whereas `Solution.Valid` (deprecated by this PR) was True for these situations. This was necessary because otherwise no crate with undetected externals was retrievable, or could be worked with. Currently these can be retrieved/used after the initial user confirmation. 

The bottom line is that now all `Is_Complete` solutions should always properly build, whereas in the past that was not true for `Valid` solutions unless external hints where provided by the user. (Linked dirs introduced in #439 still need to be correct for a complete solution to successfully build, of course).  In other words, a complete solution after this PR has more guarantees than a valid solution had in the past.

New tests check and illustrate these changes.

In my consideration the initial problems raised in #240 are solved with #439 and this PR. There are remaining issues pointed out there, that have ties to the publishing workflow. I will address these in a new issue, since I intend to move into the publishing workflow after this and I've been thinking about it for some time, but I'd like to discuss before starting.


doc/user-changes.md
```diff
@@ -0,0 +1,37 @@
+# User-facing changes log
```
Fabien-Chouteau: We will have to include this Markdown in the website doc page.

Fabien-Chouteau:
@mosteo It is difficult to review the PR with code from other branches. Maybe you can change the base branch of this PR to make more readable.

mosteo:
> @mosteo It is difficult to review the PR with code from other branches. Maybe you can change the base branch of this PR to make more readable.

Yes, this is a problem for PRs coming from another repository since I cannot use base branches that don't exist in the target repo... I'll see what I can do.

mosteo:
I've pushed to this repo the base branch, so you should be seeing only this PR commits now.

doc/user-changes.md
mosteo: I'll take a stab at it if you don't mind, so I can familiarize a bit more with the webpage generation.

mosteo: I'm reading through the sources for `alire.ada.dev` and if you can confirm that:

- [This line](https://github.com/alire-project/alire.ada.dev/blob/1be8728139d90f5c6d08481da7f7e762e0209585/update-gh-pages.sh#L49) is already including the file in the website generation.
- Because of [this configuration](https://github.com/alire-project/alire.ada.dev/blob/1be8728139d90f5c6d08481da7f7e762e0209585/_config.yml#L66), it may be better to make each entry in the log a third-level header so they don't clutter the TOC.

So no changes needed for the website itself?

Fabien-Chouteau:
> Yes, this is a problem for PRs coming from another repository since I cannot use base branches that don't exist in the target repo... I'll see what I can do.

Why don't you push `feat/` branches to the main Alire repo directly? That's what I do.

doc/user-changes.md
Fabien-Chouteau: Here: https://github.com/alire-project/alire.ada.dev/blob/1be8728139d90f5c6d08481da7f7e762e0209585/docs/index.md

Each do file is included explicitly, which also controls how in what order they are in the table of content.

Fabien-Chouteau (approved):

mosteo:
> Why don't you push `feat/` branches to the main Alire repo directly? That's what I do.

I thought that this way I lowered the noise for you with all the rebasing going on, but if you don't mind I might go back to working against the Alire repo directly.

doc/user-changes.md
mosteo: I see, thanks.

Fabien-Chouteau:
I don't think it will create any extra noise.

PR #450, opened 2020-06-16
Detect and use Alire metadata found in pinned directory

mosteo:
Goes on top of #449.

PR #447 introduces the ability to pin a folder that needs not be an Alire crate. This allows the use of Ada code entirely out of the Alire ecosystem to fulfill a dependency. The shortcoming is that any dependencies introduced by such code will remain unknown and have to be manually dealt with. Also, if the target folder was, indeed, an Alire crate, its information is ignored.

This is an incremental feature in which, if the target folder contains Alire metadata, it is loaded and used to 

- Verify the crate we are pinning matches its dependency in name
- Add its dependencies during resolution
- Add its project paths during building
- Optionally, autodetect the crate for a shorthand `alr with --url` form.

In short, the pinned Alire crate is used normally as if it had been retrieved as a regular dependency, except for the version, which is ignored (the dependency will always be fulfilled, as if it were a "raw" pinned folder).

### DIFF
--- a/doc/user-changes.md
+++ b/doc/user-changes.md
@@ -4,6 +4,19 @@ This document is a development diary summarizing changes in `alr` that notably
 affect the user experience. It is intended as a one-stop point for users to
 stay on top of `alr` new features.
 
+### Use crate metadata when pinning to a directory
+
+PR [#450](https://github.com/alire-project/alire/pull/450).
+
+When pinning a dependency to a directory (`alr pin|with <crate> --url`), detect
+if the target contains Alire metadata (as usual, in an `alire` subdir). In that
+case, use it to determine further dependencies and project file scopes. Also,
+the target dependency name is verified.
+
+For such a target directory, a shortcut `with` command is available since the
+crate name can be determined from the metadata: `alr with --url
+/path/to/target` (note the absence of a crate name).
+
 ### Allow working with incomplete solutions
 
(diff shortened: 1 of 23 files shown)


doc/user-changes.md
```diff
@@ -4,6 +4,19 @@ This document is a development diary summarizing changes in `alr` that notably
 affect the user experience. It is intended as a one-stop point for users to
 stay on top of `alr` new features.
 
+## Use crate metadata when pinning to a directory
```
Fabien-Chouteau: Awesome I was about to suggest that when reading the main PR :)

Fabien-Chouteau:
>     * Add its dependencies during resolution

If there are more than one release in the crate, which one is going to be used for that?

Fabien-Chouteau (approved):

mosteo:
> If there are more than one release in the crate, which one is going to be used for that?

This should cause an error, because we expect the local `alire/crate.toml` to contain only the checked out release metadata. When you `alr get` a release that's what's stored. For a never-published crate, `init` also only creates one release.

Actually this ties in a bit with a follow-up discussion I want to have about the maintenance/publishing workflow, because I think we should move to making `alire/crate.toml` manually edited only and put under version control. I'm organizing my ideas and will open an issue soon before doing anything.

PR #452, opened 2020-06-19
Alire.Environment: implement new handling of env variables

Fabien-Chouteau:
Including using the variables defined in the crate and formatting them.
For instance replacing "${DISTIB_ROOT}" by "/".

### DIFF
--- a/src/alire/alire-environment-formatting.adb
+++ b/src/alire/alire-environment-formatting.adb
@@ -1,0 +1,127 @@
+with Ada.Strings.Unbounded; use Ada.Strings.Unbounded;
+
+with Alire.OS_Lib;
+with Alire.Directories;
+with Alire.Platform;
+with Alire.Paths;
+
+package body Alire.Environment.Formatting is
+
+   ----------------
+   -- Find_Start --
+   ----------------
+
+   function Find_Start (Str  : Unbounded_String;
+                        From : Positive)
+                        return Natural
+   is
+      Loc : Natural := From;
+   begin
+
+      loop
+         Loc := Index (Str, "${", Loc);
+
+         exit when Loc = 0;
+
+         --  Check for escape character before the pattern
+         if Loc = 1 or else Element (Str, Loc - 1) /= '\' then
+            return Loc;
+         else
+
+            --  Pattern found but with an escape character
+            Loc := Loc + 2;
+         end if;
+      end loop;
+
+      return Loc;
+   end Find_Start;
+
+   --------------
+   -- Find_End --
+   --------------
+
+   function Find_End (Str  : Unbounded_String;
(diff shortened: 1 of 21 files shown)


src/alire/alire-environment.adb
```diff
@@ -0,0 +1,263 @@
...
+with Alire.Properties.Scenarios;
+with Alire.OS_Lib;
+with Alire.GPR;
+with Alire.Utils;
+with Alire.Environment.Formatting;
+
+package body Alire.Environment is
+
+   function "+" (Str : String) return Unbounded_String
+                 renames To_Unbounded_String;
```
mosteo: There is one `"+"` already visible in `alire.ads`

mosteo (approved):

src/alire/alire-environment-formatting.adb
```diff
@@ -0,0 +1,127 @@
...
+           Alire.Directories.Current;
+      begin
+
+         if Id = "DISTRIB_ROOT" then
+            Replace_Slice (Str, From, To, Platform.Distribution_Root);
+
+         elsif Id = "CRATE_ROOT" then
+            Replace_Slice
+              (Str, From, To,
+               Working_Folder /
```
Fabien-Chouteau, line 83: I guess the way I compute the root directory of the crate is not compatible with softlink pins.

If possible it would be best to have a function to get that info on any release.

Fabien-Chouteau (commented):
@mosteo There are a couple things that wont work with this implementation.

One of them is the export of the environment variable when running the `post-fetch` actions: https://github.com/alire-project/alire/blob/0db9ca8c78370ebf7772023ad72c3fcff3c2b40b/src/alr/alr-commands-get.adb#L159

src/alire/alire-environment-formatting.adb
mosteo: Right. Let me take a stab at this.

mosteo:
> One of them is the export of the environment variable when running the `post-fetch` actions:

You mean that we should set up the environment before running these, right?

Fabien-Chouteau:
Yes, for instance `sdlada` `post-fetch` is compiling stuff that require `C_INCLUDE_PATH` to be set.

mosteo:
Hi Fabien, can I rebase this PR onto `master`? This way #456 will be testable.

Fabien-Chouteau:
> Hi Fabien, can I rebase this PR onto `master`? This way #456 will be testable.

Sure no problem.

mosteo:
For some reason, now on Windows the python interpreter is generating unix-ended strings while alr output is coming in with DOS-ended strings.

I've made a crude fix for this, but I'm not sure this is the way to go.

PR #453, opened 2020-06-22
Show linked paths and per-release dependencies in `--solve` solutions

mosteo:
This PR adds some bits of new info to the output of `alr with --solve`:

In the graph section of the solution printout, the concrete dependency for each release that brings in another release is now shown. This may help in understanding why a release is chosen (or unsolvable), particularly when there are different restrictions on a same crate introduced by different releases.

For the linked folders with Alire metadata introduced in #450, the source folder is shown in the releases section. Otherwise, such source folders are listed in the externals section without a concrete release (as we cannot determine a version for them).

Example output (from the test) with the new dependency set after the solved milestone:
```
Dependencies (graph):
   hello=1.0.1      --> libhello=1.0.1 (^1.0)
   superhello=1.0.0 --> libhello=1.0.1 (~1.0)
```

### DIFF
--- a/doc/user-changes.md
+++ b/doc/user-changes.md
@@ -4,6 +4,20 @@ This document is a development diary summarizing changes in `alr` that notably
 affect the user experience. It is intended as a one-stop point for users to
 stay on top of `alr` new features.
 
+### Show release-specific dependency sets in solutions
+
+PR [#453](https://github.com/alire-project/alire/pull/453).
+
+The dependency solution shown with the `--solve` switch now details for each
+release the particular version set with which a dependency is brought into the
+dependency closure. For example:
+
+```
+Dependencies (graph):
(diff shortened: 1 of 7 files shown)


Fabien-Chouteau (approved):

doc/user-changes.md
```diff
@@ -4,6 +4,20 @@ This document is a development diary summarizing changes in `alr` that notably
 affect the user experience. It is intended as a one-stop point for users to
 stay on top of `alr` new features.
 
+### Show release-specific dependency sets in solutions
```
Fabien-Chouteau, line 7: BTW @mosteo, this is now live on the website: https://alire.ada.dev/docs/#user-facing-changes-log

mosteo: Great! I had indeed forgotten about updating the website... :S

PR #456, opened 2020-06-23
Set up environment before post-fetch actions

mosteo:
Fixes on top of #452 to ensure the environment is ready during post-fetch actions. Also, generate proper paths for softlinked dependencies.

This PR refactors `Alr.Build_Env` into `Alire.Environment` so environment generation is not split into two locations.


src/alire/alire-roots.ads
```diff
@@ -137,11 +153,30 @@ private
...
+         +This.Path
+      elsif This.Solution.State (Crate).Is_Solved then
+          (+This.Path)
+         / Paths.Working_Folder_Inside_Root
+         / Paths.Dependency_Dir_Inside_Working_Folder
+         / Release (This, Crate).Unique_Folder
+      elsif This.Solution.State (Crate).Is_Linked then
+         This.Solution.State (Crate).Link.Path
+      else
+         raise Program_Error with "release must be either solver or linked");
```
Fabien-Chouteau: `s/solver/solved/`?

src/alire/alire-environment.adb
```diff
@@ -269,6 +315,8 @@ package body Alire.Environment is
    procedure Export (This : Context) is
    begin
       for Var of This.Compile loop
+         Trace.Debug ("Exporting environment: "
```
Fabien-Chouteau: There is already a debug message in `Setenv`.

src/alr/alr-commands-build.adb
```diff
@@ -4,23 +4,36 @@ with Alire.Properties.Actions.Executor;
...
+   -------------
+
+   function Execute (Export_Build_Env : Boolean) return Boolean is
    begin
       Requires_Full_Index;
 
       Requires_Valid_Session;
 
-      Alr.Build_Env.Export (Alr.Root.Current);
+      if Export_Build_Env then
```
Fabien-Chouteau, line 34: If would be nice to export the environment just when we need it, when spawning a command.

In python it is doable with `popen`, but I don't know any similar API for Ada: https://docs.python.org/3/library/subprocess.html#subprocess.Popen

Fabien-Chouteau:
Looks like a great cleanup, thanks @mosteo :)

src/alire/alire-environment.adb
mosteo: Good catch

src/alr/alr-commands-build.adb
mosteo: Ah, I see. That's too bad we don't have something portable like that in GNAT or GNATCOLL...

mosteo:
@Fabien-Chouteau, could you please take a look at the two last commits I pushed? These introduce changes since you reviewed.

I'm going to check if something like `popen` is doable portably next, but my idea is to merge this PR and #452 simultaneously ASAP and if so make another PR later for that.

mosteo:
It seems an option could be to rely on [Boost.Process](https://www.boost.org/doc/libs/1_73_0/doc/html/process.html). According to their docs, this part of Boost is headers-only, so conceivably we could use it directly without changing anything in our build process. I haven't done any testing though.

Fabien-Chouteau:
> I'm going to check if something like `popen` is doable portably next

Don't worry about that, I was thinking out loud. What we have now is good I think.

src/alire/alire-releases.adb
```diff
@@ -426,7 +426,10 @@ package body Alire.Releases is
          for File of Files loop
             if Contains (File, "/") then
                Paths.Include
-                 (File (File'First .. Fixed.Index (File, "/", Backward)));
+                 (File (File'First .. Fixed.Index (File, "/", Backward) - 1));
```
Fabien-Chouteau, line 429: Are you sure there is always a `/` at the end of this String?

Fabien-Chouteau: `if Contains (File, "/") then`, my bad...

PR #457, opened 2020-06-23
Rework Solutions.Diffs internals to use States

mosteo:
Now that the Solution type stores information for all dependencies via Dependencies.States, making the diffs directly from the solution is both simpler, more maintainable and more accurate.

This PR reworks the implementation of `Alire.Solutions.Diffs` without changing the interface.

New tests for the expected output are included.

### DIFF
--- a/src/alire/alire-solutions-diffs.ads
+++ b/src/alire/alire-solutions-diffs.ads
@@ -1,20 +1,6 @@
 with Semantic_Versioning.Extended;
 
-private with Ada.Containers.Indefinite_Ordered_Maps;
-
 package Alire.Solutions.Diffs is
-
-   type Changes is
-     (Added,      -- A new release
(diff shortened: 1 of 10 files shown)


src/alire/alire-solutions-diffs.ads
```diff
@@ -1,28 +1,14 @@
...
-     );
-
    type Diff is tagged private;
    --  Encapsulates the differences between two solutions
 
    function Between (Former, Latter : Solution) return Diff;
    --  Create a Diff from two solutions
 
-   function Change (This : Diff; Crate : Crate_Name) return Changes;
+   --  function Change (This : Diff; Crate : Crate_Name) return Changes;
```
Fabien-Chouteau: This should probably be removed.

Fabien-Chouteau (approved):

PR #458, opened 2020-06-26
implementation of an automatic gpr file with

Fabien-Chouteau:
When adding or removing dependency with `alr with`, a list of `with` statement
for each project files of the dependencies can be automatically added to the
root project file:
```
-- begin auto-gpr-with --
--  This section was automatically added by Alire
with "libhello.gpr";
-- end auto-gpr-with --
project Test is
...
```

This feature can be permanently enabled or disabled with a local or global
configuration option:
```bash
alr config --global --set auto-gpr-with false
```

Crates with project files not compatible with this feature can disable it using
the `auto-gpr-with` entry:
```toml
auto-gpr-with=false
```

ref #313


doc/catalog-format-spec.md
```diff
@@ -350,6 +350,13 @@ entries:
    # An explicit empty case alternative, which is not mandatory
    ```
 
+ - `auto-gpr-with`: Boolean value that specifies if the project files of a
+   crate can be automatically 'withed' by the root project file (default is
+   true). This feature is meant to simply the process of using dependencies in
```
mosteo: s/simply/simplify/

src/alire/alire-properties-from_toml.ads
```diff
@@ -26,7 +27,8 @@ package Alire.Properties.From_TOML with Preelaborate is
                           Notes,
                           Project_Files,
                           Website,
-                          Tags);
+                          Tags,
+                          Auto_GPR_With);
```
mosteo: No need to change but for future occasions: my OCD is appeased by having these enums in alphabetical order if there's no reason for other ordering :-)

.github/workflows/ci-linux.yml
```diff
@@ -9,7 +9,7 @@ on:
       - '**.txt'
 
 env:
-  alire_index: ""
+  alire_index: "git+https://github.com/alire-project/alire-index@devel-0.3"
```
mosteo: This should not be necessary because the index is determined by `Alire.Index.Community_Branch`, so this will result in extra maintenance for no gain. In fact, it's a remainder of when we could not select an index in any other way. 

So, is it giving you some trouble that I'm missing?

testsuite/drivers/helpers.py
```diff
@@ -5,21 +5,6 @@
...
-    Assert that the `filename` text file contains at least one line that
-    contains `line`.
-    """
-    with open(filename, 'r') as f:
-        for l in f:
-            if l.rstrip() == line:
-                break
-        else:
-            assert False, 'Could not find {} in {}'.format(
-                repr(line), filename)
```
mosteo, line 20: Don't we want this one just in case for the future? Or is there something wrong with it?

.github/workflows/ci-linux.yml
Fabien-Chouteau: Actually I made several attempts to get the testsuite to run, so this might not be necessary in the end.

mosteo:
Looks great overall, Fabien, thanks. I just have a couple of doubts that I guess I'm missing by being out of context.

testsuite/drivers/helpers.py
Fabien-Chouteau: Ah yes, this should be in a separate commit, but this function is actually a duplicate. There is exactly the same one defined in the same file.

Fabien-Chouteau:
> Looks great overall, Fabien, thanks.

Thank you, it's a nice to have feature.

> I just have a couple of doubts that I guess I'm missing by being out of context.

Are you talking about the comments in your review, or do you have other doubts?

mosteo:
> Are you talking about the comments in your review, or do you have other doubts?

Just the comments you already answered :+1:

mosteo (approved):

PR #462, opened 2020-07-03
A couple of improvements on the handling of project files and paths

Fabien-Chouteau:
In particular the root dir of a crate is only included if there is a project file in it.


Fabien-Chouteau:
Damn! I made a mistake and merge this to the `feat/env-actions` branch :facepalm:

PR #463, opened 2020-07-03
A couple of improvements on the handling of project files and paths

Fabien-Chouteau:
In particular the root dir of a crate is only included if there is a project file in it.


src/alire/alire-roots.adb
```diff
@@ -141,7 +133,20 @@ package body Alire.Roots is
...
-      return Paths;
+      --  To match the output of root crate paths and Ada.Directories full path
+      --  normalization, a path separator in the last position is removed.
+      return Result : Utils.String_Set do
+         for Path of Paths loop
+            if Path'Length /= 0
+              and then
+               Path (Path'Last) = GNAT.OS_Lib.Directory_Separator
+            then
+               Result.Include (Path (Path'First .. Path'Last - 1));
```
mosteo: We expect all paths in the index to use `'/'`, so this comparison should be against that character and not the system separator. This is the reason for the failing test.

I don't think a native separator can creep up here in some cases but I'm not 100% sure. Perhaps we should check both just to be on the safe side.

testsuite/tests/auto-gpr-with/gpr_in_subdir/test.py
```diff
@@ -0,0 +1,25 @@
+"""
+Test the GPR_PROJECT_PATH for a project file in sub dirs
+"""
```
mosteo: Is this actually what the test is checking?

mosteo (changes requested):

testsuite/tests/auto-gpr-with/gpr_in_subdir/test.py
Fabien-Chouteau: Yes, but it could be clearer.

When the crate declares `dir1/dir2/dir3/prj.gpr`, the `with` statement has to be `with "prj.gpr"`.
So without the sub-dirs because `GPR_PROJECT_PATH` is set with dirs that contain the project file.

src/alire/alire-roots.adb
Fabien-Chouteau: Right, do we check the paths when loading the TOML? I guess we won't be able to do it everywhere but for project-files it is clear that a file path must be specified.

testsuite/tests/auto-gpr-with/gpr_in_subdir/test.py
Fabien-Chouteau: Yeah, it tests the `with` more than the `GPR_PROJECT_PATH` :)

src/alire/alire-roots.adb
mosteo: At some point in time there was an `Alire.Platform_Independent_Path` with an assertion that all separators were forward slashes, intended to make explicit that these paths were to be later translated to the platform when used. I can't find it anymore so it went away in some refactoring (I guess it was not used consistently).

Cursory looking in `Alire.Properties.Labeled` shows no explicit checks for these paths, so I guess that's a hole to be closed. I don't think this is either documented explicitly in the index spec.

mosteo (approved):

mosteo:
Thanks! Always adding the root dir in particular wasn't very clean.

PR #464, opened 2020-07-06
New `alr with --versions` for version-related info of dependencies

mosteo:
This new switch focuses on version-related information for crates in the solution: which constraints were found for each dependency, which version was used to fulfill them, and what's the latest known version for a crate (hence pointing to possible wanted updates).

Example of output (from the included test):
```
CRATE      DEPENDENCY      SOLVED  LATEST 
hello      ^1              1.0.1   4.0.0  
libhello   (^1.0) & (~1.0) 1.0.1   2.0.0  
superhello *               1.0.0   1.0.0  
unobtanium *               missing unknown
wip        *               /fake   unknown
xxx        (root)          0.0.0   unknown
```
The 2nd column info is shown when adding a new dependency, but cannot be obtained again at a later time. The 4th column info is new and would require individual queries with `alr show` to be obtained.

### DIFF
--- a/doc/user-changes.md
+++ b/doc/user-changes.md
@@ -4,6 +4,24 @@ This document is a development diary summarizing changes in `alr` that notably
 affect the user experience. It is intended as a one-stop point for users to
 stay on top of `alr` new features.
 
+### New `alr with --versions` switch
+
+PR [#464](https://github.com/alire-project/alire/pull/464).
+
+A new `alr with --versions` switch is available to obtain version-focused
+information of dependencies. Namely, the combined dependencies on a crate are
+shown, with the release in the solution, and the last known version for the
+crate:
+```
+CRATE      DEPENDENCY      SOLVED  LATEST 
+a_project  (root)          0.0.0   unknown
+hello      ^1              1.0.1   4.0.0  
+libhello   (^1.0) & (~1.0) 1.0.1   2.0.0  
+superhello *               1.0.0   1.0.0  
+unobtanium *               missing unknown
+wip        *               /fake   unknown
+```
+
 ### New `alr with --graph` and `alr with --tree` switches
 
 PR [#465](https://github.com/alire-project/alire/pull/465).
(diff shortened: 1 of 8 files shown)


doc/user-changes.md
```diff
@@ -4,6 +4,24 @@ This document is a development diary summarizing changes in `alr` that notably
...
+
+PR [#464](https://github.com/alire-project/alire/pull/464).
+
+A new `alr with --versions` switch is available to obtain version-focused
+information of dependencies. Namely, the combined dependencies on a crate are
+shown, with the release in the solution, and the last known version for the
+crate:
+```
+CRATE      DEPENDENCY      SOLVED  LATEST 
+a_project  (root)          0.0.0   unknown
```
Fabien-Chouteau, line 17: Why the latest version is not `0.0.0` here?

doc/user-changes.md
```diff
@@ -4,6 +4,24 @@ This document is a development diary summarizing changes in `alr` that notably
...
+hello      ^1              1.0.1   4.0.0  
+libhello   (^1.0) & (~1.0) 1.0.1   2.0.0  
+superhello *               1.0.0   1.0.0  
+unobtanium *               missing unknown
```
Fabien-Chouteau, line 21: What is the difference between `missing` and `unknown`?

Fabien-Chouteau (approved):

doc/user-changes.md
mosteo: `missing` means there is no release in the solution, `unknown` that there is no release in the index.

Perhaps `unknown` can be changed to `unindexed` and it will be clearer?

doc/user-changes.md
mosteo: Because `unknown` is used whenever there is no release in the index. 

In this case it may make sense to use the local version of the crate, although I think I prefer `unindexed` to make clear that there's no index information (see the other comment) for the crate.

doc/user-changes.md
Fabien-Chouteau: `unindexed`/`not indexed` sounds better because to me `unknown` is more a failure to get some information.

mosteo: Changed `unknown` to `unindexed`

PR #478, opened 2020-07-21
Script to migrate indexes to per-release files

mosteo:
Most changes we have planned for the publishing workflow will be easier with the index in the new per-release format, so I plan to do this migration first.

This PR (no work in it yet) will provide a Python script to update indexes to the new format. I plan to use Python because the output of its TOML library looks more natural than that of `toml-ada`.


mosteo:
The script seems to properly migrate all of our community crates (see https://github.com/alire-project/alire-index/tree/tmp-0.1) and testsuite indexes.

For future reference: Python's `toml 0.10.1` is buggy with keys starting with dots (like our `'...'`) and multiline strings containing quotes (like our `hello` crate in some tests). Talk about stress testing. So I ended using `rtoml` which with a little hand holding worked as expected (it does not reorder values to ensure atoms are emitted before tables).

Next step is to have an `alr` that can actually read these manifests...

PR #479, opened 2020-07-22
Load new per-release index manifests

mosteo:
The objective in this PR is to start using the new per-release manifests while seamlessly maintaining functionality. Rough list of goals:

- [x] Properly parse new format
- [x] Properly emit new format (during `alr get`, `alr with`)


testsuite/fix-versions.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-oldversion=0.2
-newversion=0.3
+oldversion=0.4
+newversion=0.1
```
Fabien-Chouteau, line 4: Why going to version `0.1` instead of `0.5`?

testsuite/fixtures/native_index/ma/make/make-external.toml
```diff
@@ -1,9 +1,9 @@
...
+name = "make"
+licenses = []
 maintainers = ["alejandro@mosteo.com"]
 maintainers-logins = ["mylogin"]
-licenses = []
 
 [[external]]
 kind = "system"
-origin = [] # No package name since we require it unavailable in tests
+origin = []
```
Fabien-Chouteau: The comment should probably be preserved here.

testsuite/fixtures/solver_index/ma/make/make-external.toml
```diff
@@ -1,9 +1,9 @@
...
+name = "make"
+licenses = []
 maintainers = ["alejandro@mosteo.com"]
 maintainers-logins = ["mylogin"]
-licenses = []
 
 [[external]]
 kind = "system"
-origin = [] # No package name since we require it unavailable in tests
+origin = []
```
Fabien-Chouteau: Same here.

testsuite/fix-versions.sh
mosteo: To avoid confusing potential crate contributors I when to `tmp-0.1` instead of `devel-0.x` for the migrated community index. In any case once the PRs are to be merged I planned to use the proper `devel-0.x` one. I forgot to communicate this.

testsuite/fixtures/native_index/ma/make/make-external.toml
mosteo: Oh right. I'm now concerned about other comments we will have lost. I will check it out.

mosteo: Barring error, I have recovered all meaningful comments in the test suite and community index

testsuite/fix-versions.sh
Fabien-Chouteau: ok :+1:

PR #484, opened 2020-07-28
Incrementally edit manifest on `alr with`

mosteo:
This is the change described in #466 in which `alr with` doesn't regenerate the whole manifest but only appends new dependencies at the end.

Adding dependencies is trivial; removal is allowed for static dependencies (the kind that can be added via `alr with` anyway) and will fail for a dynamic dependency introduced manually or any other unforeseen circumstance. Still, dynamic and static dependencies can now coexist in a manifest and `alr with` will not complain, unlike before where a single dynamic dependency would preclude `alr with` from working anymore (the manifest could not be regenerated).

Along the way, this PR introduces a couple of supporting types to allow safe editing of the manifest, in the sense that any error while adding/removing dependencies to the manifest will preserve the original manifest, untouched, just in case.

Also, the changes surfaced the circumstance that solved dependencies are always shown in alphabetical order, whereas direct unsolved dependencies were shown in their loading order. The latter have been fixed for consistency.

### DIFF
--- a/testsuite/tests/with/dynamic-dependencies/test.py
+++ b/testsuite/tests/with/dynamic-dependencies/test.py
@@ -1,0 +1,57 @@
+"""
+Check that dynamic dependencies don't affect the use of `alr with`
+"""
+
+import os
+import re
+
+from drivers.alr import run_alr
+from drivers.asserts import assert_eq, assert_match
+from drivers.helpers import content_of
+
+manifest = "alire/xxx.toml"
+
+run_alr('init', '--bin', 'xxx')
+os.chdir('xxx')
+
+# Manually add a regular and a dynamic dependency
+with open(manifest, 'a') as file:
+    file.write('[[depends-on]]\n'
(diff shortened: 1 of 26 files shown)


testsuite/tests/with/dynamic-dependencies/test.py
```diff
@@ -0,0 +1,57 @@
+"""
+Check that dynamic dependencies don't affect the use of `alr with`
```
Fabien-Chouteau, line 2: Looks like there is twice the same test.

Fabien-Chouteau (changes requested):

testsuite/tests/with/dynamic-dependencies/test.py
mosteo: Forgot to save :sweat_smile:. Should be fixed as of 1a3b6d8.

Fabien-Chouteau (approved):

PR #485, opened 2020-07-31
Use manifest provenance (local/index) for finer diagnostics (was: metadata-version new property)

mosteo:
The current changes to the manifest make me think that local manifests (in working copies, unindexed pinned remotes) may become obsolete as we release new stable versions of `alr`.

We currently report a generic error when the local manifest is not loadable. With the introduction of a metadata version (which will match the current index version for new manifests) we will be able to:

- As of this PR, point precisely to the user that the problem is with obsolete metadata and not some vague wrong syntax/fields.
- Eventually, we might offer to migrate manifests across stable releases (hoping we need not to change much).


Fabien-Chouteau:
I understand your goal here, but I wonded if we shouldn't first have the goal of not making breakchanges in the manifest once v1.0 is out.

And if we really have to, the `metadata-version` can be introduced later. The absence of it meaning that the manifest uses metadata version 1.0.

mosteo:
Fair enough. I will change the PR so the field is not emitted nor allowed for now. There are peripheral changes that should be merged (like the local/index manifest provenance) for reuse in followup PRs, so I will ping you when this can be reviewed again.

mosteo:
@Fabien-Chouteau This one should be ready for review again, thanks! I have removed the metadata version changes and stored the tests just in case in a new `testsuite/disabled` folder.

Fabien-Chouteau (approved):

PR #486, opened 2020-08-03
New Alire.Roots.Optional instead of Is_Valid

mosteo:
This is a cleanup to avoid the `Root.Is_Valid` smell. Alire.Roots.Root now always describes a valid root, that may exist or not on disk yet (no change there). For the case where we may have a root or not, and Outcome wrapper over Root is defined in Alire.Roots.Optional.

The change means that where before one would check a valid root with Root.Current.Valid, now it is done with Root.Current.Exists. Where we require a root, nothing changes and the absence of root will cause a Checked_Error. Where the root is optional, the client code must deal with it explicitly like before, but using the new Roots.Optional (which is also retrievable with overloaded Root.Current).


mosteo:
I got confused and ended creating the same PR twice. The new #492 describes better the final result so I'm closing this one.

PR #494, opened 2020-08-07
 Move manifest and lock files to top-level folder

mosteo:
The manifest and lock files expected location is changed:

 - Name is always `alire.toml`, `alire.lock` (crate name is now inside the manifest).
 - They're at top-level instead of inside `${crate_root}/alire`.

This way, there's less confusion that these files go under version control and the alire folder is to be ignored.

This was originally #489 but somehow it got wrongly merged due to changes to the base branch.


Fabien-Chouteau (approved):

mosteo:
@Fabien-Chouteau, I just realized we have the local configuration inside `alire`. Should we move than one out too (e.g., as `alire.conf`)?

Fabien-Chouteau:
Good question. If we compare, `git` local config is not exported to the remote repo.
So I would just keep it in `alire/` for the moment.

But what we can do is rename it into `config.toml`, no more clash with `<crate>.toml`.

mosteo:
Good thinking.

PR #500, opened 2020-09-01
Store backed up files inside `alire` folder

mosteo:
Instead of creating the ".prev" backup files in place, we move them into the `alire` folder so they don't pollute the user workspace.

### DIFF
--- a/src/alire/alire-directories.adb
+++ b/src/alire/alire-directories.adb
@@ -11,12 +11,18 @@ package body Alire.Directories is
    -- Backup_If_Existing --
    ------------------------
 
-   procedure Backup_If_Existing (File : Any_Path) is
+   procedure Backup_If_Existing (File   : Any_Path;
+                                 Base_Dir : Any_Path := "")
+   is
       use Ada.Directories;
+      Dst : constant String := (if Base_Dir /= ""
+                                then Base_Dir / Simple_Name (File) & ".prev"
+                                else File & ".prev");
    begin
       if Exists (File) then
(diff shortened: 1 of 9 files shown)


src/alire/alire-directories.adb
```diff
@@ -11,12 +11,18 @@ package body Alire.Directories is
...
+      Dst : constant String := (if Base_Dir /= ""
+                                then Base_Dir / Simple_Name (File) & ".prev"
+                                else File & ".prev");
    begin
       if Exists (File) then
-         Trace.Debug ("Backing up " & File);
-         Copy_File (File, File & ".prev", "mode=overwrite");
+         Trace.Debug ("Backing up " & File
+                      & " with base dir: " & Base_Dir);
+         Copy_File (File, Dst, "mode=overwrite");
```
Fabien-Chouteau, line 25: I think it's better to keep the `.prev` or `.back` so the file is not confusing for users.

Fabien-Chouteau (changes requested):

src/alire/alire-directories.adb
mosteo: If I'm understanding correctly, that's what's happening; the files are copied as before with the extra `.prev` but put inside `alire`.

Fabien-Chouteau: My bad, the `.prev` is added above.

Sorry for the noise.

Fabien-Chouteau (approved):

PR #502, opened 2020-09-01
[RFC] Replace `'...'` with `others` in manifest files

mosteo:
In the context of manifest refactorings this one popped into my head. I know we are not striving anymore to copy Ada syntax in the index, but I wonder if it would be more intuitive to use `others` instead of `'...'` for case expressions in toml manifests. The reasons: familiar to Adaists, it avoids the problem of forgetting quotes, it does not use a symbol that has a different meaning (ranges) in Ada.


Fabien-Chouteau:
There is a risk of `others` being a value of the case that users want to match.

mosteo:
We are the ones defining values for these environment enums, so in principle we could always avoid the clash.

Fabien-Chouteau:
I don't know, we might want to match on more stuff in the future.

If we say that this is pure string matching then we can allow regex and `...` becomes `.*`.

mosteo:
OK, if this isn't a clear change let's keep it as it is.

PR #505, opened 2020-09-02
Merge AEP-2-derived changes to manifests into master

mosteo:
Once merged:
- [x] Make `devel-0.4` the default branch for `alire-index`


mosteo:
This one bring in all the changes of late that were kept separate as most are breaking, so we only have one big break for our users.

I wonder if we should have some mitigation measure to help our current users. We could autodetect a manifest in the old location and stop with an explanation instead of erroring with the generic "cannot continue with invalid session"...

Fabien-Chouteau:
> I wonder if we should have some mitigation measure to help our current users. We could autodetect a manifest in the old location and stop with an explanation instead of erroring with the generic "cannot continue with invalid session"...

It doesn't sound too hard to do.

mosteo:
> It doesn't sound too hard to do.

Right, so I'll get on it.

Fabien-Chouteau (approved):

Fabien-Chouteau:
:partying_face: That is a huge milestone. Congrats @mosteo

PR #506, opened 2020-09-02
Warn of obsolete metadata and provide some guidance

mosteo:
Show minimal information to reset a workspace and provide a link for more detailed information on changes and how to migrate.

### DIFF
--- a/src/alire/alire-directories.adb
+++ b/src/alire/alire-directories.adb
@@ -1,11 +1,52 @@
 with Ada.Exceptions;
 with Ada.Numerics.Discrete_Random;
+with Ada.Text_IO;
 with Ada.Unchecked_Deallocation;
 
 with Alire.Properties;
 with Alire.Roots;
 
 package body Alire.Directories is
+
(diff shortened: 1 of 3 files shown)


src/alire/alire-directories.adb
```diff
@@ -92,6 +133,10 @@ package body Alire.Directories is
            Kind (Possible_Root.Crate_File) = Ordinary_File
          then
             return Path;
+         elsif GNAT.OS_Lib.Is_Directory ("alire") and then
+           Directories.Find_Single_File ("alire", "toml") /= ""
```
Fabien-Chouteau: Will this match with `config.toml`?

mosteo: For an old workspace no, because it won't exist, and in a new one it won't reach that branch because it will have detected `alire.toml` in the previous condition.

But since I haven't tested this exhaustively I'm going to add a test to exclude that possibility just in case.

Fabien-Chouteau (approved):

PR #511, opened 2020-09-05
Accept whole answers in interactive questions

### LINKED ISSUE #509: Interactive questions should accept 'yes' and 'no' as valid answers
With the new **alr with** command, we have some interactive mode and some questions are asked.

```
Do you want Alire to automatically update your project file with the new dependency solution?
[Y] Yes  [N] No  (default is Yes) yes
Invalid answer.
```

It would be nice to accept the long forms such as `yes` or `NO` as valid answers because other Linux interactive tools allow you to do so.

NicolaiRuckel:
This should fix #509.

I hope this follows your code style. It's a long time since I wrote Ada code.

### DIFF
--- a/src/alire/alire-utils-user_input.adb
+++ b/src/alire/alire-utils-user_input.adb
@@ -15,6 +15,16 @@ package body Alire.Utils.User_Input is
       No     => 'N',
       Always => 'A');
 
+   -----------------
+   -- Answer_Kind --
+   -----------------
+
+   function Answer_String (Kind : Answer_Kind) return String
+   is (case Kind is
+       when Yes    => "yes",
+       when No     => "no",
+       when Always => "always");
+
    ------------
    -- Is_TTY --
    ------------
@@ -114,6 +124,16 @@ package body Alire.Utils.User_Input is
                   end if;
                end loop;
             end if;
+
+            --- Check if the user input the whole answer
+            for Kind in Answer_Kind loop
+               if Valid (Kind)
+                 and then
+                   Char.To_Lower (Input) = Answer_String (Kind)
+               then
+                  return Kind;
+               end if;
+            end loop;
 
             TIO.Put_Line ("Invalid answer.");
          end;


NicolaiRuckel:
I'm not sure why the Windows CI fails but since it fails in the Install toolchain step, it doesn't look like it would be my fault.

mosteo:
Hi Nicolai, thanks for your contribution. The error with the toolchain installation happens from time to time and it's caused by something getting stuck in the github infrastructure. I relaunched the check and this time it's an actual test failing, but I haven't tried to ascertain the problem yet.

src/alire/alire-utils-user_input.adb
```diff
@@ -15,6 +15,17 @@ package body Alire.Utils.User_Input is
       No     => 'N',
       Always => 'A');
 
+   type String_Access is access constant String;
```
Fabien-Chouteau: Instead of using pointers here you can define and expression function to return the String:
```ada
   function Answer_String (Kind : Answer_Kind) return String
   is (case Kind is
       when Yes    => "yes",
       when No     => "no",
       when Always => "always");
```

src/alire/alire-utils-user_input.adb
```diff
@@ -115,6 +126,19 @@ package body Alire.Utils.User_Input is
                end loop;
             end if;
 
+            --- Check if the user input the whole answer
+            if Input'Length >= 1 then
```
Fabien-Chouteau: I think you can remove this check, does not seem necessary to me.

Fabien-Chouteau (commented):
Thank for the contribution @NicolaiRuckel!

I added a couple comments.

src/alire/alire-utils-user_input.adb
NicolaiRuckel: That looks a lot cleaner to me, too. I replaced my code with your function.

src/alire/alire-utils-user_input.adb
NicolaiRuckel: I agree and removed it.

Fabien-Chouteau (approved):

Fabien-Chouteau:
Thank you @NicolaiRuckel 

@mosteo you can merge.

mosteo:
Merged, thanks!

PR #512, opened 2020-09-07
Minimal publishing assistant

mosteo:
This PR introduces the bare minimum assistance (via `alr publish --prepare`) to create an index manifest. It is intended to fit in the larger scheme described in #227.

At the moment there are no checks ~and it only accepts repositories as remotes~, so it is not intended for immediate merging.


mosteo:
@Fabien-Chouteau , if you want to take a look at this one. With the general scaffolding in place I think it should be quite straightforward to add missing checks/steps from #227 on top of this one, in case you want to tackle any of the many things missing.

My immediate plans are to complete the equivalent for source archives to what is in this one for repositories, and add the whitelist for repositories.

mosteo:
The changes to support source archives are minimal so I'm appending them to this PR.

Test with:

`alr publish https://github.com/mosteo/aaa/archive/v0.2.zip`

or

`alr publish https://github.com/mosteo/aaa.git --commit 7376b76f23ab4421fbec31eb616d767edbec7343`

src/alire/alire-publish.adb
```diff
@@ -0,0 +1,257 @@
...
+         Open (Index_File, Append_File, Index_Manifest);
+         New_Line (Index_File);
+         TOML.File_IO.Dump_To_File (TOML_Origin, Index_File);
+         Close (Index_File);
+
+         Log_Success
+           ("Your index manifest file has been generated at "
+            & TTY.URL (Index_Manifest));
+         Log_Success
+           ("Please place it at " & TTY.URL (TOML_Index.Manifest_Path (Name))
```
Fabien-Chouteau: Here we can use the GitHub upload link, for instance:
```
Upload the manifest using this link:
https://github.com/alire-project/alire-index/upload/devel-0.4/index/li/libhello/
```

src/alire/alire-publish.ads
```diff
@@ -0,0 +1,12 @@
+with Alire.Origins;
+
+package Alire.Publish is
+
+   procedure Verify_And_Create_Index_Manifest (Origin : URL;
+                                               Commit : String := "");
+   --  Requires a remote URL to a source file or a git repository. Commit is
+   --  mandatory in the latter case. Produces a file `crate-version.toml` in
+   --  the current directory or raises Checked_Error with the appropriate error
```
Fabien-Chouteau, line 9: Maybe the release .toml could be in an `alire/releases/` folder.

src/alr/alr-commands-publish.adb
```diff
@@ -48,10 +63,23 @@ package body Alr.Commands.Publish is
...
+      Define_Switch
+        (Config,
+         Cmd.Commit'Access,
+         "", "--commit=",
+         "Revision in a remote repository denoting the sources of a release",
+         Argument => "ID");
+
       Define_Switch (Config,
                      Cmd.Hash'Access,
                      "", "--hash",
```
Fabien-Chouteau: I am not sure we still need the `--hash` feature here since it only does a sha512.

src/alire/alire-publish.adb
mosteo: Neat, I see it creates the PR and it seems to work even if the directory doesn't exist yet.

src/alr/alr-commands-publish.adb
mosteo: Yes, we can deprecate this.

Fabien-Chouteau:
> `alr publish https://github.com/mosteo/aaa.git --commit 7376b76f23ab4421fbec31eb616d767edbec7343`

Can we have `alr publish https://github.com/mosteo/aaa.git#commit=7376b76f23ab4421fbec31eb616d767edbec7343` I know this works for Arch's PKGBUILD?

Or just drop the `--commit` and have:
`alr publish  https://github.com/mosteo/aaa.git 7376b76f23ab4421fbec31eb616d767edbec7343`

We could also try to accept tags and branches as arguments to `alr publish`, but then convert them into commit id in the manifest.

And `alr publish` (no argument) that looks into the current repo:
 - Check if root is a git repo
 - Check if `origin` remote is in the allow-list of hosting platforms (GitHub, GitLab, Source Forge, etc.)
 - Check if there is local change
 - Use the `origin` and head commit of current branch to do the same as `alr publish <ORIGIN> --commit <COMMIT>`

src/alire/alire-publish.adb
Fabien-Chouteau: Yep, it's perfect for us :)

mosteo:
> We could also try to accept tags and branches as arguments to `alr publish`, but then convert them into commit id in the manifest.
> 
> And `alr publish` (no argument) that looks into the current repo:

Very practical. I will file these as separate enhancements.

mosteo:
@Fabien-Chouteau , this one should be complete. I think everything that was commented is either in or in one of the follow-up PRs.

I have a problem with `Ada.Directories.Delete_Tree` failing to delete a repository, only on Windows. Does it ring any bells?

Fabien-Chouteau:
I will look at the code as soon as possible.

> @Fabien-Chouteau , this one should be complete. I think everything that was commented is either in or in one of the follow-up PRs.

I tried in a git repo:
```
PS C:\Users\chouteau\Documents\GitHub\shoot-n-loot> alr publish --prepare
ERROR: Origin URL argument required.
```

> I have a problem with `Ada.Directories.Delete_Tree` failing to delete a repository, only on Windows. Does it ring any bells?

To delete a directory is always tricky on Windows.  Maybe GNATcoll VFS works better here. @Nikokrock do you know if there is a good remove dir implementation for Windows in Ada?

mosteo:
> I tried in a git repo:
> 
> ```
> PS C:\Users\chouteau\Documents\GitHub\shoot-n-loot> alr publish --prepare
> ERROR: Origin URL argument required.
> ```

Yup, this will not work until #519. 

To clarify, this one only works with an explicit remote origin.

> To delete a directory is always tricky on Windows. Maybe GNATcoll VFS works better here.

Thanks for the suggestion, will try.

Fabien-Chouteau (approved):

PR #513, opened 2020-09-07
Dim output of spawned subprocesses

mosteo:
This way the output of `alr` stands out over that of child processes (builds, clones, fetches).

### DIFF
--- a/src/alire/alire-os_lib-subprocess.adb
+++ b/src/alire/alire-os_lib-subprocess.adb
@@ -1,5 +1,7 @@
 with Ada.Text_IO;
 
+with ANSI;
+
 with GNAT.OS_Lib;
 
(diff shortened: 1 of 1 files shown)


Fabien-Chouteau:
I will have to check what this looks like on Windows.

Fabien-Chouteau:
This is what is looks like on Windows PowerShell:
```
[2mCloning into '/c/Users/Fabien Chouteau/.config/alire/indexes/community/repo'...
remote: Enumerating objects: 54, done.
```
So the feature should be disabled on Windows.

mosteo:
Of course, my bad. I forgot that by going to ANSI directly it's not following our disabling rules :facepalm:

Fabien-Chouteau (approved):

PR #514, opened 2020-09-07
Initial list of trusted code hosts

mosteo:
Check VCS origins against a list of trusted repository hostings.

I'm probably missing other known ones.


Fabien-Chouteau (approved):
In the future we will probably have requests to be able to disable this locally.
We don't have to worry about this for the moment.

Fabien-Chouteau (approved):

PR #517, opened 2020-09-07
Test build the publishing release

mosteo:


src/alire/alire-publish.adb
```diff
@@ -59,6 +60,40 @@ package body Alire.Publish is
...
+   begin
+      --  We need the alire folder, which usually will not be among sources
+      if not Ada.Directories.Exists (Paths.Working_Folder_Inside_Root) then
+         Ada.Directories.Create_Directory (Paths.Working_Folder_Inside_Root);
+      end if;
+
+      --  Unfortunately, building is one of the parts that still are hard to
+      --  extricate from Alr.*, so for now the simplest solution is to spawn
+      --  a child alr.
+      OS_Lib.Subprocess.Checked_Spawn
```
Fabien-Chouteau, line 88: I have projects that cannot be built standalone, GESTE for example. So it should be possible to ignore this check.

Fabien-Chouteau (approved):

src/alire/alire-publish.adb
mosteo: I was thinking the same if only because some builds take a long time. For now I'll add a switch to turn this entirely off. Later, I've thought that here we should run any test actions instead, defaulting to this.

PR #518, opened 2020-09-08
Fix compilation warnings with current versions of GNAT

yakobowski:


mosteo:
Hello @yakobowski, thanks for the PR. Out of curiosity, what GNAT version are you using to compile? We have the option to treat warnings as errors enabled so that means these (clearly unneeded) `with`s should be causing our CI to fail.

yakobowski:
I'm using AdaCore's current GNAT. This warning has been improved in the last month.

(There are still problems to get a fully clean build, but the last thing to fix is in `gnatcolls-slim`. I will try to submit a PR there, and then update the submodule here.)

mosteo:
Cool, thanks for the info.

mosteo (approved):

PR #523, opened 2020-09-11
alr init: add compiler option controls to template project

Fabien-Chouteau:
User can control compile checks, run-time checks, contracts, style checks.

### DIFF
--- a/src/alr/alr-commands-init.adb
+++ b/src/alr/alr-commands-init.adb
@@ -121,6 +121,7 @@ package body Alr.Commands.Init is
          Put_Line ("      when ""enabled"" => null;");
          Put_Line ("         Style_Checks_Switches :=");
          Put_Line ("           (""-gnatyg"",   -- GNAT Style checks");
+         Put_Line ("            ""-gnaty-d"",  -- Disable no DOS line terminators");
          Put_Line ("            ""-gnatyM80"", -- Maximum line length");
          Put_Line ("            ""-gnatyO"");  -- Overriding subprograms explicitly marked as such");
          Put_Line ("      when others => null;");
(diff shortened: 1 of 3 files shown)


src/alr/alr-commands-init.adb
```diff
@@ -79,30 +81,84 @@ package body Alr.Commands.Init is
...
+         Put_Line ("   end case;");
+         Put_New_Line;
+         Put_Line ("   Build_Switches := ();");
+         Put_Line ("   case Build_Mode is");
+         Put_Line ("      when ""optimize"" =>");
+         Put_Line ("         Build_Switches := (""-O3"",     -- Optimization");
+         Put_Line ("                            ""-gnatn""); -- Enable inlining");
+         Put_Line ("      when ""debug"" =>");
+         Put_Line ("         Build_Switches := (""-g"",   -- Debug info");
+         Put_Line ("                            ""-O0""); -- No optimization");
```
mosteo: What do you think about using `-Og` instead of `-O0`? According to https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html

> It is a better choice than -O0 for producing debuggable code because some compiler passes that collect debug information are disabled at -O0.

Personally, I haven't noticed anything ever but...

src/alr/alr-commands-init.adb
```diff
@@ -63,6 +63,8 @@ package body Alr.Commands.Init is
          Filename : constant String :=
             +Full_Name (Directory / (+Lower_Name & ".gpr"));
       begin
+         --  Use more than 80 colums for more readable strings
+         pragma Style_Checks ("M200");
```
mosteo, line 68: I'm going to copy this trick at some point...

src/alr/alr-commands-init.adb
Fabien-Chouteau: Ok for me

src/alr/alr-commands-init.adb
Fabien-Chouteau: We should avoid it :)

mosteo (approved):

PR #530, opened 2020-09-15
Tweaks for detection of git URLs

mosteo:
- A leading `git+` or trailing `.git` is no longer required for a plain `https://` URL. This is detected for remote URLs on github and gitlab, or any local repository.
- A private `git@github`, `git@gitlab` URL is transformed into its public `https://` counterpart. I've been unable to determine if this transformation is valid in general, so for now there's an explicit list.

### DIFF
--- a/src/alire/alire-vcss-git.adb
+++ b/src/alire/alire-vcss-git.adb
@@ -3,6 +3,7 @@ with Alire.OS_Lib.Subprocess;
 with Alire.Errors;
 with Alire.Utils;             use Alire.Utils;
 with Alire.Utils.Tools;
+with Alire.Utils.TTY;
 
 package body Alire.VCSs.Git is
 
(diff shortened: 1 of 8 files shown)


src/alire/alire-vcss-git.adb
```diff
@@ -260,6 +269,30 @@ package body Alire.VCSs.Git is
       end if;
    end Status;
 
+   -------------------------
+   -- Transform_To_Public --
+   -------------------------
+
+   function Transform_To_Public (Remote : String) return URL is
```
Fabien-Chouteau, line 276: It would be good to have unit testing for this kind of functions.

Fabien-Chouteau (approved):

src/alire/alire-vcss-git.adb
mosteo: We have a custom setup at https://github.com/alire-project/alire/blob/master/src/alire/alire-selftest.adb, should we continue to use that or should we move to some real unit testing framework?

Fabien-Chouteau: Putting in place a unit testing framework would be better, but let's use that for now :+1:

PR #536, opened 2020-09-22
Tool and library version number

Fabien-Chouteau:


mosteo (changes requested):
This was indeed needed, thanks.

Only thing is, I tagged old versions up to `0.6.1`, so I guess the current one should at least be `0.7`.

Fabien-Chouteau:
Will do.

mosteo (approved):

Issue #54, opened 2018-10-04
alr edit command


mosteo:
To directly launch gnatstudio or other editors.

Current implementation is partial, only looking for `gnatstudio`. We want also the ability to open custom editors (which will require some way of defining the expected command line).

mosteo:
Related: #379

mosteo:
Implemented in #545.

Fabien-Chouteau:
We can open another issue, but the current implementation is only partial. We want to support starting user defined editors.

mosteo:
Ah, right. Let's reopen this one.

kevlar700:
It would be nice to be able to set the editor on the fly by overriding the config on the command line, without changing the project default?

Fabien-Chouteau:
In the dev branch we have `alr exec --` which will let you start any editor.

PR #545, opened 2020-09-25
Alr.Commands: new Edit command

Fabien-Chouteau:
Ref #54


testsuite/tests/workflows/edit/test.py
```diff
@@ -0,0 +1,18 @@
+"""
+Test a basic auto withing of gpr file.
+"""
```
mosteo: From another test?

Fabien-Chouteau: Sorry, will fix.

mosteo (approved):

Issue #455, opened 2020-06-23
Use configuration values for new crates (email, name, login, etc.)


Fabien-Chouteau:
I found the code where this needs to go but `Alire.Releases` is pre-elaborate so it cannot depend on `Alire.Config`: 
https://github.com/alire-project/alire/blob/738e4b43cf696f5c387ed3b313390ad189d99428/src/alire/alire-releases.adb#L42

Fabien-Chouteau:
We could also prompt the user for those info when they are not in the config already.

mosteo:
There's the nuclear option of abandoning explicit preelaboration everywhere; for such a big/complex thing like libalire I'm doubtful about any advantages, since parts of the library can't be preelaborable anyway. I've read comments from some ARG lawyers saying that preelaboration pragmas are a broken idea but we're stuck with it. Frankly this is one of these things that irk me about Ada. 

A workaround I've used for this kind of situation in the past is to set up a hook during initialization into a preelaborable package. It's pure code noise though.

Fabien-Chouteau:
Now would be the best time to remove all preelaboration. Do you remember why it was required in the first place?

Fabien-Chouteau:
I did a ``sed -i 's/with Preelaborate //g' on all sources and it seems to be ok. Compiles and testsuite clean.

mosteo:
>  Do you remember why it was required in the first place?

No need in the case of Alire, but that I try to mark everything that's Pure/Preelaborable as such. I guess we lose nothing down the road with half the packages not being preelaborable already.

Fabien-Chouteau:
> I did a ``sed -i 's/with Preelaborate //g' on all sources and it seems to be ok.

Actually it's not when doing a full re-build.

Fabien-Chouteau:
@mosteo can we provide the user info to the sub-program that create the release in `alr init`?
I tried to follow the callstack but I am not sure what is there to change.

mosteo:
> @mosteo can we provide the user info to the sub-program that create the release in `alr init`?
> I tried to follow the callstack but I am not sure what is there to change.

I realize I didn't see/forgot this ping from you, my apologies. As we said this morning I'll try to pick up #461 to use these values during init.

mosteo:
Implemented in #548.

PR #548, opened 2020-09-28
Gather user info during `alr init` for manifest creation

mosteo:
~This is still in flux, depending on previous PRs, so don't review yet (or check only the last commit).~

In the end I find more convenient in this particular instance to create the manifest by hand, since it is simple and intended for user reading, so we can format it as we please. So this PR changes from `To_TOML` export to plain `Put_Line` creation. We still use TOML functions to ensure the strings are properly encoded.

I have reformatted the layout of the `User_Input.Query_String` function, because I found the original interaction a bit confusing. If you want to go back to your previous version just say and I will revert that change.

Implements   #455.

### DIFF
--- a/src/alire/alire-config.ads
+++ b/src/alire/alire-config.ads
@@ -110,6 +110,17 @@ package Alire.Config is
    procedure Print_Builtins_Doc;
    --  Print a Markdown documentation for the built-in configuration options
 
+   package Keys is
+
+      --  A few predefined keys that are used in several places. This list is
+      --  not exhaustive.
+
+      User_Email        : constant Config_Key := "user.email";
+      User_Name         : constant Config_Key := "user.name";
+      User_Github_Login : constant Config_Key := "user.github_login";
+
+   end Keys;
+
 private
 
    function Load_Config_File (Path : Absolute_Path) return TOML.TOML_Value;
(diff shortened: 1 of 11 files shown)


src/alire/alire-config.ads
```diff
@@ -110,6 +110,17 @@ package Alire.Config is
    procedure Print_Builtins_Doc;
    --  Print a Markdown documentation for the built-in configuration options
 
+   package Keys is
+
+      --  A few predefined keys that are used in several places. This list is
+      --  not exhaustive.
+
+      User_Email        : constant Config_Key := "user.email";
```
Fabien-Chouteau, line 118: You can also use those keys in the built-ins below.

src/alire/alire-utils-user_input.adb
```diff
@@ -178,18 +178,19 @@ package body Alire.Utils.User_Input is
...
          end if;
 
-         TIO.Put_Line ("Default: '" & Default & "'");
-
          --  Flush the input that the user may have entered by mistake before
          --  the question is asked.
          Flush_TTY;
 
+         --  Print a prompt
+         TIO.Put ("> ");
```
Fabien-Chouteau: I think `Flush_TTY` should be the last thing before calling `Get_Line`.

Fabien-Chouteau (changes requested):

Fabien-Chouteau (approved):

PR #551, opened 2020-10-01
Create and upload assets to alr releases

mosteo:
When a release is created manually or via tag, create and upload the binaries as assets of the release.

In addition to our usual binary artifacts, an AppImage that should be generally Linux-portable is generated.

See, for example, this test release: https://github.com/mosteo/alire/releases/tag/v0.7-test


.github/workflows/release-assets.yml
```diff
@@ -0,0 +1,286 @@
...
+
+    - name: Check out submodules
+      run: git submodule update --init --recursive
+
+    - name: Pull docker image
+      run: docker pull alire/gnat:${{ matrix.tag }}
+
+    - name: Build
+      run: >
+        docker run -v${PWD}:/alire -w /alire
```
Fabien-Chouteau: I don't like that alr is rebuilt just in this workflow. The binary released is not that same that the one tested.

Can't we have this publishing actions optional in the usual workflow scripts?

Fabien-Chouteau (changes requested):

.github/workflows/release-assets.yml
mosteo: With current workflow limitations it will be doable but ugly. I'll try to come with a decent solution.

Fabien-Chouteau: In that case I prefer a manual upload of the release binaries from the CI runs.

mosteo: What about testing these executables again? That would add only a step per job.

mosteo: > With current workflow limitations it will be doable but ugly.

Actually I'm not sure it's doable, as PR artifacts are totally disconnected from commits/releases, and the commit changes after the merge.

mosteo: > That would add only a step per job.

Actually a few, to set up the virtualenv

mosteo:
I think this does what we want: we reuse the same workflow for PRs and releases, and we reuse the binaries for tests, artifacts and assets. 

Also, since this triggers on creation of the release, we are testing the actual release commit, which we were not testing after merging a PR, and the assets correspond to the commit and the sources provided by github.

This gives me more peace of mind since merging/squashing from branches not rebased always gives me a bit of pause.

An ugly thing is the repeated `if:` entries, but we're stuck with that until github implements [early exit from a job](https://github.com/actions/runner/issues/662).

Fabien-Chouteau (approved):

PR #563, opened 2020-10-09
Update ci-linux.yml

Fabien-Chouteau:
Switch Linux CI to GitHub Ubuntu-18.04 only

### DIFF
--- a/.github/workflows/ci-linux.yml
+++ b/.github/workflows/ci-linux.yml
@@ -19,17 +19,9 @@ env:
 jobs:
 
   build:
-    name: CI on ${{ matrix.tag }}
+    name: CI on Linux
 
-    runs-on: ubuntu-latest
-
-    strategy:
-      matrix:
-        tag: # Those are our dockerhub alire/gnat:tag machines
-            - centos-latest-community-latest
-            - community-latest
-            - debian-stable
-            - ubuntu-lts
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-linux.yml
```diff
@@ -19,37 +19,37 @@ env:
...
       uses: actions/checkout@v2
       with:
         submodules: true
 
-    - name: Pull docker image
-      run: docker pull alire/gnat:${{ matrix.tag }}
+    - name: Install toolchain
+      uses: ada-actions/toolchain@ce2020
+      with:
+        distrib: community
```
mosteo: Why not the FSF one here? I would expect that [some] users building from source could be using that one. And since the CE is already tested on windows/macos...

mosteo:
I think that we should keep at least the docker build for an "unsupported" distro (centos in our tests), in a separate workflow. We added it to detect problems in systems without a native package manager. Although macos currently also doubles as that. But at some point we might support ports on macos...

.github/workflows/ci-linux.yml
Fabien-Chouteau: OK I can change that.

Fabien-Chouteau:
> We added it to detect problems in systems without a native package manager.

That's a good point indeed.

mosteo (approved):

PR #566, opened 2020-10-09
Switch from devel- to stable- index branch

mosteo:
We rely on `stable-0.4` until new features are required in the index format.


Fabien-Chouteau (approved):

Fabien-Chouteau:
Agreed, I think that `devel-` branches in the index should only be temporary until a new `Alire` release is published.

PR #575, opened 2020-10-16
Added alire.toml so alr can build/edit itself

mosteo:
Also bumped the patch number to `0.7.1`, as code changes from now own should appear in a future release, and `0.7-dev` means pre-0.7 according to semantic versioning.

### DIFF
--- a/alire.toml
+++ b/alire.toml
@@ -9,9 +9,9 @@ description = "Command-line tool from the Alire project"
 # builds are uniquely identified without requiring manual editing or other
 # preprocessing. We could also include info about all dependencies exact
 # versions and environment used during the build.
-version = "0.7.1-dev"
+version = "0.8-dev"
 
-authors = ["Alejandro R. Mosteo", "Fabien Chouteau"]
+authors = ["Alejandro R. Mosteo", "Fabien Chouteau", "Pierre-Marie de Rodat"]
 maintainers = ["alejandro@mosteo.com", "chouteau@adacore.com"]
 maintainers-logins = ["mosteo", "Fabien-Chouteau"]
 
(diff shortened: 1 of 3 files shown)


alire.toml
```diff
@@ -0,0 +1,32 @@
...
+# in ./deps, the result of `alr build` might differ from the same code built by
+# GitHub actions. We might want to add, down the road, a way to propagate the
+# version defined here into Ada code (perhaps by generating a metadata package
+# before `alr build` that the code being built can "with" if so desired), so
+# builds are uniquely identified without requiring manual editing or other
+# preprocessing. We could also include info about all dependencies exact
+# versions and environment used during the build.
+version = "0.7.1-dev"
+
+authors = ["Alejandro R. Mosteo", "Fabien Chouteau"]
```
Fabien-Chouteau: \+ Pierre-Marie de Rodat :smiling_face_with_three_hearts:

src/alire/alire.ads
```diff
@@ -9,7 +9,7 @@ with Simple_Logging;
 
 package Alire with Preelaborate is
 
-   Version : constant String := "0.7.0-dev";
+   Version : constant String := "0.7.1-dev";
```
Fabien-Chouteau: I think it should be `0.8.0` (see #564).
If some patches are interesting for a the `0.7.1` release we can cherry-pick them in the `release/0.7-beta` branch.

alire.toml
mosteo: Oh yeah!

src/alire/alire.ads
mosteo: About #564, is still in draft on purpose?

Fabien-Chouteau: I was waiting because we were still merging stuff into `0.7`.

mosteo:
I have added a new workflow that checks that `alr build` succeeds.

Fabien-Chouteau (approved):

PR #592, opened 2020-10-24
Add scenario to enable Ada 2020 features

mosteo:
I was trying to use the new generalized 'Image feature to debug, so this is a
convenient way of having Ada2020 features available without editing the gpr
file. Unfortunately, whatever is enabled by this switch breaks our parser of
manifests (spurious missing properties are reported), so it's not very useful
until that problem is pinpointed.

### DIFF
--- a/alire_common.gpr
+++ b/alire_common.gpr
@@ -25,6 +25,17 @@ abstract project Alire_Common is
       when "disabled" => Style_Check_Switches := ();
    end case;
 
+   type Any_Ada_Mode is ("current", "experimental");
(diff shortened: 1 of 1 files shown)


Fabien-Chouteau:
I don't really understand this PR. Ada2020 is not ready for production in any version of GNAT, so we should not use it at all in Alire for the moment.

mosteo:
Yep. This is because every time I want to use some 202x feature for debugging it requires editing the project file. It's not intended for any public build. Anyway no big deal.

PR #600, opened 2020-10-31
Add handling of arch and arch based distributions

BeChris:
Hello, this pull request to add detection of Arch (and Arch based) distribution
This is done by analyzing value of "id" and "id_like" attributes in /etc/os-release file.

### DIFF
--- a/src/alire/os_linux/alire-platform.adb
+++ b/src/alire/os_linux/alire-platform.adb
@@ -64,6 +64,7 @@ package body Alire.Platform is
 
             --  If no supported distribution found, fallback to id_like key
             if Cached_Distro = Distro_Unknown then
+               Trace.Debug ("Unknown distro for key 'id', falling back to 'id_like'");
                Cached_Distro := Get_Os_Release_Value_For_Key ("id_like");
             end if;
 
--- a/src/alire/os_linux/alire-platform.adb
+++ b/src/alire/os_linux/alire-platform.adb
@@ -41,30 +41,17 @@ package body Alire.Platform is
                   Normalized : constant String :=
                                  To_Lower_Case (Replace (Line, " ", ""));
                begin
-                  begin
-                     if Starts_With (Normalized, "id=") or else Starts_With (Normalized, "id_like=") then
-                        Cached_Distro :=
-                          Platforms.Distributions'Value
-                            (Tail (Normalized, '='));
-                        Distro_Cached := True;
-                        return Cached_Distro;
-                     end if;
-                  exception
-                     when others =>
-                        null; -- Not a known distro.
-                  end;
-                  begin
-                     if Starts_With (Normalized, "id_like=") then
-                        Cached_Distro :=
-                          Platforms.Distributions'Value
-                            (Tail (Normalized, '='));
-                        Distro_Cached := True;
-                        return Cached_Distro;
-                     end if;
-                  exception
-                     when others =>
-                        null; -- Not a known distro.
-                  end;
+                  if Starts_With (Normalized, "id=") or else
+                    Starts_With (Normalized, "id_like=")
+                  then
+                     Cached_Distro :=
+                       Platforms.Distributions'Value (Tail (Normalized, '='));
+                     Distro_Cached := True;
+                     return Cached_Distro;
+                  end if;
+               exception
+                  when others =>
+                     null; -- Not a known distro.
                end;
             end loop;
 
--- a/src/alire/os_linux/alire-platform.adb
+++ b/src/alire/os_linux/alire-platform.adb
@@ -37,8 +37,7 @@ package body Alire.Platform is
                         Subprocess.Checked_Spawn_And_Capture
                 ("cat", Empty_Vector & "/etc/os-release");
 
-            function Get_Os_Release_Value_For_Key (Key : String;
-                                                   Multiple_Values : Boolean)
+            function Get_Os_Release_Value_For_Key (Key : String)
                                       return Alire.Platforms.Distributions is
 
                use GNAT.Regpat;
@@ -63,20 +62,12 @@ package body Alire.Platform is
                   begin
                      Match (Regexp, Normalized, Matches);
                      if Matches (1) /= No_Match then
-                        if Multiple_Values then
-                           --  Generate Values from space separated items
-                           Values :=
-                             Split (
-                               Normalized
-                                 (Matches (1).First .. Matches (1).Last), ' ');
+                        --  Generate Values from space separated items
+                        Values :=
+                          Split (
+                            Normalized
+                            (Matches (1).First .. Matches (1).Last), ' ');
 
-                        else
-                           --  Generate Values from a single value
-                           Values :=
-                             To_Vector (
-                               Normalized
-                                 (Matches (1).First .. Matches (1).Last));
-                        end if;
                         for Value of Values loop
                            begin
                               return Platforms.Distributions'Value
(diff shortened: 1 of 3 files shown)


src/alire/os_linux/alire-platform.adb
```diff
@@ -41,15 +41,30 @@ package body Alire.Platform is
...
-                     Cached_Distro :=
-                       Platforms.Distributions'Value (Tail (Normalized, '='));
-                     Distro_Cached := True;
-                     return Cached_Distro;
-                  end if;
-               exception
-                  when others =>
-                     exit; -- Not a known distro.
+                  begin
+                     if Starts_With (Normalized, "id=") then
```
mosteo: Better that repeating the whole block, I'd prefer to have 

```suggestion
                     if Starts_With (Normalized, "id=") or else Starts_With (Normalized, "id_like=") then
```
since the rest of the code is common, if my eyes are not missing anything.

mosteo (changes requested):
Hi Christophe, thanks a lot for this contribution, Arch sure is a lively community. Also a bleeding edge one, which may be handy for testing I guess.

In this regard, if you feel motivated, you could take a look at our https://github.com/alire-project/dockerfiles repo and prepare an Arch image that we could use to add Arch to our CI testing. Otherwise I'll come around to it once time permits.

I don't know right now the state of gnat on Arch, depending on it you might use the native compiler or the CE one, as for example does CentOS.

BeChris:
Ok just updated my branch.
I had to fight against style checks made by the compiler.

BeChris:
I will try to put in a place an Arch Docker image in another pull request.

src/alire/os_linux/alire-platform.adb
```diff
@@ -41,15 +41,17 @@ package body Alire.Platform is
                   Normalized : constant String :=
                                  To_Lower_Case (Replace (Line, " ", ""));
                begin
-                  if Starts_With (Normalized, "id=") then
+                  if Starts_With (Normalized, "id=") or else
```
Fabien-Chouteau: in my ubuntu `os-release` I have:
```
ID=ubuntu
ID_LIKE=debian
```

Looks like there is a risk of having a bad detection if `ID_LIKE` is first in the file.

Fabien-Chouteau: Why not use `ID` for Arch Linux?

BeChris: Hello Fabien,
I suggested this modification because my Linux distribution is Manjaro which is based on Arch.
In my /etc/os-release there is:
ID=manjaro
ID_LIKE=arch

I agree with you that there is a risk of bad detection.
In this case, I could revert the detection to rely only on ID and add both Arch and Manjaro as known distributions.
What do you suggest ?

Nota : there is the same problem on a linuxmint distribution:
ID=linuxmint
ID_LIKE=ubuntu

Fabien-Chouteau: Depends if there is a chance that we need to have different package names for `manjaro` and `arch`.

Fabien-Chouteau: Maybe we can look at `ID` first, and if not recognize fallback to `ID_LIKE`.

BeChris:
Just updated my branch to check first id key then, if still an unsupported distribution, falls back to id_like key.

Fabien-Chouteau (changes requested):
Looks good to me except that a debug message when falling back on `id_like` would be better.
```ada
   Trace.Debug ("Unknown distro for key 'id', falling back to 'id_like'");
```

BeChris:
> Looks good to me except that a debug message when falling back on `id_like` would be better.
> 
> ```ada
>    Trace.Debug ("Unknown distro for key 'id', falling back to 'id_like'");
> ```

Debug trace added.

Fabien-Chouteau (approved):

BeChris:
Just saw [here](https://www.freedesktop.org/software/systemd/man/os-release.html) that id_like can contain several values separated by space character.

src/alire/os_linux/alire-platform.adb
```diff
@@ -42,8 +42,15 @@ package body Alire.Platform is
...
+
+               --  Regexp accepting lines not starting with '#' like:
+               --  key=value
+               --  key='value'
+               --  key='value1 value2'
+               --  key="value"
+               --  key="value1 value2"
                Regexp : constant Pattern_Matcher :=
-                 Compile (Key & "\s*=\s*""?([^""]+)""?");
+                 Compile ("^" & Key & "=[""']?([^""']+)[""']?");
```
Fabien-Chouteau, line 52: This means only the first item in the list will be detected?

mosteo:
To add another data point, the Arch docker image in alire-project/dockerfiles#1 contains `ID=` but not `ID_LIKE=`.

src/alire/os_linux/alire-platform.adb
```diff
@@ -34,30 +35,90 @@ package body Alire.Platform is
...
+                             Split (
+                               Normalized
+                                 (Matches (1).First .. Matches (1).Last), ' ');
+
+                        else
+                           --  Generate Values from a single value
+                           Values :=
+                             To_Vector (
+                               Normalized
+                                 (Matches (1).First .. Matches (1).Last));
```
mosteo: Given that, according to https://www.freedesktop.org/software/systemd/man/os-release.html, ID cannot contain spaces, I'd say this `Multiple_Values` isn't necessary, and splitting on spaces can be always attempted, which will simplify this snippet a bit.

src/alire/os_linux/alire-platform.adb
mosteo: There's a loop below over the items in this list, IIUC.

BeChris: This RegExp is used to find the value (being it a single string or several items separated by space characters) in lines like:
- key=value(s)
- key='value(s)'
- key="value(s)"

Depending on Multiple_Values paramater value, single or multiple values are converted into a Vector of strings.
Then a check is made on each Vector value.

I made some informal tests and following examples found in different os-release files are correctly processed:
- ID=ubuntu (identified as Ubuntu)
- ID=debian (identified as Debian)
- ID=manjaro (correctly identified as Unknown distribution)
- ID_LIKE=arch (identified as Arch)
- ID="centos" (correctly identified as Unknown distribution)
- ID="opensuse-tumbleweed" (correctly identified as Unknown distribution)
- ID_LIKE="opensuse suse" (correctly tries opensuse then suse => identified as Unknown distribution)
- ID_LIKE="rhel fedora" (correctly tries rhel then fedora => identified as Unknown distribution)

src/alire/os_linux/alire-platform.adb
BeChris: Ok, to be honest I was a bit afraid to always try to split ID values.
But you are right : I will remove the distinction and simplify the processing.

BeChris:
> To add another data point, the Arch docker image in [alire-project/dockerfiles#1](https://github.com/alire-project/dockerfiles/pull/1) contains `ID=` but not `ID_LIKE=`.

I agree with that but a Manjaro (based on Arch) contains:
`ID=manjaro`
`ID_LIKE=arch`

mosteo:
@BeChris , I think we are good to go here. Do you mind rebasing on top of `master` so the newly added Arch CI test is performed too?

BeChris:
> @BeChris , I think we are good to go here. Do you mind rebasing on top of `master` so the newly added Arch CI test is performed too?

I sometimes encounter problems with rebase command but If I have not been wrong it should be rebased correctly.

mosteo (approved):

mosteo:
Merged, thanks!

PR #601, opened 2020-11-03
Add handling of CentOS, Fedora, Red Hat and their package manager (#20)

### LINKED ISSUE #20: RPM-based distros

BeChris:
Hello,

As every rpm based distribution uses its own wrapper around rpm command, this pull request should fix #20 but only for RedHat based distributions.


src/alire/alire-origins-deployers-system-rh_rpm_wrapper.adb
```diff
@@ -0,0 +1,125 @@
...
+         if Matches (1) /= No_Match then
+            Trace.Debug ("Candidate version string: "
+                         & Line (Matches (1).First .. Matches (1).Last));
+            return
+              Version_Outcomes.New_Result
+                (Semantic_Versioning.Parse
+                   (Line (Matches (1).First .. Matches (1).Last),
+                    Relaxed => True)); --  Relaxed because some Fedora
+                                       --  versions have extra version pts,
+                                       --  e.g. 5.8-3.fc33
```
mosteo: According to semantic versioning, `5.8-3 < 5.8` (dash signals a pre-release). If Fedora instead uses it to provide custom info for true `5.8` versions, it should be parsed as `5.8+3`, for example.

BeChris:
A bit overkill but Rh_To_Semver function should do the trick.

src/alire/alire-origins-deployers-system-rh_rpm_wrapper.adb
BeChris: A bit overkill but Rh_To_Semver function should do the trick.

BeChris: Consult [Fedora Package Versioning Examples](https://fedoraproject.org/wiki/Package_Versioning_Examples) : there are a lot of possibilities !
In Redhat there are even sometimes 4 digits to identify a version !

mosteo: Yup, same happens with debians (https://github.com/alire-project/alire/blob/906095fcb3f018331c46adef6ab2df391acffbcf/src/alire/alire-origins-deployers-system-apt.adb#L44). Fortunately they provide a convenient regexp. But, bottom line, the important thing is that RH does not use "-" in the Semver sense, so we cannot simply parse their versions as-is, so thanks for taking care of that.

BeChris: That's what I did with commit 723da69.

Would you like to review newly added Rh_To_Semver function ?

mosteo: Right, I saw it, I haven't had the time to review it yet :)

mosteo:
@BeChris , do you happen to know what's the status of GNAT packaging on these distros?

BeChris:
> @BeChris , do you happen to know what's the status of GNAT packaging on these distros?

Unfortunately no.

BeChris:
> @BeChris , do you happen to know what's the status of GNAT packaging on these distros?

I did some search and found following results:
### Fedora
Supported and in official repositories (need to install gcc-gnat and gprbuild packages)

### Centos
Supported up to Centos 7 but not found in version 8.
Supported gnat version is somewhat old (found only FSF gcc 4.8 with ada support by installing gcc-gnat package)
No package for gprbuild
It may be better to use GNAT community edition

### Redhat
No packages found at all for gnat nor gprbuild !
No other choices than using GNAT community edition

### OpenSUSE (began a local branch for an eventual pull request)
Found a recent 10.2 FSF gnat package but only on Tumbleweed which a rolling release edition
On more classical Leap edition found only an old gnat package (version 7, last update on 2018-06-05)
It may be better to use GNAT community edition

mosteo:
Thanks a lot for doing the research, Christophe. In summary, only Fedora has a native compiler that we might want to test.

mosteo:
One consequence of this PR is that CentOS would cease to be "unknown", as we are using it currently to test how `alr` behaves in unsupported distributions. 

So we need a replacement for the unsupported distro, something unlikely to become supported soon and with enough support of Python3 to pass our test suite. Any suggestions?

Or perhaps a cleaner solution is to have a specific workflow in which we force the distribution to be undetectable, so we aren't caught in this situation ever again.

Fabien-Chouteau:
> Or perhaps a cleaner solution is to have a specific workflow in which we force the distribution to be undetectable, so we aren't caught in this situation ever again.

We can try a Windows one with msys2 install disable in Alire config.

mosteo:
Hello @BeChris, sorry for the long silence. I finally got around to review the details, and play around a bit on Fedora with `alr`.

In the end, I simplified the version parsing to match what we are doing in other platforms. So, thanks for the rh_to_semver code even if we ended not using it.

Also, from my experiments with `yum -y list`, I see that we have also to exclude epoch numbers, just as in other distros, so I added that too.

mosteo (approved):

PR #621, opened 2020-11-18
Platform detection: Do not raise if /etc/os-release doesn't exist

mosteo:
Simply return distro unknown

### DIFF
--- a/src/alire/os_linux/alire-platform.adb
+++ b/src/alire/os_linux/alire-platform.adb
@@ -21,6 +21,8 @@ package body Alire.Platform is
    -- Distribution --
    ------------------
 
+   OS_Identity_File : constant String := "/etc/os-release";
+
    Cached_Distro : Alire.Platforms.Distributions;
    Distro_Cached : Boolean := False;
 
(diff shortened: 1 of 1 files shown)


src/alire/os_linux/alire-platform.adb
```diff
@@ -30,12 +32,16 @@ package body Alire.Platform is
...
+         Distro_Cached := True;
+         Cached_Distro := Distro_Unknown;
+         return Cached_Distro;
       else
          declare
             use Utils;
             Release : constant Utils.String_Vector :=
                         Subprocess.Checked_Spawn_And_Capture
-                ("cat", Empty_Vector & "/etc/os-release");
+                ("cat", Empty_Vector & OS_Identity_File);
```
Fabien-Chouteau, line 44: We could read the file directly instead of spawning `cat`.

Fabien-Chouteau (approved):

src/alire/os_linux/alire-platform.adb
mosteo: Yes, this is really the result of a low-glucose moment :sweat_smile:

PR #623, opened 2020-11-19
Separate workflow to test unsupported package manager platform

mosteo:
In the end I went for a Linux instead of the suggested Windows in #601 because Windows is at present giving us more problems, with sporadic failures with unidentified causes. Also the test is quicker not requiring the CE setup.

Depends on #621 

This PR adds a new config option, `distribution.disable-detection`, that results in an unknown distribution always being reported.

To break circularity problems, loading the configuration no longer requires to load the manifest, only to locate it. This caused one test to err slightly later (as it does not need the manifest contents), which accounts for the test modification.

### DIFF
--- a/.github/workflows/ci-unsupported.yml
+++ b/.github/workflows/ci-unsupported.yml
@@ -1,0 +1,51 @@
+# The purpose of this CI is to run tests in a platform where there is no
+# detected native package manager
+
+name: CI unsupported
+
+on:
+  pull_request:
+    paths-ignore:
+      - 'doc/**'
+      - '**.md'
+      - '**.rst'
+      - '**.txt'
+
+jobs:
+
+  build:
+    name: CI on unsupported Linux
+
+    runs-on: ubuntu-18.04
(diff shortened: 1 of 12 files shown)


.github/workflows/ci-unsupported.yml
```diff
@@ -0,0 +1,51 @@
...
+      with:
+        distrib: community
+    
+    - name: Install Python 3.x (required for the testsuite)
+      uses: actions/setup-python@v2
+      with:
+        python-version: '3.x'
+
+    - name: Remove identifying file from OS
+      run: sudo rm -f /etc/os-release
```
Fabien-Chouteau: This could have unexpected side effects.
If you want we can add a config key to disable distrib features.

mosteo: I will do this. It may be interesting even for normal users to disable using crates not coming from sources.

Fabien-Chouteau (approved):

PR #629, opened 2020-11-20
Alire.Properties.Licenses: Add support of SPDX expression 

### LINKED ISSUE #191: Use SPDX license identifiers in the index

Fabien-Chouteau:
Fixes #191


src/alire/alire-properties-licenses.adb
```diff
@@ -1,5 +1,84 @@
...
+          when PostgreSQL         => "PostgreSQL",
+          when Unlicense          => "Unlicense",
+          when WTFPL              => "WTFPL",
+          when Zlib               => "Zlib",
+          when GMGPL_2_0          => "GPL-2.0-only WITH GCC-exception-2.0",
+          when GMGPL_3_0          => "GPL-3.0-only WITH GCC-exception-3.1",
+          when Public_Domain | Custom | Unknown =>
+             raise Checked_Error
+               with Errors.Set ("Invalid legacy license conversion for '" &
+                                  Legacy'Img & "'"));
```
mosteo: For my edification: does this mean that SPDX does not support custom licenses? (so we lose this capability?)

Fabien-Chouteau: Yes SPDX doesn't allow license outside of the list they provide. And that's going to be a problem for the existing index.

I think there are several options:
 - `Custom` is not allowed anymore and the `licenses` must only be used for SPDX licenses.
 - Add an option to accept `custom` in the SPDX library.
 - Add a detection of `custom` in `Alire` to bypass the SPDX library.

Fabien-Chouteau: I will try to add a `other-<License-name-here>` support in the SPDX library.

mosteo:
I think the only thing missing here is bumping the `spdx` dependency to `~0.2`, so the self-build succeeds.

Fabien-Chouteau:
@mosteo This is ready for review,

mosteo (approved):

PR #631, opened 2020-11-22
Feature/windows tar

rveenker:
I have made changes to Alire publish for Windows.
The tar command under Windows needs a different approach in order to create a tar file with the current directory as root.
To do this you can change directory to the parent directory using '-C ..' and provide the directory name as the source for tar.
Also the Windows tar does not support '-s' and it uses a dash for the compression command. (i.e. '-czf' instead of 'cfj')

Note that I previously changed the bzip2 (tbz2) into tgz for Windows, but it appears that this change is not necessary.

The Alire test suite uses Msys2, so my first attempt failed. My fix was intended for running Alire on a Windows system without Msys2.

### DIFF
--- a/src/alire/alire-publish.adb
+++ b/src/alire/alire-publish.adb
@@ -329,9 +329,10 @@ package body Alire.Publish is
       Archive    : constant Relative_Path :=
                      Target_Dir
                        / (Milestone
-                          & (if Is_Repo
-                             then ".tgz"
-                             else ".tbz2"));
+                          & (if Is_Repo or GNATCOLL.OS.Constants.OS
+                              in GNATCOLL.OS.Windows
+                            then ".tgz"
+                            else ".tbz2"));
       use Utils.User_Input;
 
       -----------------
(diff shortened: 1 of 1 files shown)


Fabien-Chouteau (changes requested):
Thank you @rveenker , unfortunately the different versions of `tar` that can be found on Windows are difficult to deal with.
As we have learn working on the `Untar` procedure:
https://github.com/alire-project/alire/blob/ff9fb73ca6f7c46af3c1e5f2c0410f816685e219/src/alire/alire-origins-deployers-source_archive.adb#L42
This is one of the reasons why we recommend using `msys2` on Windows.

Here I don't think we should have this complex code to handle all the cases. Let's just stop using `--transform`/`-s` at all, and make the code the same for all versions of `tar`.

rveenker:
Fabien,
I agree with your comment that on Windows you may find different versions of tar, however the one that is now supplied by Microsoft is one that most Windows (10) users will find.
But I don't mind if this change is not adopted into the main project.
Rob

Fabien-Chouteau:
I am not saying that we shouldn't fix the issue, what I am saying is that we can make the same code to work more cases.

Issue #650, opened 2020-12-21
Should unknown enum values in the index be an error?


mosteo:
For our dynamic expressions we use enums with predefined values. Finding an unknown value is currently an error that stops Alire.

As adding values can be a transparent change for current users (e.g., adding a new supported distro), we could consider silently ignoring unknown enum values, unless we are `alr publish`ing. This would allow index updates without bumping the index version.

We could also add an `alr index --check` that reported "errors" that are being ignored for this kind of backwards compatible changes. Certain debug messages could be promoted to warnings with that command.

This happened for example with this submission: https://github.com/alire-project/alire-index/pull/215

Fabien-Chouteau:
I don't think completely silent is a good idea because one could miss a typo for instance.

mosteo:
Yup, that's the main problem.

Having warnings would probably be too polluting, and could scare users that did nothing but update the community index. I'm not sure how to strike the precise balance.

Fabien-Chouteau:
> could scare users that did nothing but update the community
Right...

Let's do the --check then

PR #651, opened 2020-12-21
Bump index version to devel-0.5

mosteo:
The recent addition of more enum values to support other distributions is a breaking change. This commit bumps the index and alr versions.

This raises the question of whether we should by default silently ignore unknown enum values, as this is not a syntactic change. This way, supporting more configurations wouldn't require a new index branch.

Creating issue https://github.com/alire-project/alire/issues/650 to track this idea.


Fabien-Chouteau (approved):

Fabien-Chouteau:
We should take this new index format opportunity to merge the SPDX license change.

PR #669, opened 2021-01-28
Warn about ^0.x dependencies

### LINKED ISSUE #639: Clarify semver caret and tilde operator semantics
We don't make any special interpretation of these operators for `<1.0.0` versions (and semver spec says that there's no guarantee at all for such versions). We can:

- [x] Clarify this interpretation at the relevant point in the catalog spec.
- [x] Emit a warning for `^0.x.x` dependencies, which are likely to be intended as `~0.x.x`.
   - [x] Make this warning optional via some config option.

mosteo:
This is a suspicious use, as most likely `~0.x` is intended.

This PR documents this particular in the catalog, but also warns when such a dependency is added, and at publishing time. The warning can be disabled with the new option `warning.caret`.

Fixes #639.


src/alire/alire-warnings.ads
```diff
@@ -0,0 +1,15 @@
+package Alire.Warnings with Preelaborate is
+
+   --  TODO: thread-unsafe
+
+   procedure Warn_Once (Text : String;
+                        ID   : String := "";
```
Fabien-Chouteau: Maybe the ID could be an enum here.

mosteo:
Circularity horror again. Sigh.

src/alire/alire-warnings.ads
mosteo: The drawback I see with that is that is that it doesn't allow to use the Text instead of the id. For example, if the warning contains some custom info, we may want to repeat it, but not for the same info.

mosteo: A middle ground is by convention define any Ids in that same spec, so they're more visible and documented. I'll try it with the one we have now.

mosteo:
This one would be ready now, if you don't mind the IDs not being an enum.

Fabien-Chouteau (approved):

PR #670, opened 2021-01-28
Separate non-preelaborable parts of Config

mosteo:
These parts are now in Alire.Config.Edit. This allows querying the configuration while remaining preelaborable. New GNAT versions are able to find more preelaboration orders, but older ones still benefit from this separation to break circularities.


mosteo:
This solves the problem in #669 and hopefully will make our life easier in the future.

Fabien-Chouteau (approved):

PR #671, opened 2021-01-28
Rename `alr list` to `alr crates`

### LINKED ISSUE #369: alr list vs alr search --list
There are two ways to list the available packages:

```
alr list
```
and
```
alr search --list
```
Both commands have different outputs.
I would suggest to avoid duplicating ways to list things.
By the way, the output produced by the search is far more useful since you know the package version.

mosteo:
Fixes #369. 

Hopefully this makes more clear that `alr crates` operates at crate level while `alr search` looks into releases.


Fabien-Chouteau:
Why not remove `alr list`/`alr crate` altogether? And put the same functionality in  an `alr search --list-all`.

mosteo:
> Why not remove `alr list`/`alr crate` altogether? And put the same functionality in an `alr search --list-all`.

Given that that would produce an entirely different listing, I find that more confusing. Also `alr search --list` already lists releases. 

What about having `alr search --crates` to signal that we want to look into crates instead of releases (as if a currently non-existent `--releases` were the default). Then the same `--list` that currently looks into releases could be reused.

Fabien-Chouteau:
Ok for me. But I think that the difference between list of crates and list of release will always be a bit confusing.

mosteo:
Updated with the `alr search --crates` option.

Fabien-Chouteau (approved):

Issue #633, opened 2020-11-23
`post-fetch` and `pre-build` actions not running at all expected moments


mosteo:
During packaging of `rclada` it seems that
- `post-fetch` is not run for the root crate.
- `pre-build` is not run for dependencies.

If true (need to verify), this means that `post-fetch` only works for dependencies and `pre-build` only works for root crates.

Update 2021-01-28: this is the intended working per `Alire.Properties.Actions` documentation. It's still counterintuitive and the catalog spec is not clear at all about this.

mosteo:
It's probably worthwhile to homogenize these events so they're run for all releases in the solution at all the described moments. However, these changes would have profound impact on the current build workflow. 

The necessary changes would fit well with what is proposed in #628. I suggest we document the current behavior more clearly and leave the big changes for post-1.0. @Fabien-Chouteau, do you agree?

Fabien-Chouteau:
This will have to be a post 1.0 fix indeed.

onox:
I'm trying to publish over 6 crates for wayland-ada, but I'm stuck at number 2 because alire fails. I've added crate 1 (wayland_ada_scanner) to https://github.com/onox/alire-index and to my local index. Now I'm trying to `alr build -vv -d` crate 2 (wayland_client_ada), which fails:

```
detail: Running action: Pre_Build run: <project>/./mkdir -p src/protocols
debug: Entering folder: .
detail: Spawning: ["mkdir", "-p", "src/protocols"]
detail: Running action: Pre_Build run: <project>/./wayland_ada_scanner /usr/share/wayland/wayland.xml src/protocols
debug: Entering folder: .
detail: Spawning: ["wayland_ada_scanner", "/usr/share/wayland/wayland.xml", "src/protocols"]
debug: Process errored with code 1
stderr: Command ["wayland_ada_scanner", "/usr/share/wayland/wayland.xml", "src/protocols"] exited with code 1
warn: A pre-build action failed, re-run with -vv -d for details
error: Compilation failed.
```

What can I do to make alire find the executable?

mosteo:
I guess you know the relative path for `wayland_ada_scanner` within its crate. Then, it's a matter of adding this path (in `wayland_ada_scanner`), e.g.:

```
[environment]
PATH.append = "${CRATE_ROOT}/bin"
```

Fabien-Chouteau:
Except this probably won't work because `wayland_ada_scanner` is not built before the `post-fetch` action of `wayland_client_ada`.

[...]

Issue #323, opened 2020-02-18
Simplify conditional expression internals


mosteo:
The complex set of generic instances in use over each platform enumeration made sense to make the old Ada index fully type checked at compile time. There is room for significant simplification without affecting functionality or checks that would simplify maintainability.

Low priority for a boring afternoon.

PR #673, opened 2021-01-28
Support for crate configuration

Fabien-Chouteau:
This is the possibility for crates to define configuration variables
that will then generate Ada, C, or GPR files based on the values set by
depending crates.

For embedded projects where compile optimization and static memory usage
are important, it is possible to define sizes of cache buffers or
maximum amount of some resources.

It allows to enable or disable features at compile time, such as debug
logging.

Example:
```toml
[config_variables]
Buffer_Size= {type="Integer", first=0, last=256, default = 20}
Freq = {type="Real", first=0.0, last=440.0, default=220.124}
Device_ID = {type = "String", default = "my device"}
Debug_Level = {type = "Enum", values=["Warn", "Info", "Debug"], default = "Warn"}
Print_Trace = {type = "Boolean", default = true}

[config_settings]
test.Buffer_Size= 100
test.Device_ID = "test config value"
```
`test_config.ads`:
```ada
--  Configuration for test generated by Alire
package test_Config is

   Device_ID : constant String := "test config value";

   Freq_First : constant := 0.0;
   Freq_Last : constant := 440.0;
   Freq : constant := 220.124;

   Print_Trace : constant Boolean := True;

   Buffer_Size_First : constant := 0;
   Buffer_Size_Last : constant := 256;
   Buffer_Size : constant := 100;

   type Debug_Level_Kind is (Warn, Info, Debug);
   Debug_Level : constant Debug_Level_Kind := Warn;

end test_Config;
```

Open questions:
 - Should it be possible to use configuration values in `case`? e.g.:
   ```toml
   [config_variables]
   Enable_Logs = {type = "Boolean"}

   [depends-on.case(Enable_Logs).True]
   logging = "1.0.0"
   ```


mosteo:
Looks great! 

A bit of feedback on your open questions:

> * When should we first generate the configuration output (.ads, .gpr, .h)?
>   
>   * Before `post-fetch`?

Yes, and also before `pre-build`, in case there's been some manual edition.

> * Should it be possible to use configuration values in `case`? e.g.:
>   ```toml
>   [config_variables]
>   Enable_Logs = {type = "Boolean"}
>   
>   [depends-on.case(Enable_Logs).True]
>   logging = "1.0.0"
>   ```

That's an amazing idea. How would conditional dependencies work with GPR files? In tandem with conditional `project-files` expressions?

Could this serve in the future to retrieve variants?: `alr get somecrate --config Ssh_Enabled:=True`

Back to earth, I'm unsure about the implementation difficulties, but my gut feeling is that it is reasonably doable. I would delay it though after #323, which should greatly simplify maintenance of all that part. Also the rework could be done with this idea in mind.

> * Should variable names be case sensitive? (they are right now)

I'd go with Ada syntax here, given the users expectations, hence no.

> * Use a different format for type definition?
>   ```toml
>   [config_variables]
>   var1 = {type = "Real", First = "0.0", Last = "100.0", default = "50.0"}
>   var2 = {type = "Enum", Values = ["A", "B", "C"], default = "B"}
>   ```

My first impulse is to go for verbatim Ada syntax that you can stitch when generating the files, and no hidden assumptions. For example (also putting in a suggestion for the key naming):

```toml
# optional, only for new types:
[config.types] # everything after "Type X is". Probably [config.subtypes] would be needed too.
Log_Levels = "(Debug, Warn, Error)"
Buffer_Sizes = "new Integer range 128 .. 1024"

[config.variables] # everything after the colon in a declaration, default included
Buffer_Last = "constant Buffer_Sizes := 0"
Min_Level = "Log_Levels := Debug" # Constant optional?
Foo = 'constant String := "Bar"'

[config.settings] # override values for something existing in [config.variables]
crate.Min_Level = "Warn"
crate.Buffer_Last = "1024"
```

would generate

```ada
type Log_Levels is (Debug, Warn, Error);
type Buffer_Sizes is new Integer range 128 .. 1024;
Buffer_Last : constant Buffer_Sizes := 0;
Min_Level : Log_Levels := Warn;
Foo : constant String := "Bar";
```

This way you can also have several vars of the same type, and would not arbitrarily limit the kind of declarations you can use. May be a case of too much liberty though (although we could impose restrictions I guess, e.g. only literals for defaults).

Also, IIUC, normally one crate would define the types/defaults, and a dependent would set/override them, right? This means we would need to travel dependencies from leaves to root to properly override all the way up? Or at least detect conflicting settings?

Fabien-Chouteau:
> Looks great!

Thanks Alejandro
 
> > * When should we first generate the configuration output (.ads, .gpr, .h)?
> >   * Before `post-fetch`?
> Yes, and also before `pre-build`, in case there's been some manual edition.

Indeed.

> > * Should it be possible to use configuration values in `case`? e.g.:
> That's an amazing idea. How would conditional dependencies work with GPR files? In tandem with conditional `project-files` expressions?

That a good question. Since gpr file do not have conditional `with` this could be a can of worm...

> Could this serve in the future to retrieve variants?: `alr get somecrate --config Ssh_Enabled:=True`

I never thought about a command line interface for this.
But now that you are mentioning it we could ask users to provide values for variables that don't have default.

> My first impulse is to go for verbatim Ada syntax that you can stitch when generating the files, and no hidden assumptions. For example (also putting in a suggestion for the key naming):

I just pushed a version that relies more on the TOML syntax. My rationale is that it is better to rely on the TOML library rather than implementing a parser just for this. And I do like the result.

> 
> ```toml
> # optional, only for new types:
> [config.types] # everything after "Type X is". Probably [config.subtypes] would be needed too.
> Log_Levels = "(Debug, Warn, Error)"
> Buffer_Sizes = "new Integer range 128 .. 1024"
>
> [config.variables] # everything after the colon in a declaration, default included
> Buffer_Last = "constant Buffer_Sizes := 0"
> Min_Level = "Log_Levels := Debug" # Constant optional?
> Foo = 'constant String := "Bar"'
> 
> [config.settings] # override values for something existing in [config.variables]
> crate.Min_Level = "Warn"
> crate.Buffer_Last = "1024"
> ```

In my opinion this is going too far in complexity.

> Also, IIUC, normally one crate would define the types/defaults, and a dependent would set/override them, right? This means we would need to travel dependencies from leaves to root to properly override all the way up? Or at least detect conflicting settings?

There is already a detection of conflicting settings. But right now I do not check if a crate only sets variable of its own dependencies, i.e. you can set values for crate you don't depend on.

mosteo:
> > Could this serve in the future to retrieve variants?: `alr get somecrate --config Ssh_Enabled:=True`
>
> I never thought about a command line interface for this.
> But now that you are mentioning it we could ask users to provide values for variables that don't have default.

I'm not so much concerned with the particular interface, but interested in this being an opportunity to allow packaging different minor variants without resorting to different crates.

> > My first impulse is to go for verbatim Ada syntax that you can stitch when generating the files, and no hidden assumptions. For example (also putting in a suggestion for the key naming):
> 
> I just pushed a version that relies more on the TOML syntax. My rationale is that it is better to rely on the TOML library rather than implementing a parser just for this. And I do like the result.

OK. Here, I was thinking on how to support easily things like `delta`, `digits`, etc, that seem important (from my inexperienced POV) for embedded.

> In my opinion this is going too far in complexity.

Different personal perceptions, I guess. To me, closer to Ada is always more attractive. I was thinking in the interim to fully go to something like
```toml
config.declare = '''
< full Ada declarations >
'''
```
and just parse the type/var names and defaults from there (!). But no matter, if you like what you have.

> There is already a detection of conflicting settings. 

You're several steps ahead :)

Fabien-Chouteau:
> OK. Here, I was thinking on how to support easily things like `delta`, `digits`, etc, that seem important (from my inexperienced POV) for embedded.

The code only generates named numbers with the idea that users can define their own types from them. This has to be explained the documentation, but so far I have this comment in the code generation part:
```ada
   --  For Real and Int we do not declare a ranged type such as:
   --
   --  type Test_Type is range 0 .. 100;
   --  Test : constant Test_Type := 50;
   --
   --  because this would limit the possibilities of users for things
   --  such as alignment size, etc. Instead we define named numbers
   --  (constant) for the First, Last and actual value of the
   --  variable:

   Test_First : constant := 0;
   Test_Last : constant := 100
   Test : constant := 50;

    --  User can then define their own types:
   type Test_Type is range Config.Test_First .. Config.Test_Last
     with Size => 32;
```
With this you can also define `delta`, `digits`, etc.

> > In my opinion this is going too far in complexity.
> 
> Different personal perceptions, I guess. To me, closer to Ada is always more attractive. I was thinking in the interim to fully go to something like
> 
> ```toml
> config.declare = '''
> < full Ada declarations >
> '''
> ```

I don't see the benefits of this? What would you generate?

mosteo:
OK, now I understand. So it's merely definition of constants and the types have to be explicitly written by the user, relying on these constants. Then I would not bother with assuming these constants are going to be used for ranges, and make the user to explicitly define whatever is needed (that is, give one definition for each of `Foo_Min`, `Foo_Max`... Or whatever the library needs). 

My comments were on the assumption that the types would also be generated by Alire. But your idea is certainly simpler and flexible enough.

Fabien-Chouteau:
> OK, now I understand. So it's merely definition of constants and the types have to be explicitly written by the user, relying on these constants.

I should have started with this.

> Then I would not bother with assuming these constants are going to be used for ranges, and make the user to explicitly define whatever is needed (that is, give one definition for each of `Foo_Min`, `Foo_Max`... Or whatever the library needs).

The range comes from my first experiments in particular with buffer where you don't want negative sizes. So having an upper and lower bounds makes sense, and let the user know as soon as possible that the configuration not valid. I also think that max (and min?) string length would be nice.



> My comments were on the assumption that the types would also be generated by Alire. But your idea is certainly simpler and flexible enough.

Fabien-Chouteau:
> > > * Should it be possible to use configuration values in `case`? e.g.:
> > >   That's an amazing idea. How would conditional dependencies work with GPR files? In tandem with conditional `project-files` expressions?
> 
> That a good question. Since gpr file do not have conditional `with` this could be a can of worm...

Thinking again about this, we could change `auto-gpr-with` feature to generate `with`s in the `XXX_config.gpr` instead of the main `gpr` file. So the main `gpr` only has to with `config/XXX_config.gpr`.

doc/catalog-format-spec.md
```diff
@@ -536,6 +595,171 @@ available.'case(toolchain)'.user = false
...
+provide an Ada API that will result in compile-time optimization or static
+memory usage.
+
+For instance, discriminant are an effective way to let user define the size of
+a buffer:
+
+```ada
+   type Buffered_Thing (Size : Postive) is private;
+private
+   type Buffer_Array is array (Positive range <>) of Unsigned_8;
```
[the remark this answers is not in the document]
Fabien-Chouteau: Thanks bot ^^

Fabien-Chouteau:
@mosteo I will need your help to know where to call `Root.Generate_Configuration;`.

It basically has to be called when any of the crate manifests is changed because that could change the variable definitions or settings.

mosteo:
> @mosteo I will need your help to know where to call `Root.Generate_Configuration;`.
> 
> It basically has to be called when any of the crate manifests is changed because that could change the variable definitions or settings.

For the first generation after retrieval of a crate, around here would be the place: https://github.com/alire-project/alire/blob/23f8d980369968f13fe08f15fbc4094d1607b83e/src/alire/alire-workspace.adb#L168

Problem is, this won't update the root crate after its manifest is edited, as this is a new necessity. (Until now, manifest edition could only affect dependencies). For regenerating the root crate config, the proper place is around here: https://github.com/alire-project/alire/blob/23f8d980369968f13fe08f15fbc4094d1607b83e/src/alire/alire-workspace.adb#L46

Finally, there's the problem of linked crates, which are likely targets for edition. These are currently skipped as they don't need deployment. (Side note: This makes me think that we have a lurking bug around there, as their actions won't ever run.) In any case, to deal with your immediate need, you must check `Rel.Is_Linked`, in which case you can regenerate its config, as the first case in this if: https://github.com/alire-project/alire/blob/23f8d980369968f13fe08f15fbc4094d1607b83e/src/alire/alire-workspace.adb#L77

I was thinking the other day (in the context of #633) that a cleaner solution would be to have something like `Alire.Events` with procedures that get called in the proper order at the proper time for all crates/actions. Since that would stall you for a while, I think you can get by with the previous indications, and I can refactor your insertions to the proper places when doing #633.

Fabien-Chouteau:
> I was thinking the other day (in the context of #633) that a cleaner solution would be to have something like `Alire.Events` with procedures that get called in the proper order at the proper time for all crates/actions. Since that would stall you for a while, I think you can get by with the previous indications, and I can refactor your insertions to the proper places when doing #633.

It would indeed be interesting to have this kind of mechanism.

What about using time-stamps of `toml` files in the dependency tree for detecting changes?

mosteo:
> What about using time-stamps of toml files in the dependency tree for detecting changes?

This is happening already for the root manifest (see e.g. [alire/src/alire/alire-roots.ads:99](https://github.com/alire-project/alire/blob/a689ee3db955cebbb161189ee927475f84ce8b91/src/alire/alire-roots.ads#L99)). The places I commented that you can place your hooks shouldn't even get called if the manifest hasn't changed.

The current assumption is that regular dependencies won't be modified by the user, as they're buried in `alire/cache/dependencies`. In fact, we don't even generate manifest/lockfile for them. This could be changed to allow quick experimentation by modifying the regular dependencies.

In any case, this has to be implemented for linked crates.

doc/catalog-format-spec.md
```diff
@@ -441,6 +441,65 @@ static, i.e. they cannot depend on the context.
...
+
+         - `Integer`: an integer value that can be encoded in 64-bit
+
+         - `Real`: a real value that can be encoded in IEEE 754 binary64
+
+      - `default`: optional default value for the variable. Will be used if no
+        crates explicitly set a value for this variable. Must be a valid value
+        for the type.
+
+      - `first`: (optional) for `Real` and `Interger` types only. Defines the
```
mosteo: Typo in `Interger`

doc/catalog-format-spec.md
```diff
@@ -547,6 +606,171 @@ available.'case(toolchain)'.user = false
...
+## Using crate configuration
+
+`Alire` provides a mechanism for crates to expose a list of variables that can
+be set by other crates depending on them. The configuration variables will then
+be converted to Ada, C and GPR source files that can be used to change the
+behavior or feature set of the code.
+
+Let's start with a simple example. A crate named `test` can print debug log on
+the console. However printing on the console has a performance impact, for
+embedded project it can even have a significant code size impact. Therefore it
```
mosteo, line 618: [an] embedded

doc/catalog-format-spec.md
```diff
@@ -547,6 +606,171 @@ available.'case(toolchain)'.user = false
...
+is be best if this logging can be disabled/enabled at compile time.
```
mosteo: s/is be/would be/

doc/catalog-format-spec.md
```diff
@@ -547,6 +606,171 @@ available.'case(toolchain)'.user = false
...
+It is possible for multiple depending crates to set `test.Enable_Logs` to the
+same value, however if two depending crates set the variable to a different
+value then the configuration is invalid and `Alire` will print an error. If no
+depending crates set the `test.Enable_Logs` variable, then its default value is
+used.
+
+### When to use crate configuration?
+
+Usually when something has to be static or known at compiler-time, either for
+performances or memory usage.
```
mosteo: s/compiler-time/compile time/, s/performances/performance/

doc/catalog-format-spec.md
```diff
@@ -547,6 +606,171 @@ available.'case(toolchain)'.user = false
...
+
+### When _not_ to use crate configuration?
+
+When the Ada languages provides a better alternative. There are many ways to
+provide an Ada API that will result in compile-time optimization or static
+memory usage.
+
+For instance, discriminant are an effective way to let user define the size of
```
mosteo: s/discriminant/discriminants/

mosteo: s/let user/let the user/

doc/catalog-format-spec.md
```diff
@@ -547,6 +606,171 @@ available.'case(toolchain)'.user = false
...
+Enumerations variables in crate configuration can be used to set a level of log
+verbosity:
+```toml
+[config_variables]
+Log_Level = {type = "Enum", values = ["Info", "Debug", "Warn", "Error"], default = "Warn"}
+```
+
+#### Buffer size
+
+Integer variables can be use the define the size of a static buffer:
```
mosteo: s/use/used/

doc/catalog-format-spec.md
```diff
@@ -547,6 +606,171 @@ available.'case(toolchain)'.user = false
...
+file-system:
+```toml
+[config_variables]
+Max_Filename_Length = {type = "Integer", first = 5, last = 128}
+```
+
+#### Select algorithm in GPR project file
+
+Crate configuration also generates a GPR project file, therefore it can be used
+to control which unit are compiled in the project.
```
mosteo: s/unit/units/

Fabien-Chouteau:
@mosteo I think this is ready for review as a first pass.

Things we might want to change is having a top level table called `configuration` instead of having `config_variables` and `config_settings`:
```toml
[configuration.variables]
test = {type="Boolean"}

[configuration.settings]
lol.test=42
```
This will also allow extra fields that are going to be useful in the future:
```toml
[configuration]
output_dir = "config"
disabled=false # Disable the configuration feature for this crate
generate_ada=true
generate_c=false
generate_gpr=true
```
This is the layout I tried when I first started to work on this feature but I couldn't get it to work in the code.

doc/catalog-format-spec.md
```diff
@@ -547,6 +606,171 @@ available.'case(toolchain)'.user = false
...
+### When to use crate configuration?
+
+Usually when something has to be static or known at compiler-time, either for
+performance or memory usage.
+
+### When _not_ to use crate configuration?
+
+When the Ada languages provides a better alternative. There are many ways to
+provide an Ada API that will result in compile time optimization or static
+memory usage.
```
mosteo, line 676: Should we mention `gprbuild -X` flags here? Simple cases can be done by selecting different versions of a spec file through a scenario variable.

src/alire/alire-crate_configuration.adb
```diff
@@ -0,0 +1,401 @@
...
+         --  Normally we would generate a warning, but since that will pollute
+         --  the output making it unusable, for once we write directly to
+         --  stderr (unless quiet is in effect):
+
+         if not Alire_Early_Elaboration.Switch_Q then
+            GNAT.IO.Put_Line
+              (GNAT.IO.Standard_Error,
+               TTY.Warn ("warn:") & " Generating incomplete environment"
+               & " because of missing dependencies");
+         end if;
```
mosteo, line 48: Does this output to stderr make sense here and in `80:92`, or would a `Trace.Warning` be enough? I don't remember any parseable output being generated for this feature?

mosteo (commented):
I have had a look through the code to get a general idea of the changes, not much to say as I don't get all the details :). I will play a bit with the feature to see if I have any other suggestions, and will have a go at the `configuration` top-level table, but to me it looks okay.

If you prefer, we can merge as is and I can make my changes on a later PR.

doc/catalog-format-spec.md
Fabien-Chouteau: We will see how this evolves but crate config should replace scenario variables in most cases.

src/alire/alire-crate_configuration.adb
Fabien-Chouteau: I think this is copy/paste from legacy code.

Fabien-Chouteau: Actually there is a comment above:
```
         --  Normally we would generate a warning, but since that will pollute
         --  the output making it unusable, for once we write directly to
         --  stderr (unless quiet is in effect):
```
I think this is coming from you. I don't know if it is still applicable.

doc/catalog-format-spec.md
mosteo: OK.

src/alire/alire-crate_configuration.adb
mosteo: Yes, I remember writing that. That's because when generating the output of `printenv`, our logs should not intermingle with the intended output. If this has no relation, it can be replaced with a mere `Trace.Warning`.

mosteo: Fixed in #699.

mosteo (approved):

PR #674, opened 2021-01-29
Fix problem with temporaries in Windows

### LINKED ISSUE #522: Can't delete temporary folder that contains a git repo [Windows only]
This happens in the context of cloning a remote during publishing checks.

Both `Ada.Directories` and `GNATCOLL.VFS` fail. Python reports:

```
PermissionError: [WinError 5] Access is denied: alr-fwrt.tmp\\.git\\objects\\1d\\696bed4ef917b1adbdef18723016987ed62e41'
```

Doubtful this is about permissions, since we run tests as root. It might be that the file is still open, but I don't currently understand how that can be happening.

mosteo:
Turns out that git marks pack files as read-only, and `Ada.Directories.Delete_Tree` won't delete those on Windows.

I've resorted to forcing any temporary folder as writable on Windows.

Fixes #522.


src/alire/alire-directories.adb
```diff
@@ -349,6 +350,22 @@ package body Alire.Directories is
          return;
       end if;
 
+      --  Force writability of folder when in Windows, as some tools (e.g. git)
+      --  that create read-only files will cause a Use_Error
+
+      if Kind (This.Filename) = Directory and then
+        GNAT.OS_Lib.Directory_Separator = '\'
```
Fabien-Chouteau: Is this to detect Windows? I think we have a better way:
```ada
GNATCOLL.OS.Constants.OS = GNATCOLL.OS.Windows
```

Fabien-Chouteau (changes requested):

src/alire/alire-directories.adb
mosteo: Right you are

Fabien-Chouteau (approved):

PR #675, opened 2021-01-29
Automatically narrow down dependencies given as "any"

mosteo:
I think `npm` also works like this. When a user adds a dependency without giving a version range, after this PR we use the solved version to narrow down the dependency to be added to what the user probably intends. E.g.: `alr with foo` previously resulted in `foo*` being added; now it will result in either `foo~0.x` or `foo^x.y`, etc being presented as the requested dependency.

A `*` dependency can still be added by requesting it explicitly with `alr with 'foo>=0'` (or manually editing, of course).

This may merit a config option to disable this behavior.


onox:
> I think `npm` also works like this.

`npm` (and `yarn`) will look up the `latest` tag of the crate and then return the resolved version (seems to be x.y.z) with the `^` prefix. So they implicitly use `foo@latest` when you write `foo`.

Yarn will use `~` prefix if `--tilde` is used and if `--peer` is used, the default range is `*` and no resolving is done. (A peer dependency is used for crates that are plug-ins which are only known to work for specific x.y version of their host crate)

mosteo:
Thanks for the precision, @onox. I think the end result is similar (at least for regular crates... The idea of peer/host crates is new to me).

Fabien-Chouteau (approved):

PR #676, opened 2021-02-02
Checking if folder is here

Fabien-Chouteau:


Fabien-Chouteau:
@mosteo Looks like msys2 is not installed in the correct location.

Fabien-Chouteau:
Well.. https://github.com/msys2/msys2-installer/pull/20

mosteo:
Well spotted... Urgh, I had just found the new install location via https://github.com/mosteo/alire/pull/11 but couldn't figure how that could be happening. Thanks!

Fabien-Chouteau:
Unfortunately that means all `0.7.x` versions of `alr` are broken.

mosteo:
After the pending merges I think we are ready to branch out `releases/1.x`

PR #677, opened 2021-02-02
alr-platforms-windows.adb: msys2 install command line changed

Fabien-Chouteau:
https://github.com/msys2/msys2-installer/pull/20 changed the name of the argument for installation prefix of the msys2 installer.
All previous versions are impacted and won't be able to install msys2 at the correct location.


mosteo (approved):

Fabien-Chouteau:
We should use a github URL to a version of the script that is known to work, not just the latest on `master`.

mosteo:
For reference: done in #678.

PR #687, opened 2021-02-11
Roots: retrieve GPR files from solution releases

mosteo:
@Fabien-Chouteau let me know if this is what you need. (It's untested so it may need tweaking.)

### DIFF
--- a/src/alire/alire-releases.ads
+++ b/src/alire/alire-releases.ads
@@ -208,7 +208,9 @@ package Alire.Releases is
                            P         : Alire.Properties.Vector;
                            With_Path : Boolean)
                            return Utils.String_Vector;
-   --  with relative path on demand
+   --  With relative path on demand. With path refers to the relative
+   --  paths specified in the manifest, not to the actual path on disk of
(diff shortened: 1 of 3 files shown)


Fabien-Chouteau:
Thank you @mosteo, I don't think this is what I need.
I want to create a gpr config file for each release of a solution. And for each of the files I need the project files of the direct dependencies of that release.

The calling code will look something like that: 
```ada
     for Rel of Solution.Releases.Including (Root.Release) loop
        declare
           List_Of_Withs : ........ := Solution.Get_Direct_Withs (Rel);
        begin
           This.Generate_GPR_Config  (Rel.Name, "config.gpr", List_Of_Withs);
        end;
      end loop;
```

`Solution.Get_Direct_Withs` would return the list of direct "withs" for the given release.

mosteo:
OK, now I understand

mosteo:
Fabien, check the new `Alire.Roots.Direct_Withs` function.

Fabien-Chouteau:
That works, thank @mosteo

PR #690, opened 2021-02-23
Cache lockfiles, remove Root global, and related refactorings

### LINKED ISSUE #618: Alire.Roots.Solution needs caching
As every call reloads the `alire.lock`-stored solution, load times start to become noticeable after adding a few dependencies, for operations that use this call several times.

Caching must take into account that the lockfile is updated when dependencies are modified, so the timestamp has to be checked before each use.

mosteo:
Fixes #618 and parts of #160.

This started as a focused change to cache the lockfile reads, but to fully take advantage of it I got carried away and ended removing the many implicit usages of roots. Main changes are:

- The lockfile is only read whenever it has actually changed (detected by size/timestamp).
- `Alr.Commands.Command` stores an `Alire.Roots.Optional.Root` that conveniently can be reused and be passed along. This is the most interesting part of the PR, as it removes many implicit/hidden interactions that relied on the global `Root.Current`. It's also a significant improvement towards eventual separation of the `libalire`/`alr` parts.
- `Alire.Workspace` disappears (refactored into `Alire.Roots` mostly) as these subprograms only made sense in the context of roots, actually.

This won't pass the `alr build` check as there are new commits in crate `aaa` that don't have a release yet.


mosteo:
I have disabled the failing self-build check for now, for two reasons: it will make more sense when we are nearing a new stable release, and once we have pinning to remotes it will be less painful to have up-to-date dependencies in the manifest.

Fabien-Chouteau (approved):

PR #699, opened 2021-03-02
Nest configuration tables under [configuration]

mosteo:
Goes on top of #673. Indeed it wasn't trivial to do when having separate properties for the nested tables; that was something I didn't realize at the time: the loading machinery expects each top-level entry to be its own property. @Fabien-Chouteau, to minimize changes I have internally kept your two properties, which are loaded indirectly via the top-level new `configuration` property. I don't think this will have any further impact.

I'm now a bit wary about using `[configuration.settings]` when we really mean `[configuration.values]` or `[configuration.assignments]`, specially if in the future actual Alire configuration settings will go under the `[configuration]` table. Not sure if it's worth changing though.

~This PR needs a small change in `TOML`, so it will fail the `alr build` check.~ It builds, but the result wouldn't past the tests.


src/alire/alire-releases.adb
```diff
@@ -884,8 +887,10 @@ package body Alire.Releases is
         "short_description: " & Utils.YAML.YAML_Stringify (R.Description) &
         ASCII.LF &
         "dependencies: " & R.Dependencies.To_YAML & ASCII.LF &
-        "config_variables: " & Props_To_YAML (R.Config_Variables) & ASCII.LF &
-        "config_settings: " & Props_To_YAML (R.Config_Settings) & ASCII.LF;
+        "configuration.variables: " &
```
Fabien-Chouteau: I am not sure this will work for YAML. But I can change it later if needed.

Fabien-Chouteau (approved):

Fabien-Chouteau:
Thank you @mosteo !

> I'm now a bit wary about using `[configuration.settings]` when we really mean `[configuration.values]` or `[configuration.assignments]`, specially if in the future actual Alire configuration settings will go under the `[configuration]` table. Not sure if it's worth changing though.

I am perfectly fine changing the word `settings`.

mosteo:
> I am perfectly fine changing the word `settings`.

In that case I think I prefer `configuration.values`. I will update the documentation too, which I forgot for the new structure.

src/alire/alire-releases.adb
mosteo: Ah, because those are YAML dictionary keys... Got it.

mosteo:
@Fabien-Chouteau , I think this one is ready so we could merge both this and #673.

Another thing that I observed while testing: do we want the config to be generated for all crates, even those without config variables? On the one hand I see how one could want to be sure that `crate_config` always exists; on the other it's kind of an unnecessary pollution (e.g., in the root crate it will always pop up in control version even if you have no configuration in your crate).

Fabien-Chouteau:
> Another thing that I observed while testing: do we want the config to be generated for all crates, even those without config variables? On the one hand I see how one could want to be sure that `crate_config` always exists; on the other it's kind of an unnecessary pollution (e.g., in the root crate it will always pop up in control version even if you have no configuration in your crate).

I plan to have an opt-out field in `[configuration]`, but the default should be on in my opinion. One reason is that the auto-gpr-with will be in the generated gpr config file instead of the root one.

mosteo:
All right, makes sense to me too. Then I'm going to merge these two.

PR #706, opened 2021-03-10
Error scopes

mosteo:
This is a generalization of the idea behind `TOML_Adapters.Key_Queue.Descend` (nested information to display on error). The new type `Errors.Scope` allows to push a new message into the error stack; whenever an error is stored with `Errors.Set`, the complete error stack is stored for the later report. If a scope is left, the corresponding info is removed from the stack.

This will allow to provide more precise error information also outside of the context of TOML parsing. `TOML_Adapters.Key_Queue` has been reworked to rely on the error stack too. With this unified error mechanism, an exception does not result in the complete loss of the TOML information stack.

### DIFF
--- a/src/alire/alire-conditional_trees-toml_load.adb
+++ b/src/alire/alire-conditional_trees-toml_load.adb
@@ -1,4 +1,5 @@
 with Alire.Conditional_Trees.Case_Nodes;
+with Alire.Errors;
 with Alire.TOML_Keys;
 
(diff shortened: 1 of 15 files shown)


Fabien-Chouteau (approved):
I am surprised that non of the tests requires modification after this change.

mosteo:
Right... I hope that's a good sign.

PR #711, opened 2021-03-22
Store relative pins as such

### LINKED ISSUE #662: `alr with` stores value of --use as absolute path in alire.lock
I have a monorepo with a number of crates, including wayland_ada_scanner (a tool) and wayland_client_ada (a library).

During development or CI I would like to use the local wayland_ada_scanner crate in wayland_client_ada. I added the scanner as a dependency to wayland_client_ada as follows:

```
alr with wayland_ada_scanner^0.1.0 --use=../wayland_ada_scanner
```

This results in the following alire.lock file:

```
[solution]
[solution.context]
solved = true
[[solution.state]]
crate = "wayland_ada_scanner"
fulfilment = "linked"
pinned = false
transitivity = "direct"
versions = "^0.1.0"
[solution.state.link]
kind = "softlink"
path = "file:/absolute/path/to/wayland-ada/wayland_ada_scanner"
```

This obviously won't work for other users who clone the git repo. If the `--use` option is given a relative path, can this relative path be stored in the lock file as well? Perhaps the path can first be made absolute and then transformed to relative again to deal with symlinks and ../ folders.

mosteo:
Fixes #662

### DIFF
--- a/src/alire/alire-externals-softlinks.adb
+++ b/src/alire/alire-externals-softlinks.adb
@@ -3,8 +3,11 @@ with Ada.Directories;
 with Alire.URI;
 with Alire.Utils.TTY;
 
+with GNATCOLL.VFS;
+
 package body Alire.Externals.Softlinks is
 
+   package Adirs renames Ada.Directories;
    use TOML;
 
    package Keys is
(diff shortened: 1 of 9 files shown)


mosteo:
Hi @onox, could you at your convenience check if this patch improves you situation? Thanks!

Fabien-Chouteau (approved):

PR #713, opened 2021-03-22
Transitively add pinned dirs

mosteo:
When a pin to a local crate is added, recursively add its pins to the solution too.

Goes on top of #711.

@onox, if you haven't yet tested #711, you can test this one directly for the use case we talked about on gitter.

Testcase incoming.

### DIFF
--- a/src/alire/alire-solutions.adb
+++ b/src/alire/alire-solutions.adb
@@ -6,6 +6,7 @@ with Alire.Dependencies.Containers;
 with Alire.Dependencies.Diffs;
 with Alire.Dependencies.Graphs;
 with Alire.Index;
+with Alire.Roots.Optional;
 with Alire.Root;
 with Alire.Solutions.Diffs;
 with Alire.Utils.Tables;
@@ -286,6 +287,36 @@ package body Alire.Solutions is
             --  one missing a crate in the other solution is worse.
       end case;
    end Is_Better;
+
+   -------------
+   -- Linking --
+   -------------
+
+   function Linking (This  : Solution;
+                     Crate : Crate_Name;
+                     Link  : Externals.Softlinks.External)
+                     return Solution
+   is
+      Linked_Root : constant Roots.Optional.Root :=
+                      Roots.Optional.Detect_Root (Link.Path);
+   begin
+      return Result : Solution := (Solved       => True,
+                                   Dependencies =>
+                                     This.Dependencies.Including
+                                       (This.State (Crate).Linking (Link)))
+      do
+         if Linked_Root.Is_Valid and then Linked_Root.Value.Has_Lockfile then
+            for Dep of Linked_Root.Value.Solution.Links loop
+               Result :=
+                 Result
+                   .Depending_On (Dep)
+                   .Linking
+                     (Dep.Crate,
+                      Linked_Root.Value.Solution.State (Dep.Crate).Link.Path);
+            end loop;
+         end if;
+      end return;
+   end Linking;
 
    ------------------
    -- New_Solution --
--- a/src/alire/alire-solutions.adb
+++ b/src/alire/alire-solutions.adb
@@ -6,6 +6,7 @@ with Alire.Dependencies.Containers;
 with Alire.Dependencies.Diffs;
 with Alire.Dependencies.Graphs;
 with Alire.Index;
+with Alire.OS_Lib;
 with Alire.Roots.Optional;
 with Alire.Root;
 with Alire.Solutions.Diffs;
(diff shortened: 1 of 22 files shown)


onox:
If I add orka in orka_glfw then `alr` detects orka's dependencies, but I get a failed precondition after proceeding:

```
detail: Narrowing down dependency orka* as orka~0.2.0
detail: Using default value for configuration 'warning.caret': 'TRUE'
detail: Using default value for configuration 'warning.caret': 'TRUE'
info: Requested changes:
info: 
info:    ✓ orka ~0.2.0 (add)
info: 
info: Changes to dependency solution:
info: 
info:    =  libglfw3        3.2.1 (unchanged)                 
info:    +⊙ orka            0.2.0 (new,pin=../orka)           
info:    +⊙ orka_egl        1.0.0 (new,pin=../orka_egl)       
info:    +⊙ orka_simd       1.0.0 (new,pin=../orka_simd)      
info:    +⊙ orka_transforms 1.0.0 (new,pin=../orka_transforms)
info:    +⊙ orka_types      1.0.0 (new,pin=../orka_types)     
info: 
Do you want to proceed?
[Y] Yes  [N] No  (default is Yes) yes
detail: Manifest updated, fetching dependencies now
error: failed precondition from alire-solutions.ads:313
```

onox:
I guess there is at the moment no way to regenerate a lock file with a command (because the paths are stored in the lock file)? I had to remove the dependencies from the toml file and then re-add them to recreate the lock file.

mosteo:
Thanks for the report, @onox. I'll try to reproduce.

> I guess there is at the moment no way to regenerate a lock file with a command (because the paths are stored in the lock file)? I had to remove the dependencies from the toml file and then re-add them to recreate the lock file.

`alr update` should rebuild the lockfile, taking into account pre-existing pins. In this case with the exception, I'm not sure yet what would happen.

mosteo:
@onox, I could reproduce your problem using your repository. That's a lot of crates you have there :D. It should work after my last push.

onox:
Thanks, the failed pre-condition is now gone.

I notice that `alr` adds the relative paths twice (once as a relative path and once as an absolute path):

```
detail: Env: crates appends '../orka' to 'GPR_PROJECT_PATH'
detail: Env: crates appends '/absolute/path/to/orka' to 'GPR_PROJECT_PATH'
```

Fabien-Chouteau:
@onox it looks to me like you are making you life harder having so many crates.
What is the reason to have `orka-types`, `orka-smid` and `orka-transform` outside of `orka`?

onox:
Because `orka-simd` and `orka-transforms` provide bindings for SSE/AVX and vector/matrix operations. I also use these for a numerical integrator for physics. These two small crates are also perfectly reusable by other projects that do not need or want to depend on the OpenGL and rendering stuff.

Anyway, I've fixed the build problems I was having.

src/alire/alire-solutions.adb
```diff
@@ -287,6 +289,53 @@ package body Alire.Solutions is
...
+      --  Recursively find any other links
+
+      return Result : Solution := (Solved       => True,
+                                   Dependencies =>
+                                     This.Dependencies.Including
+                                       (This.State (Crate).Linking (Link)))
+      do
+         if Linked_Root.Is_Valid and then Linked_Root.Value.Has_Lockfile then
+            for Dep of Linked_Root.Value.Solution.Links loop
+               Result := Result
```
Fabien-Chouteau: This statement is not very readable in my opinion.

Maybe the result of `Externals.Softlinks.New_Softlink` could be in constant.
same for `Linked_Root.Value.Solution.State`.

src/alire/alire-utils-user_input.adb
```diff
@@ -25,6 +27,28 @@ package body Alire.Utils.User_Input is
        when No     => "no",
        when Always => "always");
 
+   ------------------
+   -- Approve_Path --
```
Fabien-Chouteau: Discrepancy with actual function name.

Fabien-Chouteau (approved):

src/alire/alire-utils-user_input.adb
mosteo: Thanks, fixed.

src/alire/alire-solutions.adb
mosteo: I have refactored things a bit for clarity.

mosteo:
> I notice that `alr` adds the relative paths twice (once as a relative path and once as an absolute path):

Thanks for catching that.

> Anyway, I've fixed the build problems I was having.

So these weren't caused by alr?

Fabien-Chouteau (approved):

onox:
> So these weren't caused by alr?

Correct, not caused by alr.

PR #716, opened 2021-03-26
Complete/rename the set of feedback procedures with icon prefix

mosteo:
To clarify things a bit, `Log_*` are the general logging procedures and `Put_*` are intended to report events to the user at the regular Info level, but prefixed with an icon: success (✓), failure (✗), warning (⚠), info (ⓘ).

### DIFF
--- a/src/alire/alire.ads
+++ b/src/alire/alire.ads
@@ -248,11 +248,21 @@ package Alire with Preelaborate is
    procedure Log_Exception (E     : Ada.Exceptions.Exception_Occurrence;
                             Level : Simple_Logging.Levels := Debug);
 
-   procedure Log_Info (Text : String; Level : Trace.Levels := Info);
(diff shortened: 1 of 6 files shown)


src/alire/alire.ads
```diff
@@ -9,7 +9,7 @@ with Simple_Logging;
 
 package Alire with Preelaborate is
 
-   Version : constant String := "1.1.0-dev";
+   Version : constant String := "1.1.0-dev+c122dc5d";
```
Fabien-Chouteau, line 12: Why change the version here?

Fabien-Chouteau (approved):

src/alire/alire.ads
mosteo: This is to have an easy unique id in case someone reports problems using a build from master; otherwise all share the same version.

PR #723, opened 2021-04-14
Binary crates

mosteo:
This is a bare-bones implementation of what we discussed in our last meeting. In essence, there is a new origin (`Binary_Archive`) that contains the same information as our current `Source_Archive`, but that accepts conditional expressions. These are deployed in the configuration cache instead of the workspace cache, so they would be shared by workspaces using the same configuration.

I have a couple of example crates for the repo you @Fabien-Chouteau sent me here: https://github.com/mosteo/alire-index/tree/feat/binaries (see the [gnat](https://github.com/mosteo/alire-index/blob/feat/binaries/index/gn/gnat/gnat-10.2.0.toml) and [gprbuild](https://github.com/mosteo/alire-index/blob/feat/binaries/index/gp/gprbuild/gprbuild-21.0.0.toml) crates).

Then, it's a matter of creating a crate with a dependency on those. I tried gprbuild only (as gnat takes its time to download) and it's properly invoked, but it gives this error:

```
$ alr build
gprconfig: can't find a native toolchain for language 'ada'
gprconfig: can't find a native toolchain for language 'ada'
xxx.gpr:4:09: no compiler for language "Ada", cannot compile "xxx.adb"
```

This happens although I have the system Ada compiler in the path, without the system gprbuild installed.


Fabien-Chouteau:
Thanks @mosteo that's great :)

> Then, it's a matter of creating a crate with a dependency on those. I tried gprbuild only (as gnat takes its time to download) and it's properly invoked, but it gives this error:

This is an issue with gprconfig compiler detection. I guess you are on Debian/Ubuntu, they have patches for that: https://sources.debian.org/patches/gprconfig-kb/21.0.0-2/gprconfig.patch/

The problem is that I don't know how compatible this is across distribs.

mosteo:
I closed this one by mistake while doing branch cleanup. I'll reopen once the pins ones are merged.

Issue #729, opened 2021-05-03
v1.0.0 build failure with FSF GCC 11.1.0


simonjwright:
The build generates a lot of warnings! But also some errors.

This in `alire-conditional_trees.ads`,
```
   525.    function Root (This : Tree) return Node'Class is
   526.      (This.Constant_Reference);
              |
        >>> actual for explicitly aliased formal is too short lived
```
This could well be a compiler issue, of course.

Likewise in `alire-dependencies-states.adb`,
```
     8.    overriding function "=" (L, R : Stored_Release) return Boolean
     9.    is
    10.       use type Milestones.Milestone;
    11.       use type Origins.Origin;
    12.    begin
    13.       return
    14.         (L.Is_Empty and then R.Is_Empty)
    15.         or else
    16.           (not L.Is_Empty and then not R.Is_Empty
    17.            and then
    18.            L.Constant_Reference.Milestone = R.Constant_Reference.Milestone
                   1                                2
        >>> actual for explicitly aliased formal is too short lived
        >>> actual for explicitly aliased formal is too short lived

    19.            and then
    20.            L.Constant_Reference.Origin = R.Constant_Reference.Origin);
                   1                             2
        >>> actual for explicitly aliased formal is too short lived
        >>> actual for explicitly aliased formal is too short lived

    21.    end "=";
```

Also, more definitely a problem,
```
alire-manifest.adb:47:07: actual for "This" must be a variable
alire-manifest.adb:258:26: actual for "This" must be a variable
alire-manifest.adb:269:07: actual for "This" must be a variable
```
the first of which is
```
    47.       Replacer.Replace; -- All went well, keep the changes
              |
        >>> actual for "This" must be a variable
```
(`Replacer` is declared `constant`)

mosteo:
Thanks for the early warning!

simonjwright:
I’m now pretty sure the `too short lived` problem is a compiler bug. Not so sure about the `Replacer` one, though.

mosteo:
To be frank, I'm always a bit uneasy when messing with references and their pointers. But in that particular case it seems to be returning a copy of the pointee, right?

simonjwright:
On [c.l.a](https://groups.google.com/g/comp.lang.ada/c/7dw7Oqi8lIk/m/t6UJkTIHDQAJ), Randy said

> We spent a lot of time and effort in the ARG talking about this case (see 
> AI12-0402-1). The Ada 2012 RM does indeed say this case is illegal. The 
[...]

PR #731, opened 2021-05-08
Workround for issue #729

simonjwright:
This came up on compiling alire (v1.0.0) on macOS Big Sur with FSF GCC 11.1.0. [A conversation on c.l.a](https://groups.google.com/g/comp.lang.ada/c/7dw7Oqi8lIk/m/5NUBhFN8DQAJ) indicates that the same issue will likely come up with GNAT CE 2021.

Updated rules in 3.10.2 and 6.4.1(6) caused the use of
Constant_Reference, which takes the Container as an aliased parameter,
to trip over its use on a container passed as an un-aliased parameter.

### DIFF
--- a/src/alire/alire-conditional_trees.adb
+++ b/src/alire/alire-conditional_trees.adb
@@ -22,7 +22,7 @@ package body Alire.Conditional_Trees is
    function To_YAML (This : Tree) return String
    is (if This.Is_Empty
        then ""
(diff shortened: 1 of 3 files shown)


Fabien-Chouteau (approved):

mosteo (approved):

mosteo:
Merged, thanks, Simon.

PR #746, opened 2021-06-10
Fix new warnings in CE2021

mosteo:


Fabien-Chouteau:
You also change the CI runs by using the new version of the toolchain GitHub Action: `uses: ada-actions/toolchain@ce2021`

mosteo:
> You also change the CI runs by using the new version of the toolchain GitHub Action: `uses: ada-actions/toolchain@ce2021`

Is it normal that it is failing for macOS?

Fabien-Chouteau:
> Is it normal that it is failing for macOS?

Yes, I forgot that there is not ce2021 for macOS so you have to use ce2020 there.

Fabien-Chouteau (approved):

PR #751, opened 2021-06-18
Fixes for recursive link pins

### LINKED ISSUE #736: Automatic update is needed after pinning to a remote
Currently, we solve and then fetch pins. However, these pins may introduce new dependencies, which aren't known until they're downloaded, which in turn require (currently) a new manual update. We must automate the recursive discovery of all needed dependencies.

mosteo:
Fixes to properly load and use pins that involve indirect link pins (i.e. pinning to a crate that itself has pins).

Subtasks in this PR:
- [x] Tests for this functionality
- [x] Remote pins recursive deployment (Fixes #736)

@onox, if you want to test this one with your crates, that would be great. In my own testing I'm hitting some build errors but I don't think these are coming from `alr`.


onox:
@mosteo Alire is converting the relative paths to absolute if I run `alr update`.

onox:
I can build and run AWT again :+1: 

>  In my own testing I'm hitting some build errors but I don't think these are coming from `alr`.

Are you referring to:

```
awt-linux.gpr:8:06: unknown project file: "evdev"
awt-linux.gpr:9:06: unknown project file: "inotify"
```

when you run `make tools` in orka repo? I get those too. If I run `alr update` followed by `alr build` in `orka_tools` crate then I can build that crate. But it stores absolute paths in the lock file.

mosteo:
> But it stores absolute paths in the lock file.

I'm working on that, but that should not be a problem anyway as pin info now comes from the manifest, where it will always be whatever you put in there. The lockfile can be considered internal info. And `alr pin` should show you proper relative paths.

> Are you referring to:

> awt-linux.gpr:8:06: unknown project file: "evdev"
> awt-linux.gpr:9:06: unknown project file: "inotify"

Actually no, it was something about missing source folders in a project file. This is more troublesome as there should be no missing project files.

Just to be sure, as you may be mixing old lockfiles with new code, either touch your manifest or delete lockfiles or run `alr update` when you try a new build.

onox:
> The lockfile can be considered internal info.

Does that mean projects would have to delete their `alire.lock` from their git repositories?

> Just to be sure, as you may be mixing old lockfiles with new code, either touch your manifest or delete lockfiles or run `alr update` when you try a new build.

The error has disappeared after running `alr update`.

mosteo:
> Does that mean projects would have to delete their `alire.lock` from their git repositories?

It may well be the case. Once user pins are out of there, all information in the lockfile is rebuilt from the manifest and it serves just as a cache for the current solution. But I still have to think through all the ramifications.

onox:
I thought one of the purposes of the lock files was that an `alire.toml` can specify version "^a.b.0" for a dependency, but the `alire.lock` then specifies the actual used version, which could be "a.b.1" for example.

onox:
Lock files are OK now :+1:

mosteo:
> I thought one of the purposes of the lock files was that an `alire.toml` can specify version "^a.b.0" for a dependency, but the `alire.lock` then specifies the actual used version, which could be "a.b.1" for example.

This is it, in essence. When there is no lockfile, a solution with concrete versions (`a.b.1`) has to be produced from the version sets (`^a.b.0`) in the manifest. This is potentially time consuming, so once found the solution is stored in the lockfile. (We currently store more info than strictly necessary in the lockfile, but it boils down to knowing the exact release to use for each dependency.) In the beginning Alire hadn't lockfiles and a solution was recomputed every time.

So removing the lockfile from version control only entails a fresh solving when a crate is used for the first time.

Fabien-Chouteau:
> The lockfile can be considered internal info.

Should we move it to the `alire/` folder then?

mosteo:
> > The lockfile can be considered internal info.
> 
> Should we move it to the `alire/` folder then?

Yes, I wanted to play with this idea after the binary crates.

mosteo:
This one ended being somewhat of a refactor to make things cleanly. Also it lays a better foundation for the incoming branch pins.

Fabien-Chouteau (approved):

PR #743, opened 2021-05-28
Pins in the manifest

mosteo:
Pins are now to be defined in the manifest:

```
[[pins]]
foo = { version = "1.3.2+bugfix" }
bar = { path = "../my/bar" }
baz = { url = "https://github.com/baz.git" } # No commit, will use current HEAD, will update on `alr update`
gru = { url = "https://gitlab.com/gru.git" commit="123456890abcdef..." } # Explicit commit, won't update
```

This PR is still incomplete, but is testable: the new pins work, and the old pins also work, but such old pins will be forgotten after any update (explicitly requested, or by editing the manifest), as the update process syncs with the pins in the manifest.

To do:
- [x] Remove old pins
- [x] Tests specific to the new format

Fixes #721

PR #754, opened 2021-06-21
Pins to remote branches

mosteo:
An enhancement on top of #743 and #751. Now you can do

```
[[pins]]
foo = { url="http://blah", branch="bar" }
```


src/alire/alire-user_pins.adb
```diff
@@ -114,9 +133,10 @@ package body Alire.User_Pins is
             return;
          end if;
 
-         --  Finally update
+         --  Finally update. In case the branch has just been changed by the
+         --  user in the manifest, the following call wil also take care of it.
 
-         if not VCSs.Git.Handler.Update (Destination).Success then
+         if not VCSs.Git.Handler.Update (Destination, Branch).Success then
```
Fabien-Chouteau, line 139: Will this do a recursive update? (submodules)

testsuite/drivers/helpers.py
```diff
@@ -66,6 +68,14 @@ def dir_separator():
     return '/' if os.name != 'nt' else '\\'
 
 
+def touch(file):
```
Fabien-Chouteau: E3 has a touch function: `from e3.os.fs import touch`, it should be available.

Fabien-Chouteau (changes requested):

src/alire/alire-user_pins.adb
mosteo: Yes, it's at https://github.com/alire-project/alire/blob/01264f9c9e1785ed19e04d41e9ef5598f2fc5123/src/alire/alire-vcss-git.adb#L436 and https://github.com/alire-project/alire/blob/01264f9c9e1785ed19e04d41e9ef5598f2fc5123/src/alire/alire-vcss-git.adb#L443

testsuite/drivers/helpers.py
mosteo: OK, I'll use that one then.

mosteo: And I'll read through the `e3` APIs, I hadn't realized there were so many things there that can be useful.

Fabien-Chouteau (approved):

PR #755, opened 2021-06-21
Detect circularities in link pins

mosteo:
Otherwise we recurse until stack exhaustion


Fabien-Chouteau (approved):

mosteo:
Just rebases since review, so merging.

PR #757, opened 2021-06-22
Stable release 1.0.1 with fixed msys installer

mosteo:


mosteo:
Unfortunately we hit the change of compiler so this is no longer as straightforward as it could have been. Oh well.

Fabien-Chouteau (approved):

jquorning:
The action `Run test script/ALR SEARCH` emits two stack traces on Windows and Linux when executing `alr search -q -d --list --external` before crate list is shown. macOS looks fine.
This does not look like a new issue.

mosteo:
> The action `Run test script/ALR SEARCH` emits two stack traces on Windows and Linux when executing `alr search -q -d --list --external` before crate list is shown. macOS looks fine.
> This does not look like a new issue.

Yes, this is because when `-d` is in effect any internal failed operation will be logged (these involving `Outcome_Failure`) even if harmless. It's something that could be enhanced to avoid confusion.

PR #765, opened 2021-06-30
Implement sha256 and hash use test

mosteo:


Fabien-Chouteau:
Hi Alejandro, why do we need sha256?

mosteo:
I saw this was the hash generated for the GNAT releases at https://github.com/alire-project/GNAT-FSF-builds/releases/tag/gnat-10.3.0-1

We could compute the sha512, of course. I just thought that it made sense to use the same hash that's provided by the build process.

Fabien-Chouteau:
Since sha512 is our standard for crates I should also use that for the GNAT FSF packages.

mosteo:
OK, I'm going to keep this one around for a while for my tests, and I'll remove it before the final merge.

mosteo:
Merged into temporary branch.

PR #775, opened 2021-07-26
Binary crates, toolchain management and use

mosteo:
This PR implements a variety of changes to support the use of binary compilers packaged through Alire:

- New binary crates.
- `provides` property to allow mixed `gnat`/`gnat_xxx` dependencies.
   - Changes in the solver and related types to use this property properly.
   - Changes in the solver to prefer locally installed binary crates to remote available ones.
- `alr toolchain` to inspect/[un]install/select toolchains.

Still missing:
- [x] Passing existing tests.
- [x] New tests.
- [x] Scrub the code for leftover comments / self-review.

Documentation might be a bit sparse yet, and require some tweaking.

@Fabien-Chouteau , now the toolchain selection assistant is run the first time a valid alire folder is required. We might disable this entirely, or run it only if there is no gnat/gprbuild in the PATH, or something else if you see fit.


mosteo:
Hi @Fabien-Chouteau . I still have to see what's going on with Windows, but this one is ready for review. I apologize in advance because on top of being a large PR in itself, I had to do a couple of refactors towards the end to break circularities that mud things even more. I'll try to compartmentalize things better next time :sweat:

doc/toolchains.md
```diff
@@ -0,0 +1,80 @@
...
+  prioritization:
+   1. A target-specific compiler that is already a dependency.
+   1. The default compiler, if it has been defined.
+   1. A native compiler that is already deployed.
+   1. A cross-compiler that is already deployed.
+   1. A native compiler from the index, that will be deployed with the rest of dependencies.
+   1. A cross-compiler from the index, that will be deployed with the rest of dependencies.
+
+- Solution with dependencies on a target-specific GNAT (`gnat_native`, `gnat_ricv_elf`,
+  etc.): The `alr` solver will provide one compiler in the environment, applying the
```
Fabien-Chouteau, line 74: Is it possible to have both an explicit dependency on `gnat_native` and `gnat_riscv64_elf`? It should be.

doc/toolchains.md
```diff
@@ -0,0 +1,80 @@
...
+     available in the user's environment prior to running `alr`. 
+      - Note that compilers deployed with `alr toolchain --install` but not
+        selected as the default will not be used in this case.
+
+For crates with dependencies on GNAT compilers, the following two cases simultaneously apply:
+
+- Solution with generic dependencies on `gnat`: The solver will provide a
+  concrete GNAT to satisfy this dependency, applying the following
+  prioritization:
+   1. A target-specific compiler that is already a dependency.
```
Fabien-Chouteau, line 66: I don't understand this list.
If there is no default compiler defined, it should be whatever is in the PATH and gprbuild will fail if no toolchain is available.

doc/toolchains.md
```diff
@@ -0,0 +1,80 @@
+# Toolchain management
```
Fabien-Chouteau, line 1: I feel like there is a simpler way to explain the toolchain management. I'll try to propose something.

src/alire/alire-dependencies-states.ads
```diff
@@ -102,6 +96,11 @@ package Alire.Dependencies.States is
 
    function Is_Pinned (This : State) return Boolean;
 
+   function Is_Provided (This : State) return Boolean;
+   --  True when the release name is different from the dependency crate
+
+   function Is_Shared (This : State) return Boolean;
```
Fabien-Chouteau, line 102: Should it be `Is_Installed`?

doc/toolchains.md
mosteo: Simultaneously in a crate? How's that supposed to work?

doc/toolchains.md
mosteo: This is how the solver tries to find a release for `gnat`. Do you mean that the solver should not bring in compilers that aren't already installed? (At the moment, if a crate requests a compiler and none is available, it will pick one in the index and auto-install it).

src/alire/alire-dependencies-states.ads
mosteo: I started using the idea of 'shared releases' internally, as these are just regular releases in a shared location. But it can be changed if confusing (likewise then for the `Alire.Shared` package).

doc/toolchains.md
Fabien-Chouteau: > Do you mean that the solver should not bring in compilers that aren't already installed? (At the moment, if a crate requests a compiler and none is available, it will pick one in the index and auto-install it).

Not if the user didn't specify a default, and there is no explicit dependency on `gnat_native`, `gnat_arm_elf`, etc.

If the user decided not to have a default native compiler, which I would say I hope not to be a very common case, then for me it means he/she is in charge of providing the compiler.

doc/toolchains.md
Fabien-Chouteau: I have one video game crate (`shoot_n_loot`) that builds both the embedded version of the game and a native simulator to play it on the computer. But maybe that's a very bad idea ^^

src/alire/alire-dependencies-states.ads
Fabien-Chouteau: It will depend on how and if we implement an `alr install` I guess. We can keep shared for now, up to you.

doc/toolchains.md
mosteo: OK, I understand. So `gnat` dependencies only rely on locally available compilers. Should be easy to fix.

doc/toolchains.md
mosteo: I actually was worried about this use case but I guess I didn't traslate it properly in our talks. This is the one where I saw your idea on conditional dependencies on something provided via command line/dependent crate configuration could be useful:
```
alr build target:=native  # arbitrary name, not predefined (or defined in the configuration section of the crate)
```
and then we can have in the toml:
```
[[depends-on."case(crate.target)".native]]
gnat_native = "*"
[[depends-on."case(crate.target)".riscv]]
gnat_riscv_elf = "*"
```

There should be a default somehow for automated builds though.

For a later date obviously...

Otherwise, I guess we can do an exception about the `gnat` crate not being conflicting among compilers. But something nags me at that being unkind to people not interested in all targets.

doc/toolchains.md
mosteo: Right. As a professor I'm not very proud of the several convoluted explanations I've come up with.

src/alire/alire-dependencies-states.ads
mosteo: My current mood is to avoid `install` entirely. I was thinking of using `alr share` for non-toolchain crates shared in the configuration, and `alr export` for the `gprinstall-in-common-prefix` idea. So we convey that exported crates are something the user wants to use out of Alire. But I'm easy to sway on these naming things.

For now if you don't mind I'll leave it as is then.

doc/toolchains.md
mosteo: This is implemented and checked in a new test in the last commit (328e0512546bf970a6f6c24cf7cd9e2b84af7a3a).

As of now, and correcting my last message, to avoid any surprises for the user, an external compiler will be preferred unless the user selects its preference, or the crate specifies a targeted gnat (`gnat_something`).

BTW while installed compilers are for the whole config, we can make the selected compiler optionally a local config option, so different working crates may select a different compiler among the installed ones. This can be a follow-up PR.

src/alire/alire-dependencies-states.ads
onox: If the user wants to make some application available system-wide, I would say the usual keyword used by package managers and build systems is `install` instead of `export`. `export` sounds like an action to extract data from an application.

doc/toolchains.md
Fabien-Chouteau: Let's keep it simple for now and only allow one compiler.

Fabien-Chouteau: I have one other use case. 

A cross project that uses native tools to generate code, for instance https://alire.ada.dev/crates/are.

mosteo: Right, that is a legitimate situation that we ought to support. I'll try to come up with a solution in a separate PR.

mosteo: Addressed in https://github.com/alire-project/alire/pull/780

src/alire/alire-dependencies-states.ads
mosteo: In this case we would be "extracting" binaries from the Alire ecosystem. But I see your point. `alr share` for shared dependencies across workspaces and `alr install` for deploying crates to a common prefix can also work for me.

Another possibility is to have it under `get`, e.g., `alr get --install`. The operation in itself is essentially an `alr get + alr build + gprinstall`.

mosteo:
Hi @Fabien-Chouteau , with #780 and #781 in the pipeline I think we are ready feature-wise for this one. There is the matter of whether to disable sha256, and the reworded explanation in `toolchains.md`. I can take another stab at it if you want, unless you're already on it.

Fabien-Chouteau:
Thanks @mosteo 

> I can take another stab at it if you want, unless you're already on it.
Unfortunately no, I didn't take the time to work on this.

Maybe we can do a release candidate and ask for feedback. Based on the feedback we could be able to improve the doc...

mosteo:
OK then, in fact I should update the doc in #780 at least to match the current situation.

PR #777, opened 2021-08-02
Alr.Commands.Init: change defaults for init project to release mode

Fabien-Chouteau:
I first decided to make the default better for developers (to catch as many errors as possible), but they end up in release crates which means that the performance is bad as well as the compatibility between compiler versions.

This patch set the defaults to max performance and compatibility.

This system is clearly not very user friendly, maybe we have too many options here.
Profiles like in Cargo could be better.

### DIFF
--- a/src/alr/alr-commands-init.adb
+++ b/src/alr/alr-commands-init.adb
@@ -147,13 +147,13 @@ package body Alr.Commands.Init is
          end if;
          Put_New_Line;
          Put_Line ("   type Enabled_Kind is (""enabled"", ""disabled"");");
-         Put_Line ("   Compile_Checks : Enabled_Kind := External (""" & Upper_Name & "_COMPILE_CHECKS"", ""enabled"");");
-         Put_Line ("   Runtime_Checks : Enabled_Kind := External (""" & Upper_Name & "_RUNTIME_CHECKS"", ""enabled"");");
-         Put_Line ("   Style_Checks : Enabled_Kind := External (""" & Upper_Name & "_STYLE_CHECKS"", ""enabled"");");
-         Put_Line ("   Contracts_Checks : Enabled_Kind := External (""" & Upper_Name & "_CONTRACTS"", ""enabled"");");
+         Put_Line ("   Compile_Checks : Enabled_Kind := External (""" & Upper_Name & "_COMPILE_CHECKS"", ""disabled"");");
+         Put_Line ("   Runtime_Checks : Enabled_Kind := External (""" & Upper_Name & "_RUNTIME_CHECKS"", ""disabled"");");
+         Put_Line ("   Style_Checks : Enabled_Kind := External (""" & Upper_Name & "_STYLE_CHECKS"", ""disabled"");");
+         Put_Line ("   Contracts_Checks : Enabled_Kind := External (""" & Upper_Name & "_CONTRACTS"", ""disabled"");");
          Put_New_Line;
          Put_Line ("   type Build_Kind is (""debug"", ""optimize"");");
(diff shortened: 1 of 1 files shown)


mosteo (approved):

onox:
@Fabien-Chouteau Shouldn't the Runtime_Checks be enabled by default? Otherwise it will add `-gnatp`.

onox:
I think it would help to have the build mode control which checks get enabled. This could be something like the user being able to add `--release`, `--develop`, `--debug`, `--coverage`, and `--profiling`. And have this applied to dependencies as well when running `alr build`. Perhaps the list of modes could be standardized as well and have every .gpr file use a well-known env var, something like `ALR_BUILD_MODE`. I usually add "coverage" for `-fprofile-arcs` and `-ftest-coverage` and "profiling" for `-pg`.

Fabien-Chouteau:
> @Fabien-Chouteau Shouldn't the Runtime_Checks be enabled by default? Otherwise it will add `-gnatp`.

I decided to go for full performance.

> I think it would help to have the build mode control which checks get enabled. This could be something like the user being able to add `--release`, `--develop`, `--debug`, `--coverage`, and `--profiling`. And have this applied to dependencies as well when running `alr build`.

Yes, that would be interesting but there are several issues:
 - This is done in the gpr project files which are not "standardized"
 - Things like coding style will be different for each project
 - In my opinion we need fine control of which dependency to build with each mode

I think Cargo profiles are a good source of inspiration: https://doc.rust-lang.org/cargo/reference/profiles.html

There might be a way to implement this kind of profile using the ["Crate configuration"](https://doc.rust-lang.org/cargo/reference/profiles.html) feature.

We could have an implicit configuration variable for build profile that is available in every crate:
```toml
Build_Mode = {type = "Enum", values = ["Release", "Develop", "Check", "Coverage"], default = "Release"}
```
This would generate the following project file:
```ada
abstract project XYZ_Config is
   type Build_Mode_Kind is ("Release", "Develop", "Check", "Coverage");
   Build_Mode : Build_Mode_Kind := "Release";
```

To which Alire could automatically add another declaration that provides corresponding switches depending on the value of `Build_Mode`:
```ada
abstract project XYZ_Config is
   type Build_Mode_Kind is ("Release", "Develop", "Check", "Coverage");
   Build_Mode : Build_Mode_Kind := "Release";
   GNAT_Switches := ("-O3",    -- Optimization
                     "-gnatp", -- Supress checks
                     "-gnatn"  -- Enable inlining
                    );
```
One could change the mode of a dependency:
```toml
[[depends-on]]
xyz = "*"
[configuration.values]
xyx.Build_Mode = "Develop"
```
And that would generate:
```ada
abstract project XYZ_Config is

   type Build_Mode_Kind is ("Release", "Develop", "Check", "Coverage");
   Build_Mode : Build_Mode_Kind := "Develop";
   GNAT_Switches := ("-Og",     -- Optimization
                     "-g",      -- Debug info
                     "-gnata",  -- Enable assertions and contracts
                     "-gnatn",  -- Enable inlining
                     "-gnatwa", -- All warnings
                     "-gnatVa", -- All validity checks
                     "-gnatwe"  -- Warnings as errors
                    );
```

If someone doesn't like the list of switches generated by Alire, they don't have to use it and can make their own list.

> Perhaps the list of modes could be standardized as well and have every .gpr file use a well-known env var, something like `ALR_BUILD_MODE`. I usually add "coverage" for `-fprofile-arcs` and `-ftest-coverage` and "profiling" for `-pg`.

In my own experience, I want to be able to select different build mode for different dependency. 
Also using the same GPR external variable name is a bad idea as different declarations in different project files will lead to an error. That's why we changed the names of many GPR external in AdaCore libraries.

Fabien-Chouteau:
So... I made a quick prototype: https://github.com/alire-project/alire/pull/779

PR #780, opened 2021-08-05
Releases providing the same 3rd crate no longer implicitly conflict

mosteo:
Follow up to #775. With this, releases that provide the same 3rd release are compatible in a solution, covering the use case discussed about having both a native and a cross-compiler among the dependencies of a crate (so the native can be used to generate a tool that is necessary to prepare the cross-compilation).

One test removed and another added for this new situation.


Fabien-Chouteau (approved):

doc/toolchains.md
```diff
@@ -2,34 +2,60 @@
...
+Releases may still override this selection (for example to use a
+cross-compiler).  
+
+There are two kind of dependencies on GNAT compilers: generic dependencies on
+the `gnat` crate, which apply to every compiler, and dependencies on a precise
+native or cross-compiler.
+
+The native compiler for each platform is always `gnat_native`, whereas
+cross-compilers follow the naming convention `gnat_<target>` (e.g.,
+`gnat-riscv_elf`). Alire will only offer to install valid cross-compilers for
```
Fabien-Chouteau: `s/gnat-riscv_elf/gnat_riscv_elf/`

Fabien-Chouteau (approved):

PR #781, opened 2021-08-05
Reinstate "forbids" property

### LINKED ISSUE #105: forbids field for conflicting dependencies
Alire releases can specify a set of forbidden dependencies, that are known to be conflicting (e.g., they contain sources with the same filename).

There's no current mention to this in the index syntax. In essence, that would be a key like depends_on:
```
[depends-on]
foo="^1.0"
[conflicts-with]
bar="~1.4"
```

The original use case was for Matreshka, which brings its own dependencies that can be already in alire, and were not easy to extricate. Although, long term, that should not happen, I guess this is a feature that can come in handy in the short term, and maybe for unforeseen conflicting circumstances.

Pinging @pmderodat for awareness (and can you think of other reasons for conflicts that are not name clashes?)

mosteo:
Fixes #105.

We had this implemented but no way to use it because there was no implementation for loading it from the TOML index. This PR provides this missing load functionality and a couple of tests for it.

The `forbids` field has the same syntax as the `depends-on`. It serves to specify releases (in the form of dependencies) that should not appear in the same solution as the release using the `forbids` declaration. The two main use cases are to prevent known conflicts (e.g., two crates that have some sources with the same filenames), and to allow drop-in replacements with `provides`. In this second case it works the same as [conflicts](https://www.debian.org/doc/manuals/debian-reference/ch02.en.html#_package_dependencies) in `apt-get`: a crate may declare that it both provides and forbids a second crate. This way, it can fill for the provided crate, and no other crate will also be used for the provided crate.


Fabien-Chouteau (approved):

Fabien-Chouteau:
It would be good to have "user changes" entry for this.

PR #798, opened 2021-08-19
Enable and update self-build

mosteo:
This was disabled and forgotten during the pins rework.


Fabien-Chouteau:
We should wait before updating to a CLIC that uses `src/clic-command_line.adb`.

Fabien-Chouteau (approved):

PR #799, opened 2021-08-19
Improvements to cleanup of temp files

mosteo:
- `alr clean --temp` will now also clean the configuration cache
- Temp files are tracked and deleted upon Ctrl-C

### DIFF
--- a/src/alire/alire-directories.adb
+++ b/src/alire/alire-directories.adb
@@ -12,6 +12,8 @@ with Alire.TTY;
 
 with GNATCOLL.VFS;
 
+with SI_Units.Binary;
+
 package body Alire.Directories is
 
    package Adirs renames Ada.Directories;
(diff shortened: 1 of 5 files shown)


src/alire/alire-directories.adb
```diff
@@ -12,10 +12,64 @@ with Alire.TTY;
...
+         Registry.Include (Ada.Directories.Full_Name (Path));
+      end Add;
+
+      ---------
+      -- Del --
+      ---------
+
+      procedure Del (Path : Any_Path) is
+      begin
+         Registry.Exclude (Path);
```
Fabien-Chouteau, line 61: Will the relative path match the absolute path added above?

I think all the subprograms here should probably only accept absolute paths.

mosteo: You're right. I think these paths are all absolute anyway or it wouldn't be working already, but I'll make sure.

mosteo: Should be fixed now.

Fabien-Chouteau (approved):

Issue #707, opened 2021-03-10
Build modes


onox:
Would it be possible to build a whole tree of crates with a certain mode? Like `alr build --release` or `alr build --debug`?

Perhaps `alr build` could support a few modes, like release, develop, debug, coverage, profiling?

Some of the modes could enable/disable style checks and/or contract checks.

Fabien-Chouteau:
Hello @onox,

Right now the question of build mode is highly project specific, there is no standard way to specify that.

PR #806, opened 2021-08-24
Extract sub-command handling (alr <command>)

Fabien-Chouteau:
The goal of this big refactoring is to extract the handling of
sub-commands to make it available for other projects.

The SubCommander packages (name open to changes) can be extracted to a
dedicated crate or to AAA.

There is at least one regression which is the support of -X for alr
build.


src/alr/alr-commands-topics-naming_convention.ads
```diff
@@ -0,0 +1,23 @@
...
+
+with Alire.Crates;
+
+package Alr.Commands.Topics.Naming_Convention is
+
+   type Topic is new SubCommander.Help_Topic with null record;
+
+   overriding
+   function Name (This : Topic) return String
+   is ("indentifiers");
```
mosteo: typo

src/alr/alr-commands.adb
```diff
@@ -516,7 +207,7 @@ package body Alr.Commands is
    procedure Reportaise_Wrong_Arguments (Message : String) is
    begin
       Alire.Errors.Pretty_Print (Message);
-      raise Wrong_Command_Arguments with Alire.Errors.Set (Message);
+      raise SubCmd.Wrong_Command_Arguments with Alire.Errors.Set (Message);
```
mosteo: I find this mix of `CamelCase` and `Camel_Case` quite jarring... I'd prefer to stick to the latter style.

src/alr/subcommander.ads
```diff
@@ -0,0 +1,86 @@
+with AAA.Strings;
+with GNAT.Command_Line;
+
+package SubCommander is
+
+   --  This root package defines the interface types to be unsed in the
```
mosteo: s/unsed/used/

src/alr/subcommander-instance.adb
```diff
@@ -0,0 +1,861 @@
...
+      Hash            => Ada.Strings.Unbounded.Hash,
+      Equivalent_Keys => Ada.Strings.Unbounded."=");
+
+   Registerd_Commands : Command_Maps.Map;
+   --  A map of commands based on their names
+
+   Registerd_Topics : Topics_Maps.Map;
+   --  A map of topics based on their names
+
+   Registerd_Groups : Group_Maps.Map;
```
mosteo: Should be `Registered`?

mosteo (changes requested):
Neat! I just saw very little things during the review. There is some ugliness that I crafted in there, ouch. Way better having it isolated.

I understand that now `alr build` simply doesn't see the `-X` switches. I remember that someone was using that. Perhaps this PR can go in after `1.1` then.

Fabien-Chouteau:
@mosteo I did my final refactoring on this.

Pending checks and reviews, the remaining questions are:
 - Should the name `SubCommander` change?
 - Where should we put this code? AAA or its own crate?

mosteo (approved):

mosteo:
Huge effort here, thanks Fabien.

> Where should we put this code? AAA or its own crate?

Nowadays I don't see a reason for huge libraries, so I favor splitting things unless they're very thematically related. AAA is going to probably be my last hodgepodge project. So I'd say its own crate.

> Should the name SubCommander change?

Besides the camel case (I'd prefer simply `Subcommander`) I don't have ideas for another name right away...

Fabien-Chouteau:
> Huge effort here, thanks Fabien.

Thank you, I was not expecting to fall in this rabbit hole ^^

> > Where should we put this code? AAA or its own crate?
> 
> Nowadays I don't see a reason for huge libraries, so I favor splitting things unless they're very thematically related. AAA is going to probably be my last hodgepodge project. So I'd say its own crate.

Thinking about this, we have other things like TTY, user input, config files that we can extract from Alire in to a Command Line Interface crate:
For the name?:
 - `clic`: Command Line Interface Components
 - `clic`: Command Line Interface Crate
 - Just `cli`
 - Something else...

> > Should the name SubCommander change?
> 
> Besides the camel case (I'd prefer simply `Subcommander`) I don't have ideas for another name right away...

Ok for `Subcommander`.

mosteo:
> * `clic`: Command Line Interface Components

I like this one a lot, `cli` is too generic but `clic` has personality

onox:
> I remember that someone was using that. Perhaps this PR can go in after `1.1` then.

I use `-X` to set some scenario vars (one to switch between AVX and AVX2 code (I run some `gcc` command in a Makefile to detect which extension is supported)) and another to switch between release and coverage build modes. Although the latter wouldn't be needed if something like #707 is implemented.

Fabien-Chouteau:
@mosteo I successfully moved `Subcommander`, `TTY` and `User_Input` under a `CLIC` package hierarchy. Next week I will move `CLIC` to a dedicated repository in `alire-project`.

I think `Alire.Config` can also move to `CLIC` but there is some refactoring to do.

mosteo:
Thanks for the update, @Fabien-Chouteau .

@onox, the `-X` parameters are supported again in this PR, so the situation stays as it was.

.gitmodules
```diff
@@ -45,3 +45,6 @@
 [submodule "deps/si_units"]
 	path = deps/si_units
 	url = https://github.com/HeisenbugLtd/si_units.git
+[submodule "deps/clic"]
+	path = deps/clic
+	url = git@github.com:alire-project/clic
```
mosteo: I suspect this git@ URL won't work for people outside of the alire-project org.

mosteo (approved):

.gitmodules
onox: This should require and use the SSH key of the user. Better change it to `https://github.com/alire-project/clic`? :slightly_smiling_face:

Fabien-Chouteau: Will fix. I had to do it that way when the repo was private.

mosteo (approved):

mosteo:
Thanks a lot, Fabien, now merged.

PR #810, opened 2021-08-28
Alire.Crate_Configuration: generate a constant with the version of the crate

Fabien-Chouteau:
With this addition users only have to worry about the version in
alire.toml and it will be propagated to code and gpr files.


src/alr/alr-commands-init.adb
```diff
@@ -114,7 +114,8 @@ package body Alr.Commands.Init is
             return;
          end if;
          Put_Line ("abstract project " & Mixed_Name & "_Config is");
-         TIO.Put (File, "end " & Mixed_Name & "_Config;");
+         Put_Line ("   Crate_Version := ""0.0.0"";");
```
mosteo, line 117: Is this `0.0.0` proper? Seeing that in other places it's replaced by the actual version.

Fabien-Chouteau: This is for `alr init` where we generate all the files by hand. `alr init` starts the crates with version `0.0.0`, so that's also what I am doing here.

We might want to do an `alr update` after generating the files from `alr init`, it would make this kind of things easier.

mosteo: Ah, right, thanks for refreshing my mind.

mosteo (approved):

PR #832, opened 2021-09-09
Incremental building of dependencies

### LINKED ISSUE #628: Make executables in dependencies available to dependents
If I have a tool dependency, I want to be able to use it during pre-build actions
- Only affects dependencies with declared executables
- Can the build of such dependencies be launched recursively before pre-build actions?
   - Are pre-build actions in dependencies being run right now?

Related:
- cargo install -> .cargo/bin/<exec>
    - [ ] What happens with libraries?
    - [ ] Can a binary be added as a dependency [in cargo]?

Related to #482

mosteo:
Fixes #628 

For context: we want to build dependencies in a safe order, so tools built by a dependency become available down the building chain.

@Fabien-Chouteau , this is istill work in progress, but I've hit a situation. When trying to build `shoot_n_loot` with this patch, it fails because it tries to build the project files in `samd51_hal`. My question would be if this is a problem with our current idea, or with some crate configuration.

Right now, the feature works by trying to build all project files in dependencies that have `auto_gpr_with` enabled.  Perhaps we could restrict it further to crates explicitly declaring executables?

### DIFF
--- a/src/alire/alire-os_lib-subprocess.adb
+++ b/src/alire/alire-os_lib-subprocess.adb
@@ -107,7 +107,7 @@ package body Alire.OS_Lib.Subprocess is
       if Exit_Code /= 0 then
          Raise_Checked_Error
            ("Command " & Image (Command, Arguments) &
-              " exited with code" & Exit_Code'Img);
+              " exited with code " & Utils.Trim (Exit_Code'Image));
       end if;
    end Checked_Spawn;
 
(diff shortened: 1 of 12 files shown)


Fabien-Chouteau:
> Perhaps we could restrict it further to crates explicitly declaring executables?

Yes that's a better option because the libraries can be compiled for different targets and Alire doesn't know that.

Fabien-Chouteau (approved):

PR #834, opened 2021-09-09
Fix missing actions triggers

### LINKED ISSUE #633: `post-fetch` and `pre-build` actions not running at all expected moments

mosteo:
Fixes #633 

Furthermore, the semantics of `post-fetch` were not entirely well-defined. The intention is to run it whenever sources are deployed, so just to err on the safe side, `post-fetch` is now run after any deployment or update, for all releases in the solution that have them, in case there is some interdependence.


mosteo:
Not sure why the first two commits are showing, as the `next` branch is clearly at the second commit (the merge). It makes the review a bit more cumbersome.

Issue #496, opened 2020-08-13
Running unit tests


m-kru:
Are there any plans to add command for running unit tests? `alr test` exists, but it `Tests the compilation of all or some releases`.

mosteo:
Not right now, because `alr test` is intended to verify integration into alire, not integrity of the release itself. But I see how it could be a useful part of the development workflow, so thanks for the suggestion.

kalkin:
I also would like to specify some dependencies as test only, so the consumer of the library does not need e.g. `aunit`.

onox:
> I also would like to specify some dependencies as test only

Many package managers allow you to specify a list of "developer dependencies". These are dependencies that are not downloaded and built when the crate is used as a dependency of another crate, but only when performing certain tasks in the project of the crate. These include tools like linters, code formatters, dependencies for tests, etc.

In this case `aunit` should indeed be added to this list of developer dependencies. In my own projects I had to workaround this missing feature by creating an extra unpublished "tests" crate with `aunit` as a regular dependency.

IMO `alr test` should run (unit) tests and perhaps another command like `alr prove` could run the `gnatprove` tool.

kalkin:
@onox Thanks for your reply. I solved it actually the same way you proposed :-).

mosteo:
@onox , IIUC, this could go hand in hand with customizable user tasks? Something like:

```
[[task]]
name = "prove"
command = ... # what has to be run for this task
[task.depends-on]
# Extra dependencies
[task.environment]
# Extra environment

[[task]]
name = "unit-testing"
command = ... 
```

and then you issue `alr task prove` or `alr task unit-testing`, etc? Overkill?

onox:
Customizable user tasks are definitely useful, but I'm not sure about specifying the dependencies per task :thinking: Other package managers like poetry, cargo, yarn/npm seem to allow you to specify just 1 list called `dev-dependencies`. It could work exactly like `[[depends-on]]`, supporting `'case(os)'`, etc.

Would it make sense to have something like `[dev-depends-on.'case(task).A|B']` so you can specify a particular dev-dependency for multiple tasks? A project may have tasks like `lint:check` and `lint:fix`, both requiring some tool.

mosteo:
Thanks for the feedback, onox. Just exploring ideas for now. My concern is that some task my have a "scarce" dependency that might raise the entry cost for other would-be developers.

kalkin:
[...]

PR #847, opened 2021-09-14
New On_Demand action type

mosteo:
The topic has arisen about the possibility of defining custom user tasks (e.g. #496). I realized that we essentially have that already via actions.

So, to not break the index yet again, this PR allows giving a name to an action, and to use a new action type `on-demand`. Currently unusable, and it can stay undocumented for `1.1`.

The idea would be to have in the future an `alr action <name>` that manually triggers the action. And, since actions already can take a relative path, this saves even the need for a wrapper script. For example, to run some unit tests:
```
[[actions]]
type = "on-demand"
name = "unit-tests"
directory = "utests"
command = ["alr", "run"]
```
(Here I'm assuming `utests` contains a nested crate with a default executable that runs the testing framework.)

Likewise one could run examples, etc.

No test yet included until we discuss it.

### DIFF
--- a/src/alire/alire-properties-actions-runners.adb
+++ b/src/alire/alire-properties-actions-runners.adb
@@ -1,3 +1,5 @@
+with AAA.Enum_Tools;
+
 package body Alire.Properties.Actions.Runners is
 
    -------------
@@ -56,6 +58,9 @@ package body Alire.Properties.Actions.Runners is
          Has_Name : Boolean;
          Path     : TOML_Value;
          Has_Path : Boolean;
+         Moment   : Moments;
+
+         function Is_Valid is new AAA.Enum_Tools.Is_Valid (Moments);
       begin
          if not From.Pop (TOML_Keys.Action_Type, Kind) then
             From.Checked_Error ("action type missing");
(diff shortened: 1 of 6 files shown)


Fabien-Chouteau:
This is in the same area as my current work:
```console
$ alr exec bash -c 'echo $PATH'
   [run bash in Alire context/environment]
```
Combined with aliases
```console
$ alr config --global --set alias.prove "exec gnatprove"
$ alr prove
  [runs gnatprove in Alire context/environment]
```
I think the `on-demand` action is superior for some use cases because it's going to be included in the releases (unlike aliases).
I am a bit concerned about the potential can of worms that this opens, concerned about the kind of actions that users will publish in their releases.

mosteo:
I see these two as complementary. Aliases as a convenience for developers to tailor alr to their needs (like git aliases), whereas the actions, although they can do something in that direction, are more for the user of the crate, as they're tied to a particular crate.

It is true that having open-ended actions can be "exploited" by imaginative people. In this sense I don't have much experience, I just see that this exists in other PMs and some people seem to want something like it.

npm has [npm run](https://docs.npmjs.com/cli/v7/commands/npm-run-script) (same as `alr action`) but no `exec` in the sense talked here (it has [npm exec](https://docs.npmjs.com/cli/v7/commands/npm-exec) for remote execution?)

opam only has [opam exec](https://opam.ocaml.org/doc/man/opam-exec.html) (same as `alr exec`)

yarn has both [yarn run](https://yarnpkg.com/cli/run) and [yarn exec](https://yarnpkg.com/cli/exec) (same as both meanings discussed)

cargo has none by default, but there seems to be several [user](https://github.com/JoshMcguigan/cargo-run-script) implementations, as you can add anything you want to cargo as a subcommand (yikes?)

All in all it seems both are somewhat out there. We might merge the loader just in case and decide about the feature later on. Currently I'm more concerned that if we decide to allow it we'll have to change the index version again.

Fabien-Chouteau (approved):

PR #850, opened 2021-09-15
Alr.Commands.Build: forward subcommand args to gprbuild

Fabien-Chouteau:


src/alire/alire-spawn.adb
```diff
@@ -40,13 +40,15 @@ package body Alire.Spawn is
 
       Command ("gprbuild",
                Empty_Vector &
+
                  "-gnatwU" &
```
Fabien-Chouteau, line 44: @mosteo I never noticed this one before. Maybe it can be removed?

mosteo: I don't really remember, but I see below that you found some need for it?

mosteo (approved):

src/alire/alire-spawn.adb
Fabien-Chouteau: I just moved the command that drifted away from the switch. But I don't why it is needed, or if it is needed,

PR #853, opened 2021-09-16
New command exec and support of aliases

Fabien-Chouteau:


src/alr/alr-commands-exec.adb
```diff
@@ -0,0 +1,57 @@
...
+                & " build environment.")
+       .New_Line
+       .Append ("--cache:")
+       .Append ("   All downloaded dependencies will be deleted.")
+       .New_Line
+       .Append ("--temp:")
+       .Append ("   All alr-???.tmp files in the subtree will be deleted."
+                & " These files may remain when alr is interrupted via"
+                & " Ctrl-C or other forceful means.")
+      );
```
mosteo: Copy-pain? :-)

mosteo (changes requested):
Nice feats, I'll have `alr list` back :-)

mosteo:
An observation: from my testing it seems that once you define an alias, you cannot pass extra switches. For example, if `display` is an alias for `show`, running `alr display --graph` will ignore the switch. Would this make sense?

src/alr/alr-commands-exec.adb
Fabien-Chouteau: Indeed :man_facepalming:

Fabien-Chouteau:
> An observation: from my testing it seems that once you define an alias, you cannot pass extra switches. For example, if `display` is an alias for `show`, running `alr display --graph` will ignore the switch. Would this make sense?

It's not supposed to do that... I will check.

Fabien-Chouteau:
@mosteo 

It works on my side:
```console
$ alr config --set --global alias.display show
$ alr display --graph
# Synchronizing workspace...
Dependencies automatically updated as follows:  
[...]
Dependencies (graph):
                                      +-----------------------++-------------------+
```

mosteo:
Right you are... I messed something up in my previous attempt, sorry for the noise.

Fabien-Chouteau:
@mosteo I think I should introduce the `--` double-dash delimiter for `alr exec` and `alr build`.
In case we want options for these commands in the future, e.g.:
```console
$ alr build --mode=release -- -XTEST=plop -cargs -g # only "-XTEST=plop -cargs -g" is passed to gprbuild
```

what do you think?

mosteo:
Sounds very sensible

onox:
> @mosteo I think I should introduce the `--` double-dash delimiter for `alr exec` and `alr build`.

Would this help to allow global options be moved back to after the command? Having 1 place where the user can define global and command-specific options would be easier for users.

Fabien-Chouteau:
Global options would still have to be placed before the command. It's cleaner to separate the two in my opinion.

mosteo (approved):

mosteo:
Another great couple of features; thanks, Fabien.

PR #861, opened 2021-09-20
Disable assistant on --select <release>

mosteo:
We can make a patch release, or disable it in the `setup-alire` action whenever `--select` is used, or both.


mosteo:
Wrong base branch and now tests won't run properly, so closing.

PR #866, opened 2021-09-22
Bugfix for the assistant when there is external gnat but not gprbuild

### LINKED ISSUE #865: Constraint error when running alr update using Alire 1.1.0
I recently tried running `alr update` using alire version 1.1.0 on Linux. I see the following constraint error before it provides a selection prompt for gprbuild. If I select gnat_native instead of gnat_external, this error doesn't occur. I didn't compile with debug information so I the stack trace isn't very useful. I can try recompiling with that information. In case that debug info isn't needed, I provided the error below.

```
$ alr update -d
Welcome to the toolchain selection assistant

In this assistant you can set up the default toolchain to be used with any crate
that does not specify its own top-level dependency on a version of gnat or 
gprbuild.

If you choose "None", Alire will use whatever version is found in the 
environment.

ⓘ gnat is currently not configured. (alr will use the version found in the environment.)

Please select the gnat version for use with this configuration
  1. gnat_native=11.2.1
  2. None
  3. gnat_external=11.1.0 [Detected at /usr/bin/gnat]
  4. gnat_arm_elf=11.2.1
  5. gnat_riscv64_elf=11.2.1
  6. gnat_arm_elf=10.3.1
  7. gnat_native=10.3.1
  8. gnat_riscv64_elf=10.3.1
Enter your choice index (first is default): 
> 3
ⓘ Selected tool version gnat_external=11.1.0

ⓘ Choices for the following tool are narrowed down to releases compatible with just selected gnat_external=11.1.0

ⓘ gprbuild is currently not configured. (alr will use the version found in the environment.)

stderr: CONSTRAINT_ERROR                    
stderr: AAA.Strings.Vectors.Insert: Before index is out of range (too large)
stderr: raised CONSTRAINT_ERROR : AAA.Strings.Vectors.Insert: Before index is out of range (too large)
0x5583e2d18a2a
0x5583e2b2736e
0x5583e2b2b766
0x5583e2b0f66d
0x5583e22fcae7
0x5583e22b0e26
0x5583e22f49d7
0x5583e22fe67c
0x5583e21faaaa
0x5583e21fd5c2
[/usr/bin/..//lib/libc.so.6]
0x7f2f94330b23
0x5583e21fa64c
0xfffffffffffffffe

error: AAA.Strings.Vectors.Insert: Before index is out of range (too large)
error: alr encountered an unexpected error, re-run with -d for details.
```

mosteo:
Fixes #865

### DIFF
--- a/dev/devbuild.sh
+++ b/dev/devbuild.sh
@@ -1,3 +1,3 @@
 #!/bin/bash
 
-gprbuild -j0 -r -p -P `dirname $0`/../alr_env.gpr
+gprbuild -j0 -r -p -P `dirname $0`/../alr_env.gpr "$@"
--- a/src/alire/alire-toolchains.adb
+++ b/src/alire/alire-toolchains.adb
@@ -152,7 +152,7 @@ package body Alire.Toolchains is
          --  position.
 
          if Result.Choices.First_Element = None and then
-           Natural (Result.Choices.Length) >= 1
+           Natural (Result.Choices.Length) > 1
          then
             Result.Choices.Delete_First;
             Result.Targets.Delete_First;


dev/devbuild.sh
```diff
@@ -1,3 +1,3 @@
 #!/bin/bash
 
-gprbuild -j0 -r -p -P `dirname $0`/../alr_env.gpr
+gprbuild -j0 -r -p -P `dirname $0`/../alr_env.gpr "$@"
```
Fabien-Chouteau, line 3: I don't understand this change.

Fabien-Chouteau (approved):

dev/devbuild.sh
mosteo: Mmm yes, this is unrelated, just to be able to pass extra args to this script. I was fighting a "-fPIC" error but in the end the problem was elsewhere.

PR #870, opened 2021-09-24
Use Stopwatch for elapsed time image (#868)

### LINKED ISSUE #868: Saturate elapsed build time
I have been playing with `alr test --full` lately. Several times a Constraint_Error occurred when the computer sleeps during building. The problem occurs in `air-testing-markdown.adb:75` and probably in `alr-testing-console.adb` otherwise. Both places the elapsed build times are converted to a fix point type (called CS) for formatting. The range of CS is ok, but the conversion should be saturated/clamped.

I can make a PR solving the problem, but I would like to know if I should duplicate then code in the two packages or if a common conversion should be introduced in `air-testing`, `alr-util` or elsewhere.

jquorning:
`Stopwatch.Image` does not limit the range of elapsed duration. No need for saturation.


mosteo (approved):

mosteo:
Hi @jquorning , could you rebase this one on top of `master`, or merge `master` into it? A recent PR might have fixed the build failure.

mosteo:
Thanks! It worked.

PR #875, opened 2021-10-04
Bugfix for improper check related to virtual crates

mosteo:


src/alire/alire-solutions.adb
```diff
@@ -1202,7 +1202,7 @@ package body Alire.Solutions is
       end if;
 
       for Dep of This.Dependencies loop
-         if Dep.Has_Release and then Dep.Release.Provides (Dep.Crate) then
+         if Dep.Has_Release and then Dep.Release.Provides (Crate) then
```
mosteo, line 1205: This was the actual bug, ouch.

Fabien-Chouteau (approved):

Fabien-Chouteau:
I forgot in the review, but is there a tests that we can add to make sure doesn't happen again?

mosteo:
I'll prepare a reproducer, but given the new assertions in the code the root problem will have to be from other quarters for this to happen again.

Fabien-Chouteau:
@mosteo I feel like I hit this bug a lot. Should we make a patch release for it?

mosteo:
Indeed. Only let me check that #880 is not affecting the 1.1 branch because that one would also nice to have fixed.

Issue #864, opened 2021-09-21
Side effect of gprbuild checkout


reznikmm:
Hello!
I have this GitHub Action workflow:
```
on:
  push:
    branches:
      - master
name: Build binaries
jobs:
  build:
    name: Build and deploy
    runs-on: windows-latest
    steps:
      - name: Get sources
        uses: actions/checkout@v2
      - name: Get gprbuild
        uses: actions/checkout@v2
        with:
          repository: AdaCore/gprbuild
          path: gprbuild
      - name: Get GNAT toolchain with alire
        uses: alire-project/setup-alire@v1
        with:
          toolchain: gnat_native^11 gprbuild^21 --disable-assistant
          toolchain_dir: ./cached_gnat
```

The `alire-project/setup-alire@v1` step fails on windows (works on macos/linux). Error is:

```
Note: gprbuild=21.0.1 installed successfully.
Installed alr version and GNAT toolchain:
D:\a\test\test\alire_install\bin\alr.exe -n version
Alr version: 1.1.0
Alire Library version: 1.1.0
alr status is (outside) (0 releases indexed) (loaded in 0.046s)
config folder is C:\Users\runneradmin\.config\alire
source folder is C:\Users\runneradmin\.config\alire\alire
interaction flags are: force:FALSE not-interactive:TRUE
alr root is empty
alr root detection has settled on path: D:\a\test\test
alr is finding 0 GPR project files
alr session state is [OUTSIDE]
alr compiled on [2021-09-16 15:28:36] with GNAT version [Community 2021 (20210519-103)]
platform fingerprint: Windows Bits_64 Msys2
platform properties: MSYS2 WINDOWS NATIVE USER BITS_64
community index required branch: stable-1.1
D:\a\test\test\alire_install\bin\alr.exe -n toolchain
ERROR: Command ["gprbuild", "--version"] exited with code-1 and output:
Error: The process 'D:\a\test\test\alire_install\bin\alr.exe' failed with exit code 1
    at ExecState._setResult (D:\a\_actions\alire-project\setup-alire\v1\node_modules\@actions\exec\lib\toolrunner.js:592:25)
    at ExecState.CheckComplete (D:\a\_actions\alire-project\setup-alire\v1\node_modules\@actions\exec\lib\toolrunner.js:575:18)
    at ChildProcess.<anonymous> (D:\a\_actions\alire-project\setup-alire\v1\node_modules\@actions\exec\lib\toolrunner.js:469:27)
[...]

PR #876, opened 2021-10-04
Do not exit with error on external failed detection

mosteo:
An external command may behave differently in different platforms and we cannot know this in advance, so we should not allow the error to propagate. Instead, the detection fails and the crate remains unavailable.

Partially addresses #864 in that `alr` won't error anymore when the attempted detection fails.


Fabien-Chouteau:
@mosteo can we check if `Location` is a directory here: https://github.com/alire-project/alire/blob/e880cdf5fd75d486a39b40c9bdc949e75a8c96eb/src/alire/alire-externals-from_output.adb#L42 ?

mosteo:
Yes, but I don't think we gain anything wrt this PR, which is more general. 

It could be useful though to work around #864. And my comment in #864 about attempting the detection from another folder does not really make sense as, at the time of using the executable, the invocation would fail anyway. Although I still have to study this Windows behavior in detail.

Fabien-Chouteau (approved):

PR #877, opened 2021-10-06
Alr.Commands.Init: fix .gitignore for bin crates

Fabien-Chouteau:


src/alr/alr-commands-init.adb
```diff
@@ -271,7 +271,11 @@ package body Alr.Commands.Init is
          end if;
 
          Put_Line ("obj/");
-         Put_Line ("lib/");
+         if For_Library then
+            Put_Line ("lib/");
+         else
+            Put_Line ("bin/");
+         end if;
```
mosteo: If we want to be exact, we should use `/lib/` and `/bin/` to avoid ignoring subfolders we haven't created. https://git-scm.com/docs/gitignore#_pattern_format

mosteo (approved):

PR #894, opened 2021-12-08
Alire.Crate_Configuration: Generate Pure and Preelaborate Ada package

Fabien-Chouteau:
So the config package can be withed by any other package.


src/alire/alire-crate_configuration.adb
```diff
@@ -151,6 +151,9 @@ package body Alire.Crate_Configuration is
         (File, "--  Configuration for " & (+Crate) & " generated by Alire");
 
       TIO.Put_Line (File, "package " & (+Crate) & "_Config is");
+      TIO.Put_Line (File, "   pragma Pure;");
+      TIO.Put_Line (File, "   pragma No_Elaboration_Code_All;");
+      TIO.Put_Line (File, "   pragma Preelaborate;");
```
mosteo: I think that `Preelaborate` is implicit in `Pure`, so removing the former might save a warning down the line.

mosteo (changes requested):

src/alire/alire-crate_configuration.adb
Fabien-Chouteau: Indeed :+1:

mosteo (approved):

Issue #888, opened 2021-10-28
[RFC] build profile proposal


Fabien-Chouteau:
The goal here is to provide a simpler default solution for handling compilation flags in Alire crates. And offer a more accessible understanding of compilation options for the vast majority of users.

This feature is intended to be recommended, to encourage a homogeneous development experience, but not mandatory. Developers can still maintain their own set of switches in custom GPR files. They can also control run-time checks in the source code with the pragmas `Suppress`, `Unsuppress`, as well as style checks with the pragma `Style_Checks`.

[edit: s/standard/default]
[edit: Add debug info category]
[edit: take switches from external ADAFLAGS]
[edit: split `build_profiles` and `build_switches`]

## Switches categories:

The switches categories define what are the different aspect of compilation that can be controlled. They are enumeration types.

For instance we can define the following categories and their values:
   - Optimization: `performance`, `size`, `debug`
   - Debug info:  `none`, `all`
   - Run-time checks: `none`, `default`, `overflow`, `all`
   - Run-time contracts: `none`, `all`
   - Compile checks: `none`, `warnings`, `errors`
   - Style checks: `none`, `default`

(See below for the corresponding GNAT switches)

## Profiles:

Profiles are sets of values for the categories. They define the full scope of compilation switches for a given situation.

For instance we can define the following profiles:
   - Release
      - Optimization       => `performance`
      - Debug info => `none`
      - Run-time checks    => `default`
      - Run-time contracts => `none`
      - Compile checks     => `none`
      - Style checks       => `none`
   - Validation
      - Optimization       => `performance`
      - Debug info => `all`
      - Run-time checks    => `all`
      - Run-time contracts => `all`
      - Compile checks     => `errors`
      - Style checks       => `default`
   - Development
      - Optimization           => `debug`
      - Debug info => `all`
      - Run-time checks    => `all`
      - Run-time contracts => `all`
      - Compile checks     => `warnings`
      - Style checks          => `default`

## Default Profile

When working on a crate, the dependency crates are compiled with the `Release` profile, and the root crate is compiled with `Development` profile. It is possible to override the profile of the root crate with an `alr` switch (to be define):
 - `alr build --release`
 - `alr build --validation`

## Customize Profiles

Crates can modify the default switches of profiles. For instance a crate that is written and proved with SPARK can safely disable all run-time checks:

```toml
[build_switches]
release.runtime_checks = "none"
release.contracts = "none"
```

It is also possible to change actual switches:
```toml
[build_switches]
[...]

PR #895, opened 2021-12-09
Build profiles and switches system

Fabien-Chouteau:
Implementation of RFC #888


src/alire/alire-toml_keys.ads
```diff
@@ -9,6 +9,8 @@ package Alire.TOML_Keys with Preelaborate is
    Author         : constant String := "authors";
    Auto_GPR_With  : constant String := "auto-gpr-with";
    Available      : constant String := "available";
+   Build_Profile  : constant String := "build_profile";
+   Build_Switches : constant String := "build_switches";
```
mosteo: I would keep on using dashes as with the other similar keys

mosteo (changes requested):

mosteo:
If I understood the tests correctly, this is the version in which the switches themselves can only be overridden for the own crate, right? 

This is a great addition, thanks Fabien.

Fabien-Chouteau:
> If I understood the tests correctly, this is the version in which the switches themselves can only be overridden for the own crate, right?

Yes indeed. Also the switch to change the default build mode of the root crate is missing (right now it's development).

mosteo (approved):

PR #915, opened 2022-01-15
Fix several typos in comments

jquorning:


src/alire/alire-types.ads
```diff
@@ -7,7 +7,7 @@ with Semantic_Versioning.Extended;
 
 package Alire.Types is
 
-   --  Recopilation of types for convenient use and documentation
+   --  Recompilation of types for convenient use and documentation
```
mosteo: This is actually a mistranslated word from Spanish "recopilación". So I guess a proper word would be "collection".

mosteo (changes requested):

mosteo:
Thanks for the nice sweep, @jquorning.

mosteo (approved):

PR #932, opened 2022-02-04
New property for detected host architecture

### LINKED ISSUE #849: Alire does not attempt to detect the host architecture
We need a new property to complete the platform fingerprint.

We'll also need to tag the binary crates with the intended architecture.

mosteo:
This will allow indexing the proper binaries for different host architectures.

~~Since using this property will require changing the index version, for now I would just leave the PR unmerged and merge it for the next major release.~~

There are other backwards-incompatible changes merged, but that's no problem because the stable index is forward-compatible.

Fixes #849

### DIFF
--- a/src/alire/alire-platforms-common.adb
+++ b/src/alire/alire-platforms-common.adb
@@ -1,0 +1,54 @@
+with AAA.Enum_Tools;
+
+with Alire.OS_Lib.Subprocess;
+
+package body Alire.Platforms.Common is
+
+   Arch_Detected : Boolean       := False;
+   Arch_Value    : Architectures := Architecture_Unknown;
+
+   ---------------------------
+   -- Machine_Hardware_Name --
+   ---------------------------
+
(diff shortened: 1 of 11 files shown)


src/alire/alire-platforms-common.adb
```diff
@@ -0,0 +1,54 @@
...
+      if Arch_Detected then
+         return Arch_Value;
+      end if;
+
+      Detect : declare
+         function Is_Known_Arch is new AAA.Enum_Tools.Is_Valid (Architectures);
+
+         Output : AAA.Strings.Vector;
+
+         --  uname is part of coreutils, so it should be available in
```
Fabien-Chouteau, line 43: At this point msys2 is not mandatory on Windows so it would be better to have an alternative.
Which could be to always return `X86_64`.

Fabien-Chouteau (changes requested):

Fabien-Chouteau (approved):

PR #933, opened 2022-02-04
New config option to disable community index

mosteo:
This way we can ensure that no test in our testsuite is inadvertently using the
community index, as indeed it was happening.

### DIFF
--- a/src/alire/alire-config.ads
+++ b/src/alire/alire-config.ads
@@ -25,14 +25,14 @@ package Alire.Config with Preelaborate is
 
       Editor_Cmd  : constant Config_Key := "editor.cmd";
 
-      Catalog_Autoconfig : constant Config_Key := "catalog.autoconfig";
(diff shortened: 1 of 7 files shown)


src/alire/alire-config.ads
```diff
@@ -25,6 +25,10 @@ package Alire.Config with Preelaborate is
 
       Editor_Cmd  : constant Config_Key := "editor.cmd";
 
+      Catalog_Autoconfig : constant Config_Key := "catalog.autoconfig";
```
Fabien-Chouteau: I don't find the name very meaningful. I would rather have something like `catalog.auto_community`.
Also do we use the word `catalog` a lot? I thought `index` was the right word here.

mosteo: I was trying to go for a catalog being the sum of all your configured indexes, but it's probably too confusing, yeah.

Fabien-Chouteau (approved):

Issue #737, opened 2021-05-18
Several crates per repository (monorepos)


mosteo:
I'm creating this issue to track this idea which keeps popping up so interested parties can track it / chime in with comments.

For the sake of clarity: a [monorepo](https://en.wikipedia.org/wiki/Monorepo) is a repo that contains several crates, usually related to each other. I'll call a *subcrate* each crate within a monorepo.

The issue is that, for convenience or other reasons, some people keep several possibly interrelated crates in a single repository. Publishing each one is currently a tedious affair as this requires zipping each crate independently.

Possibilities that have been discussed/attempted in the past:
1. Split your repositories. People have manifested this is not really desirable ([example](https://github.com/alire-project/alire/pull/349#issuecomment-617566644)). For new projects is still a valid approach.
1. Specialized parent/child crates. This introduces unwanted complexity, and I don't think this is the way anymore. (#108 #200 #349).
1. Add an extra attribute to repo origins so the crate can be located anywhere within a repo, not just at the root. 
   - This would allow to keep our publish strategy.
   - Each crate in a monorepo would be just a plain regular crate with its manifest.
   - As long as crates were depended upon from the same commit (e.g., same version), there isn't even wasting of space.
1. Simplify tar+upload during publishing, so no other changes are required in Alire.
1. Introduce a meta-manifest that points to the crates in the repo, to streamline publishing of several crates at the same time.

This is orthogonal to pins (#721), that can be used freely in local development to use all/part of the crates in the monorepo simultaneously. Pins are not used for publishing.

My personal standing right now is that 

- 3 could (hopefully) be a minor change that would allow simple publishing of crates, even if you don't want to publish all of them simultaneously (think of crates that are not versioned in lockstep), hence simple and flexible. 
- 4 is too much work that is not really needed if 3) works well without unexpected interactions.
   - OTOH there are people that prefer tarballs to repositories.
- 5 is doable after 3 to automate publishing of crates in batch.

Other issues that touched on related aspects: #350

onox:
@mosteo I've a small problem with the pin circularity detection (using af2bead of alire): in the orka crate I get this error:

```
error: Pin circularity detected when adding pin orka_transforms --> orka_simd:
error:    Last manifest in the cycle is /path/to/orka/orka_transforms/alire.toml
```

[...]

PR #939, opened 2022-02-12
Support for publishing of crates in subdirectories of repositories

mosteo:
These are regular crates for all purposes but for the detail that they're not at the root of a repository, so they don't introduce complex issues like in previous proposals for child crates.

The only fancy thing is that to reuse a repository checkout, instead of using the `name_version_hash` naming in the dependencies, we use a `repoBasename_hash` naming for such releases. This is out of sight of the user anyway, as it happens inside `alire/cache/dependencies`.


src/alire/alire-releases.adb
```diff
@@ -126,6 +126,116 @@ package body Alire.Releases is
...
+      if R.Origin.Is_Monorepo then
+         return R.Deployment_Folder / R.Origin.Subdir;
+      else
+         return R.Deployment_Folder;
+      end if;
+   end Base_Folder;
+
+   -------------------
+   -- Unique_Folder --
+   -------------------
```
jquorning:
```suggestion
   -----------------------
   -- Deployment_Folder --
   -----------------------
```

src/alire/alire-publish.adb
```diff
@@ -1018,6 +1093,12 @@ package body Alire.Publish is
            ("URL seems to point to a repository, but no commit was provided.");
       end if;
 
+      --  Verify that subdir is not used with an archive (it is only for VCSs)
+
+      if Subdir /= "" and then Commit = "" then
+         Raise_Checked_Error ("Cannot publish a nested crate from an archive");
```
Fabien-Chouteau, line 1099: Why do we have this limitation?

Fabien-Chouteau (approved):
To be honest I am bit worried about this monorepo support, to me it sounds like the old ways of doing things :man_shrugging: 

I have a couple questions:
 - Can two releases from a monorepo with the same version come on different commits?
 - Can I release a patch fix for a create in a mono repo without making a release for the other crates?

src/alire/alire-publish.adb
mosteo: I wanted to tackle the presumed more wanted feature first. In principle it should be doable too, but I'm not sure it makes sense to publish as a source archive and include folders over the one you want to publish. Maybe to share resources at some point?

mosteo:
> * Can two releases from a monorepo with the same version come on different commits?

This PR is just selecting a subdir inside the checkout and ignoring the rest of it, there is nothing fancy going on with versions. It's point 3 in #737 (that I should have linked in the initial submission but forgot).

So, not sure if this is what you ask, but each crate still has its own version, even if several come from the same commit, and the version needs not to match. Conversely you can have releases for different crates coming from different commits too. Or have different crates with the same version in different commits. 

Or, if you mean: can the code for the same version of the same release be found in two different checkouts of the same repo? Then yes, but their subdirs can't appear simultaneously in the environment, as only one of the two will end selected by the solver for use.

Still, I should add a test for that, if you approve this PR ;-)

> * Can I release a patch fix for a crate in a mono repo without making a release for the other crates?

Yes, every crate is published independently just like before. I just realized that `cargo` has something related: https://doc.rust-lang.org/book/ch14-03-cargo-workspaces.html (that would be point 5 in #737); not sure how that interacts with publishing though, I see no mention of it in `cargo publish`. So I expect it's also orthogonal there.

> To be honest I am bit worried about this monorepo support, to me it sounds like the old ways of doing things man_shrugging

Well, yes, we've seen the arguments before :-) I think it is orthogonal how people want to organize their crates in repositories with the publishing proper; our particular problem is that it's more trouble publishing right now if you share the repo as that forces to upload tarballs manually.

At the end of the day, for us currently is a matter of automation. If we could easily upload tars without having to mess with the public releases of the user, we could dispense with this particular PR (point 4 in #737). But that for some reason makes me even more wary. Create Alire-specific releases in their repo? How to distinguish our artifacts from theirs in the same GitHub release? It would be different if we had dedicated backend infrastructure.

Although on second though maybe indeed having a single `alire releases` release in their repo, with all our tarballs inside, is practical. GitHub says that the only limit is on the individual files size, not on the release proper.

onox:
> To be honest I am bit worried about this monorepo support, to me it sounds like the old ways of doing things man_shrugging

Managing code spread over more than a dozen repos instead of one is just very cumbersome.

> Well, yes, we've seen the arguments before :-) I think it is orthogonal how people want to organize their crates in repositories with the publishing proper; our particular problem is that it's more trouble publishing right now if you share the repo as that forces to upload tarballs manually.

Uploaded tarballs are also quite opaque and it's not easy to verify you get the exact same code as some particular commit (unlike GitHub's .zip download feature).

onox:
@mosteo Is something blocking this PR? I would like to start publishing the wayland-ada crates to the index :slightly_smiling_face:

mosteo:
> @mosteo Is something blocking this PR? I would like to start publishing the wayland-ada crates to the index slightly_smiling_face

@onox I want to include an extra test for a case that raised Fabien. Besides, if you start using this feature your releases will go into the `devel` index so they won't be generally available... Is that OK for you?

onox:
> Is that OK for you?

Sure, there's no rush.

Issue #952, opened 2022-03-23
Build flag 


pjljvandelaar:
In my old gpr file, the "-s" flag 
(recompile if compiler switches have changed since last compilation)
was added to the Builder:
```
 package Builder is
...
end Builder;
```

With alire, I see that the build phase has flags
```
Command ["gprbuild", "-gnatwU", "-j0", "-p", "-P", "tests.gpr"]
```
but not the "-s"flag.

Is this flag not needed?
If so, why?
If not, how to add it?
If I need to make changes to the generated gpr file, do you have recommendations to separate changes from generated code?

Fabien-Chouteau:
Hello @pjljvandelaar 

> Is this flag not needed?

The definition of `-s` for gprbuild is: "recompile if compilation switches have changed"
So this is indeed optional.

Now I see why you want to use this switch so you have two options:
 - 1: You can still have a package Builder with the `-s` switch in your project file:
     ```
   package Builder is
      for Default_Switches ("Ada") use ( "-s");
   end Builder;
    ```
 - 2: You can pass the `-s` to gprbuild option when running `alr build`: `$ alr build -- -s`

For the longer term, @mosteo we should consider always having the `-s` switch when running `gprbuild`.

mosteo:
> For the longer term, @mosteo we should consider always having the `-s` switch when running `gprbuild`.

Sounds like a sensible default

PR #955, opened 2022-03-28
Change default GPRbuild switches

Fabien-Chouteau:
- Remove `-gnatwU` (Recompile if compiler switches have changed):
     It is a legacy from the time when the index was in Ada. Removal may
     trigger some issue on crates with warnings-as-error. But the longer
     we wait the bigger of a problem it will be.

 - Add `-s` (Recompile if compiler switches have changed):
     As discussed in #952, it makes sense to have this default for all
     builds.


mosteo (approved):

mosteo:
Thanks, @Fabien-Chouteau

PR #956, opened 2022-03-28
Alr.Commands.Exec: -P switch to add project file on the command line

Fabien-Chouteau:
- Alr.Commands: Add built-in aliases for gnatcov and gnatprove

### DIFF
--- a/src/alr/alr-commands-exec.ads
+++ b/src/alr/alr-commands-exec.ads
@@ -1,5 +1,7 @@
 with AAA.Strings;
 
+private with GNAT.Strings;
+
 package Alr.Commands.Exec is
 
    type Command is new Commands.Command with private;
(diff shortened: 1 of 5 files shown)


src/alr/alr-commands-exec.ads
```diff
@@ -25,24 +27,40 @@ package Alr.Commands.Exec is
...
+       .New_Line
+       .Append ("The ""-P"" switch can be used to ask Alire to insert a ")
+       .Append ("""-P <PROJECT_FILE>"" switch to the command arguments.")
+       .Append ("""-P"" takes an optional position argument to specify where")
+       .Append ("to insert the extra switch. ""-P1"" means first position, ")
+       .Append ("""-P2"" second position, etc. ""-P-1"" means last position, ")
+       .Append ("""-P-2"" penultimate position, etc. ""-P"" equals ""-P1"".")
+       .Append ("For example ""alr exec -P2 -- python3 main.py arg1"" will")
+       .Append ("run the following command:")
+       .Append ("[""python3"", ""main.py"", ""-P"", ""crate.gpr"", ""arg1""]")
```
mosteo, line 39: I'm not understanding the rationale behind this. What is the problem with placing the `-P` after the `--` wherever you want it, basically.

Fabien-Chouteau: The goal is to be able to provide a cleaner interface for tools that use project file, e.g. the two built-in aliases in the second patch.

At first I was going to create dedicated commands for `gnatprove` and `gnatcov` but I quickly realize that those two commands will just be duplicated code compared to the `exec` command. So I improving `exec` to be able to run tools that require project file.

Since Alire already knows the project file for the crate, it seems suboptimal to ask the user to provide it.
We don't do `alr build -- -P hello.gpr` for instance. I want the same for `gnatprove`, `gnatcov` and other tools in the future.

mosteo: All right, so the trick here is that we want to reuse the manifest project list. I think I understand now, thanks!

mosteo (approved):

PR #964, opened 2022-04-01
`alr get --build`: skip build for projects with auto-gpr-with=false

mosteo:
We warn instead of going ahead and likely failing. The user still can force it.


Fabien-Chouteau:
I don't think `auto-gpr-with=false` will always mean that the build fails.

mosteo:
That's true. I wanted this for testing but there are other ways like defining a test action, so closing it for now.

PR #970, opened 2022-04-05
Update msys2 installer and allow user to set version

Fabien-Chouteau:


mosteo (approved):

mosteo:
I have a suspicion on why that test failed the first time, it may be related to the failure in our backend fixed in https://github.com/alire-project/alire-index/pull/493 . I'll try and narrow it down.

Fabien-Chouteau:
Thanks for checking this @mosteo

PR #972, opened 2022-04-06
Build switches: unconditionally add -gnatW8

Fabien-Chouteau:
CC @reznikmm


mosteo (approved):

mosteo:
I'm in favor of all things Unicode/UTF-8. I do remember however some discussions about this on c.l.a. about things not working out so seamlessly. 

See https://groups.google.com/g/comp.lang.ada/c/l0aflBePm_E/m/BUtW96RhAgAJ and following discussion.

I'm leaving this here before merging for a while just in case, with the idea of merging it for the next release if we found no objection. I admit to having no experience with the issue.

Fabien-Chouteau:
After some further discussion, the risk of introducing incompatibilities is too high to make this the default switch.

reznikmm:
> After some further discussion, the risk of introducing incompatibilities is too high to make this the default switch.

Could you elaborate? What incompatibilities did you find?

I've compiled these two files and got exactly assembler output:

```ada
package Pkg is
   ["03C0"] : Integer;
end Pkg;
```

    gcc -S pkg.ads
    mv pkg.s pkg.s_

```ada
package Pkg is
   π : Integer;
end Pkg;
```

    gcc -S -gnatW8 -gnatwi pkg.ads
    diff -u pkg.s pkg.s_

Fabien-Chouteau:
`-gnatW8` tells GNAT to decode sources as UTF-8 for the sources to compile, but also all their dependencies, so when building crate A, this will also affect the processing of the dependency crate B

mosteo (approved):

PR #975, opened 2022-04-08
Use proper index versioning for backwards compatibility

mosteo:
Currently we are emitting a hard error if either the index branch or version do not match.

With this PR, Alire knows the range of compatible versions (assuming we add features without breaking back compatibility), so now a warning is emitted for old compatible versions and mismatched branches. The error is only produced by a too old index version.

This should result in simpler switching between stable/devel versions, and also be less painful for people with private indexes whenever we add some new feature that does not break compatibility, as they're not forced to update their version on the spot.


src/alire/alire-toml_index.adb
```diff
@@ -101,7 +100,44 @@ package body Alire.TOML_Index is
...
+
       use type Semantic_Versioning.Version;
+
+      ----------------------
+      -- Compare_Branches --
+      ----------------------
+
+      procedure Compare_Branches (Local : String) is
+         Local_Kind    : constant String := AAA.Strings.Head (Local, "-");
+         --  Local_Version : constant Semantic_Versioning.Version :=
```
Fabien-Chouteau: To be removed?

Fabien-Chouteau (approved):

src/alire/alire-toml_index.adb
mosteo: Yup, good catch.

PR #979, opened 2022-04-11
Alire.Utils.Switches: Disable contracts for Development profile

Fabien-Chouteau:
This was probably too aggressive as contracts severely impact performances.


mosteo:
I find this counter-intuitive as I would expect contracts to be checked during development. Might this  become another `-gnato` surprise?

Fabien-Chouteau:
That was my original intent but using this system I realize that contracts can quickly make the code unbearably slow for development.

mosteo (approved):

PR #980, opened 2022-04-11
Reject manifest keys with underscores

mosteo:
These were slipping in before due to improper validation


Fabien-Chouteau (commented):
I feel like this is likely to break many existing projects.

mosteo:
I checked the community index and that one is OK, but I see your point.

We could emit a warning for now and eventually make it an error down the road with some major index version upgrade.

Fabien-Chouteau (approved):
If we are good on the community index I am ok with this.

PR #983, opened 2022-04-12
New `action` subcommand

mosteo:
No need to review yet as tests are still missing.

This PR adds a new subcommand, `alr action`, to allow listing and manual triggering of actions and so help with their development/debugging.

`alr action` simply lists any actions defined.

`alr action post-fetch` and similar trigger the  associated actions.

Adding the `--recursive` flag will list/trigger in dependencies too.

I ditched the idea discussed elsewhere of having a custom action type, since it would likely consist in running an script, and `alr exec` already covers that use case.


Fabien-Chouteau (approved):
BTW, I think we should reconsider the post-fetch action. See https://go.dev/blog/supply-chain

mosteo:
> BTW, I think we should reconsider the post-fetch action. See https://go.dev/blog/supply-chain

Lots of interesting info in that link. But how will we deal with anything requiring a makefile or extra tool? We could have some kind of safe mode where actions cause an error instead of being run.

I find most interesting the bit about not automatically updating and making transitive dependencies explicit.

Fabien-Chouteau:
> Lots of interesting info in that link. But how will we deal with anything requiring a makefile or extra tool? We could have some kind of safe mode where actions cause an error instead of being run.

Or a user query that shows the action that will be executed.

PR #987, opened 2022-04-13
Alire.Crate_Configuration: Fix pragma No_Elaboration_Code

Fabien-Chouteau:
pragma No_Elaboration_Code_All has an impact on implicitly or explicitly with’ed units:
https://docs.adacore.com/gnat_rm-docs/html/gnat_rm/gnat_rm/implementation_defined_pragmas.html


mosteo (approved):

mosteo:
It seems the new pragma is not recognized by all compiler versions. What to do about that?

mosteo (approved):

PR #1007, opened 2022-04-26
Do not allow old-style licenses in strict index mode

mosteo:


Fabien-Chouteau (approved):
Do we have a test for "strict loading"?

mosteo:
> Do we have a test for "strict loading"?

I think so, but I'll double check because my memory...

mosteo:
There's `tests/index/check_enums`. BTW, ~local~ working folder manifests are always loaded in strict mode.

mosteo:
On second thought I don't think this is worth the hassle. After 1.2 is out we can simply remove the old license format and be done with it.

PR #1017, opened 2022-05-04
Bump msys installer in devel branch

mosteo:


mosteo:
These failures are expected as the setup action is using the master branch code which is precisely what we want to fix

PR #1023, opened 2022-05-13
Create zipball with full sources for releases

### LINKED ISSUE #1022: Release source tarball does not include submodules
The source code tarballs at https://github.com/alire-project/alire/releases cannot be built without the sources for the git submodules in the deps directory, but these sources are not included in the tarball. As the releases also exclude the .git directory, there is no way to populate the deps directory based only on the content of a release tarball.

I'm trying to package Alire for Alpine Linux and it would be much easier if the release tarballs had the deps directory populated. Ideally a release tarball should be buildable on any system with a working gprbuild and gnat. My current workaround is to download hardcoded commit hashes from github, which is ugly and tedious to maintain. I don't believe this would be accepted into the Alpine aports repository as-is.

https://github.com/JeremyGrosser/aports/commit/46fdfc38ba89e1d30a92d4dd458d1d4ca5314985

mosteo:
Fixes #1022 

@JeremyGrosser can you check that the file created e.g. at https://github.com/mosteo/alire/releases/tag/v0.0.0 would suit your needs? TIA.

### DIFF
--- a/.github/workflows/tarball-full.yml
+++ b/.github/workflows/tarball-full.yml
@@ -1,0 +1,56 @@
+name: CI linux
+
+on:
+  release:
+    types: [created]
+
+jobs:
+
+  tar:
+    name: ${{ matrix.os }}
+
+    runs-on: ${{ matrix.os }}
+
+    strategy:
+      matrix:
+        os:
+          - macos-latest
+          - ubuntu-latest
+          - windows-latest
+
+    steps:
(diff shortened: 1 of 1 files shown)


JeremyGrosser:
The full-sources zip works for my Alpine build now, thanks!

Unrelated: Should the OS package be called `alire` or `alr`?

Fabien-Chouteau (approved):

PR #1030, opened 2022-05-25
Alire.Utils.Switches: Disable No_Exception_Propagation warning

Fabien-Chouteau:
This switch was always added to the init crates and should have been
included in the first version of build profiles. The warnings are only
displayed for embedded project based on run-times without exception
propagation, and can be very annoying.

### DIFF
--- a/src/alire/alire-utils-switches.adb
+++ b/src/alire/alire-utils-switches.adb
@@ -129,9 +129,11 @@ package body Alire.Utils.Switches is
           when None     => Empty_List,
           when Warnings => Empty_List
                            .Append (GNAT_All_Warnings)
+                           .Append (GNAT_Disable_Warn_No_Exception_Propagation)
                            .Append (GNAT_All_Validity_Checks),
           when Errors   => Empty_List
                            .Append (GNAT_All_Warnings)
+                           .Append (GNAT_Disable_Warn_No_Exception_Propagation)
                            .Append (GNAT_All_Validity_Checks)
                            .Append (GNAT_Warnings_As_Errors),
           when Custom   => S.List);


Fabien-Chouteau:
@mosteo This should go into `1.2`.

JeremyGrosser:
If we're reconsidering style flags, I find the 80 character line limit to be a bit small and would prefer that it either be disabled or set to a higher value by default.

Fabien-Chouteau:
> If we're reconsidering style flags, 

This is not a style flag :wink: 

> I find the 80 character line limit to be a bit small and would prefer that it either be disabled or set to a higher value by default.

I think it's important to define a minimal common coding style for open-source Ada projects, and that's why I added this in the build profile system. Of course I know we can't get everyone to agree on all the details of that coding style. The max line length is obviously one of the oldest debate in that topic.

I am OK to explore different options but I don't want to drag the 1.2 release further, and don't want this to be a showstopper for the build profile feature...

I see a couple of options:
 - Stick with 80 limit. You can override locally with `pragma Style_Check ("gnatyM120")` or add another `-gnatyM` in the GPR file compiler switches.
 - Add a switch category just for the max line length. A minor issue here is that switch configurations are supposed to be enumerations, so it will look like this (`M80`, `M100`, `M120`, `Custom`)
 - Don't use any max line length style checks at all

If we decide to remove the limit, it will be impossible add any limit in the future because that will break existing crates. 

@mosteo what's your opinion on this?

mosteo (approved):

mosteo:
Sorry, I've been traveling the last days.

Given the inevitable friction to come from style checks, I'd like to have some way of providing my own preferred defaults somehow (config settings, templates for the user gpr...). For line length, 80 or 100 is OK for me. More complicates diffs.

As long as Alire style settings in the autogenerated project file can be overriden by hand in the user project file, I don't see as very urgent what the particular override mechanism could be.

> If we decide to remove the limit, it will be impossible add any limit in the future because that will break existing crates.

Isn't this a per-project setting? So adding a limit would only affect newly created crates?

Fabien-Chouteau:
Hi @JeremyGrosser 

I just checked for a solution to your issue. You can simply add your preferred max line length limit in the root gpr file of your project in addition to the switches provided by Alire. For instance:
```ada
   package Compiler is
      for Default_Switches ("Ada") use My_Crate_Config.Ada_Compiler_Switches & ("-gnatyM120");
   end Compiler;
```
This is a simple one time change that you can use to adjust the style checks to your preference.

I think the best solution right now as I would like to preserve a max line length in the default coding style, and the adding a new switch category is too heavy.

CC @mosteo

mosteo:
Merging this one now, as it is not entirely related to the discussion. Please open a new issue/discussion if you thing the topic merits further debate. Thanks to all involved.

PR #1042, opened 2022-06-05
Reduce gprbuild startup time by using its --autoconf parameter

BeChris:
Hello,
Everytime gprbuild is executed it has to parse its gprconfig database which could be slow especially on slow filesystems.
I propose to provide --autoconf= parameter to gprbuild so it parses its databse only once : for further executions a cache of this database is used.
This cache is written into alire subdirectory and deleted with an alr clean command.

Execution times on my PC running Arch Linux with an SSD:
Before optimisation:
alr build
ⓘ Building hello/hello.gpr...                    
gprbuild: "hello" up to date
Build finished successfully in 1.27 seconds.

After optimisation:
alr build
ⓘ Building hello/hello.gpr...                    
gprbuild: "hello" up to date
Build finished successfully in 0.42 seconds.

### DIFF
--- a/src/alire/alire-paths.ads
+++ b/src/alire/alire-paths.ads
@@ -5,6 +5,9 @@ package Alire.Paths with Preelaborate is
    Crate_File_Name : constant String := "alire.toml";
    --  Name of the manifest file in a regular workspace
 
+   Cache_Project_File_Name : constant String := "build.cgpr";
+   --  Name of the "cache" file used by gprbuild
+
    Cache_Folder_Inside_Working_Folder : constant Relative_Path := "cache";
 
    Deps_Folder_Inside_Cache_Folder : constant Relative_Path := "dependencies";
--- a/src/alire/alire-spawn.adb
+++ b/src/alire/alire-spawn.adb
@@ -1,6 +1,7 @@
 with Alire_Early_Elaboration;
+with Ada.Directories;
+with Alire.Config;
 with Alire.Paths;
-with Alire.Root;
 with Alire.OS_Lib.Subprocess;
 
 package body Alire.Spawn is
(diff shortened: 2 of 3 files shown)


mosteo:
Thanks for the proposal, @BeChris.

@Fabien-Chouteau, this seems sensible but it goes beyond my experience with gprbuild. I've read a bit through https://docs.adacore.com/gprbuild-docs/html/gprbuild_ug/companion_tools.html and it seems it may touch on cross-compiling. Do you think this is safe in all cases?

Fabien-Chouteau:
I checked with the colleagues and there might be silent issues when changing the target or run-time.
Not something users will do often, but the consequences can be hard to understand...

BeChris:
I agree with @Fabien-Chouteau : with my colleagues we experimented this option and found a potential source of problems:
upon the cache generated by gprbuild, if the toolchain and/or the Ada runtime (zfp, ravenscar, ...) are changed then gprbuild will still refer to its cache and won't take into account the new toolchain and/or Ada runtime.
This obviously lead to strange behaviour ... until the cache is manualy deleted (it will be generated again on next gprbuild run).
But as we compile very often our whole project or individual units we appreciated the gain of speed and kept the --autoconf option activated.

BeChris:
As it is error prone if user change the toolchain after gprbuild has generated its cache, commit feaaadc generates the cache filename based on toolchain one.
Therefore swithing between toolchains is no more a problem.

Fabien-Chouteau:
Hello @BeChris ,

Unfortunately `Alire.Config.Keys.Toolchain_Use` is not enough to detect compiler target and run-time.
In fact Alire is not aware of neither of the target nor the run-time specified in the project file so we are not able to detect changes here.

The only workaround for now is too pass `--autoconf` on the Alire command line:
```console
$ alr build -- --autoconf=alire/autoconf.gpr
```
You can even define an alias to make it easier:
```console
$ alr config --set --global alias.autobuild  "build -- --autoconf=alire/autoconf.gpr"
$ alr autobuild
```

BeChris:
Hello @Fabien-Chouteau,
considering the explanations you gave and other corner cases which may also arise,
(I noticed another problem where the native toolchain is considered whereas a crosscompiler is referenced : like in rp2040_hal crate) 
I think the solution you provided to use a command alias is safer than my proposed implementation.

I propose to not integrate and close this pull request.

mosteo:
OK, @BeChris, thanks for the idea nonetheless. I guess being aware of this possibility may be helpful to some people, and with the aliases it is simple to apply.

PR #1046, opened 2022-06-08
Update dependencies to indexed versions

mosteo:
With this, the only dependency requiring a pin is `gnatcoll`. This is because we are using a custom stripped-down gnatcoll that does not depend on `libgpr` and `xmlada`. I guess at this point this is obstination on my part since this way we are missing any new features/fixes on gnatcoll, and at some point I will need to accept that these unused dependencies are worth using a more recent gnatcoll.

If we were to move to the regular gnatcoll, we could index `alr` within `alr`, which would be neat.


mosteo:
[It seems](https://github.com/alire-project/alire/runs/6792324675?check_suite_focus=true) `xmlada` has the same problem with its configure.sh not working on msys2. @Fabien-Chouteau, should I report this somewhere?

Fabien-Chouteau:
> [It seems](https://github.com/alire-project/alire/runs/6792324675?check_suite_focus=true) `xmlada` has the same problem with its configure.sh not working on msys2. @Fabien-Chouteau, should I report this somewhere?

It's already fixed in the development branch, there's not much we can do for previous releases...

Fabien-Chouteau (approved):

mosteo:
OK, so it's a matter of time until it makes into a release.

PR #1047, opened 2022-06-08
Alire.Crate.Configuration: disable style checks in generated code

Fabien-Chouteau:
With the current style checks it is easy to break coding style by having
to many values in an enum for instance.

In theory we could generate better code for enums. But since users can
use their own coding style, it is impossible to be compatible with in
every situation.

So the best option is to disable all checks for this file.


mosteo (approved):

Fabien-Chouteau:
@mosteo something we can push to the `1.2` branch as it can be an annoying situation to face. Speaking from experience ^^

mosteo:
OK, I'll apply it and we can push out a 1.2.1

Fabien-Chouteau:
Not sure it's worth a .1 yet.
BTW we should also add the GNAT Studio message change.

mosteo:
Yeah, we can wait for a while so things consolidate a bit. I also had the GNAT Studio message in mind, thanks for the reminder.

PR #1049, opened 2022-06-08
Keep `last_build_profile` in sync with profile requested in manifest

### LINKED ISSUE #1040: `alr config` return wrong build profile when override is in manifest
Having e.g.

```
[build-profiles]
"*" = validation
```

will properly apply the validation profile (verifiable inspecting `config/crate_config.gpr`), but `alr config` will show

`last_build_profile=DEVELOPMENT`

Put another way, it seems that the configuration stored is only for the switch used with `alr build`.

mosteo:
Fixes #1040

### DIFF
--- a/testsuite/tests/build_profile/last_profile/test.py
+++ b/testsuite/tests/build_profile/last_profile/test.py
@@ -1,0 +1,38 @@
+"""
+Check that the last build profile is stored properly, no matter its procedence
+"""
+
+from drivers.alr import run_alr, init_local_crate, alr_manifest
+from drivers.asserts import assert_eq, assert_match
+
+init_local_crate("xxx")
+
+# Check profile is the default one (development) if unspecified
+
+run_alr("build")
+assert_match(".*last_build_profile=DEVELOPMENT.*",
+             run_alr("config").out)
+
+# Check explicit profile in command line
+
+run_alr("build", "--release")
+assert_match(".*last_build_profile=RELEASE.*",
+             run_alr("config").out)
+
+# Check implicit profile when build is indirect
+
+run_alr("run")  # Causes a default build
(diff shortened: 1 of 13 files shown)


testsuite/tests/build_profile/last_profile/test.py
```diff
@@ -0,0 +1,38 @@
...
+
+# Check explicit profile in command line
+
+run_alr("build", "--release")
+assert_match(".*last_build_profile=RELEASE.*",
+             run_alr("config").out)
+
+# Check implicit profile when build is indirect
+
+run_alr("run")  # Causes a default build
```
Fabien-Chouteau, line 24: I wonder if `alr run` should keep the last build profile.

mosteo: Yep, it's probably more intuitive that way. I'll add that change.

mosteo:
I'm stuck trying to fix the errors in the testsuite. For starters, it seems to depend on the compiler version: I don't get those with Ubuntu's 20 compiler (9.x) but I do with 11.2. And then, when I run the test steps by hand, I don't get the exception either. 

I guess something in the test environment is having something to do, but event trying to set up it by hand in advance is not working.

Glad for any ideas on what to try.

mosteo:
So after some trial and error, the only "simple" solution I could find was to manually manage and trigger the finalization of the new `Configuration` field in the `Root` record. Since this is functionally equivalent, I find it all very suspect.

testsuite/tests/build_profile/last_profile/test.py
mosteo: But in a follow-up patch

Fabien-Chouteau (approved):

PR #1057, opened 2022-06-10
Alire.Utils.Switches: add -gnateE when debug info enabled

Fabien-Chouteau:
-gnateE adds extra info in exception messages.


mosteo:
Pinging: is this purposely in draft mode?

Fabien-Chouteau:
I was not sure of the impact. It should be good to go.

mosteo (approved):

PR #1080, opened 2022-07-05
Use last build profile on `alr run`

mosteo:
Instead of always defaulting to development.

I wonder if `alr build` without an explicit profile should also default to the last explicit profile.

Edit: this PR includes a bit of refactoring to get rid of a global variable (`Alire.Crate_Configuration.Root_Build_Profile`) that was making the logic start to get convoluted with the recent changes.


mosteo:
@Fabien-Chouteau this one fell through the cracks?

Fabien-Chouteau (approved):

PR #1085, opened 2022-07-09
Lazy index loading (1/2)

### LINKED ISSUE #381: Load index on demand
The scattered `Requires_Full_Index` calls where necessary in the ancient self-compiling `alr`. Nowadays we could load the index on demand when it is first queried.

### LINKED ISSUE #910: Alire checks manifest files of all releases in index for no reason
When running commands like `alr clean` or `alr run -s` (probably other commands as well), Alire seems to parse every single .toml file in the index. This adds like over half a second of startup latency to `alr run`. Why are these files parsed at all?

mosteo:
This is the first of two PRs that remove the need to load the full index for many operations.

The idea is that we know what indexes are configured, where they are on disk, and where each crate lives therein, so any time a crate is needed, we can load it at that time. All in all, this should speed-up things when indexes grow, and fewer external detections should be attempted.

There is a catch that makes this not as straightforward as it seems, which is that "provides" fields mean that a provided crate may require loading whichever other crates provide it, and this is unknown in advance until the whole index has been loaded at least once.

This part is what goes into the follow-up PR: the first time we "see" an index, it is loaded in full and `provider -> provided` mappings are stored in a new metadata file that is invalidated whenever indexes are added or updated. From then on, we know all the crates that must be loaded when one is requested, and lazy load works as intended in all cases.

Fixes #910, fixes #381

### DIFF
--- a/src/alire/alire-externals-from_output.adb
+++ b/src/alire/alire-externals-from_output.adb
@@ -47,7 +47,7 @@ package body Alire.Externals.From_Output is
 
       declare
          use GNAT.Regpat;
-         Matches : Match_Array (1 .. Match_Count'Last);
+         Matches : Match_Array (1 .. 1); -- We should have at most one match
          Lines   : AAA.Strings.Vector;
          Status  : constant Integer :=
                      OS_Lib.Subprocess.Unchecked_Spawn_And_Capture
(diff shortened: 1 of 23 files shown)


Fabien-Chouteau:
> new metadata file that is invalidated whenever indexes are added or updated

What if I add a new crate in my local index? Will this be detected?

 (I didn't look at the PR content yet)

Fabien-Chouteau (approved):

mosteo:
> What if I add a new crate in my local index? Will this be detected?

You're right, this won't be detected, although running `alr index --update-all` is all it takes to fix it. I don't see any simple way of detecting this that doesn't involve complex filesystem monitoring.

I will add a note about it in the publishing documentation.

PR #1099, opened 2022-07-22
Fix solver continuing after user abandoned

mosteo:
The solving was not being totally abandoned, but if there were more exhaustive searches to perform, it would counter-intuitively continue.

A problem remains in that answering 'Always continue' will still re-ask at each exhaustiveness level. (Separate PR).


mosteo:
Superseded by incoming PR, sorry about the noise.

PR #1127, opened 2022-08-01
Reason for missing releases in solution

mosteo:
This is an internal improvement to the solver so now when a dependency can't be solved it will give the reason (via an extra `:reason` part after the usual `missed`). For example, in the output of `alr with --solve`:

```
Dependencies (external):
   hello(=1.0) & (=1.0.1) (direct,missed:conflict)
   libhello^777 (direct,missed:unavailable)
   unobtanium* (direct,missed:unindexed)
Dependencies (graph):
   some_dep=0.1.0-dev --> hello=1.0.1 
   xxx=0.1.0-dev      --> hello=1.0    
   xxx=0.1.0-dev      --> libhello^777
   xxx=0.1.0-dev      --> unobtanium*
```

- Unindexed: the crate doesn't exist at all in our indexes
- Unavailable: the requested version does not exist (but the crate does)
- Conflict: two dependencies on the same crate cannot be reconciled. This can be further diagnosed in the graph portion of the solution (no change on the info shown there).

### DIFF
--- a/src/alire/alire-dependencies-states.adb
+++ b/src/alire/alire-dependencies-states.adb
@@ -56,6 +56,7 @@ package body Alire.Dependencies.States is
       Link         : constant String := "link";
       Pin_Version  : constant String := "pin_version";
       Pinned       : constant String := "pinned";
+      Reason       : constant String := "reason";
       Release      : constant String := "release";
       Shared       : constant String := "shared";
       Transitivity : constant String := "transitivity";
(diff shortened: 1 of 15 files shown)


Fabien-Chouteau (approved):
Looks good to me except for the word "unindexed". To me it feels like the crate is available somewhere but not indexed, which might not be the case. Maybe "unknown" would be better.

mosteo:
I see your point. Will change it.

PR #1128, opened 2022-08-02
Enhancements to actions documentation/testing

mosteo:


doc/catalog-format-spec.md
```diff
@@ -403,6 +410,20 @@ static, i.e. they cannot depend on the context.
    # An explicit empty case alternative, which is not mandatory
    ```
 
+   The aforementioned TOML syntax is only when there is only one conditional
```
Fabien-Chouteau: "syntax is only when there is only" seems like a word is missing.

Fabien-Chouteau (changes requested):

mosteo:
Requested changes have been applied. Please reopen if there are further issues.

Issue #1083, opened 2022-07-09
alr publish refuses to build if a crate with same name/version exists in the loaded index


stcarrez:
I have a local index where I populate several project crates before publishing widely.
```
alr index
# NAME      URL                                                         PATH                                              
1 community git+https://github.com/alire-project/alire-index#stable-1.2 /home/ciceron/.config/alire/indexes/community/repo
2 local     file:///src/awa/alire-index-local                           /src/awa/alire-index-local
```
At some time in the publish process, it is necessary to re-publish to fix issues but when you do so, you'll get the error:
```
alr publish
✓ Local repository is clean.
error: Target release elada=1.8.4 already exist in a loaded index
```
and I have to remove the crate from my local repository to be able to run the command.

The `-f` (force) option does nothing.
May be Alire should:

* honor the `-f` option and ignore that error,
* don't report an error if the crate+version is found in a local repository

stcarrez:
I've managed to hack alire and make the `-f` option being taken into account.

It works like a charm :-)

mosteo:
Thanks for the report, Stephane. Probably in this case we should check that the release doesn't exist taking into account only the community index. This is not a trivial change though with our current loading infrastructure.

As for making this an overridable error with `--force`, it can certainly be an interim solution.

stcarrez:
The `--force` option works for that purpose so there is no need to keep this opened.

PR #1130, opened 2022-08-03
Honor the --force option when publishing a crate (#1083)

stcarrez:
When publishing and checking for existing crate with same version,
if the crate already exist and the force option is used,
emit a warning message and continue.

### DIFF
--- a/src/alire/alire-publish.adb
+++ b/src/alire/alire-publish.adb
@@ -187,15 +187,9 @@ package body Alire.Publish is
 
       Index_On_Disk.Loading.Load_All (Strict => True).Assert;
       if Index.Exists (Release.Name, Release.Version) then
-         if not Force then
-            Raise_Checked_Error
-               ("Target release " & Release.Milestone.TTY_Image
-                & " already exist in a loaded index");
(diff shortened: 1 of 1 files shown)


src/alire/alire-publish.adb
```diff
@@ -187,9 +187,15 @@ package body Alire.Publish is
...
-            & " already exist in a loaded index");
+         if not Force then
+            Raise_Checked_Error
+               ("Target release " & Release.Milestone.TTY_Image
+                & " already exist in a loaded index");
+         end if;
+         Ada.Text_IO.New_Line;
+         Trace.Warning
+            ("Target release " & Release.Milestone.TTY_Image
+             & " already exist in a loaded index (ignored)");
```
mosteo: Thanks for the PR, @stcarrez. You can better use here `Alire.Recoverable_Error`, which does essentially the same with some other minor tweaks and is a drop-in replacement for `Raise_Checked_Error`.

mosteo (changes requested):

mosteo (approved):

PR #1132, opened 2022-08-04
Emit error for all unset configuration variables

jquorning:
Improve user experience by emitting error for each unset configuration variable.

Output of `alr get lvgl_ada`:
```
error: Configuration variable 'lvgl_ada.density_per_inch' not set and has no default value.
error: Configuration variable 'lvgl_ada.pixel_bit_depth' not set and has no default value.
error: Configuration variable 'lvgl_ada.horizontal_resolution' not set and has no default value.
error: Configuration variable 'lvgl_ada.vertical_resolution' not set and has no default value.
error: Configuration failed
```
Before:
```
error: Configuration variable 'lvgl_ada.density_per_inch not set and has no default value.
```


Fabien-Chouteau (approved):
Looks good to me. Thanks @jquorning

Fabien-Chouteau:
@jquorning you might want to add a test case for that.

mosteo (approved):

mosteo:
I was bitten by this very same case once. Thanks!

mosteo:
Minor last thing: I would move the test from the `get` to the `config` dir, as it applies outside of the get command.

mosteo:
Thanks, merged.

Issue #22, opened 2018-03-26
freeBSD


mosteo:

stcarrez:
The biggest issue today is the availability of a recent compiler and recent gprbuild.
A user must first build and install a recent gprbuild (it works).
Then, the gcc6-aux will fail to build alr because several Ada 2012 pre-conditions are not recognized by gcc6.

mosteo:
Do you think it would be easy/useful to "downgrade" alr to gcc6? Or will freeBSD upgrade to gcc7+ in the near term?

stcarrez:
I have no visibility of Ada support on FreeBSD. The gcc6-aux is still a separate compiler and package installation. I don't forecast any change soon but I may be wrong.

If we can downgrade easily, I would suggest to make some FreeBSD branch (or whatever name) so that the mainstream keeps using the Ada 2012 features.

mosteo:
Given how shorthanded we are right now, perhaps the idea of a FreeBSD branch could be undertaken by some Adaist actively involved in/using FreeBSD

stcarrez:
FYI, gcc6-aux has been removed in Feb 2022 from FreeBSD ports.

Now Alire builds with gcc 12.1 on FreeBSD 13.1 if you manage to have the gcc 12 compiler running on it.

The OS detection is lacking support for FreeBSD, in particular the `case(os)` condition cannot be setup for `freebsd` which is annoying. I can have a look at this.

Now, the biggest issue is not technical but if we introduce a new OS and we start to use it in crates, then the alire repository will have to change its version I think.  In other words, how can we achieve backward compatibility if we add a new OS and want to publish a crate that defines conditions on it?

mosteo:
Great news about the GCC update.

Actually, unknown enum values are allowed in crates from an index, just for this use case. But still, publishing will be inconvenient as the checks are strict about not accepting unknown values, and so a new alr release is needed in order to be able to publish these crates normally.

We could be proactive and include plausible near-future variable values in these enums anyway.

And `1.2.1` is not far away...

mosteo:
Merged in #1148

PR #1143, opened 2022-08-18
Add FreeBSD as a known OS

mosteo:
Since it seems `alr` is working on it with the new gcc, and in preparation for future contributions (https://github.com/alire-project/alire/issues/22#issuecomment-1207224025)

This is the bare minimum to be able to use it in manifests. We still don't have OS-specific files to make the detection.


stcarrez:
Ah!, I've not seen your pull request! When I tried, it was not enough to be usable on FreeBSD.

See https://github.com/alire-project/alire/pull/1148

mosteo:
Yup, yours is much more exhaustive.

mosteo:
Superseded by #1148

PR #1148, opened 2022-08-19
Add support for FreeBSD (#22)

stcarrez:
* update GNAT project to recognize -XOS=freebsd and configure accordingly
* add FreeBSD as a valid operating system
* add AMD64 as an alias for X86_64 because uname -m returns amd64 on FreeBSD
* add platform specific implementation for FreeBSD

To build on FreeBSD, we should use

gprbuild -j0 -P alr_env -XOS=freebsd

### DIFF
--- a/src/alire/alire-platforms.ads
+++ b/src/alire/alire-platforms.ads
@@ -3,8 +3,8 @@ package Alire.Platforms with Preelaborate is
    --  Platform information necessary for some releases
 
    type Extended_Architectures is
-     (ARM64, -- Equivalent to AARCH64
-      AMD64, -- Equivalent to X86_64 (FreeBSD)
+     (AMD64, -- Equivalent to X86_64 (FreeBSD)
+      ARM64, -- Equivalent to AARCH64
       End_Of_Duplicates,
       --  Up to this point, these are architectures that we want to rename to
       --  some of the following because they are equivalent.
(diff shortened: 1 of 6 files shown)


src/alire/alire-platforms.ads
```diff
@@ -76,6 +78,7 @@ private
    is (case Arch is
           when ARM64         => AARCH64,
           when Architectures => Arch,
+          when AMD64         => X86_64,
```
Fabien-Chouteau: You also have to add this conversion in the testsuite: https://github.com/alire-project/alire/blob/master/testsuite/drivers/helpers.py#L106

Fabien-Chouteau (commented):
Thanks @stcarrez

src/alire/alire-platforms.ads
mosteo: Then I'd kindly request from @stcarrez to add a comment here pointing to the link given by Fabien.

mosteo: Just for my OCD, I'd also prefer if these values are in alphabetical order.

src/alire/alire-platforms.ads
```diff
@@ -21,6 +22,7 @@ package Alire.Platforms with Preelaborate is
 
    type Operating_Systems is (Linux,
                               MacOS,
+                              FreeBSD,
                               Windows,
                               OS_Unknown);
```
mosteo: Here too I'd prefer to have FreeBSD first in alphabetical order.

mosteo (changes requested):
Thanks @stcarrez, just some trivial tweaks on my side.

mosteo (approved):

PR #1154, opened 2022-08-22
Migrate workflows with explicit ubuntu-18.04 to latest

mosteo:
No reason AFAICS for these to rely on a particular version.


mosteo:
I'm merging this one so some failing checks can be carried out. Please comment if there's any issue with it.

Issue #1155, opened 2022-08-23
Generate config file for use in shell-scripts or Makefiles


RREE:
I propose to have a 4th kind of generated config file for the use in shell scripts and Makefiles. In many projects the build of an Ada program is only one of several steps. The subsequent steps have to have the possibility to refer to some of configuration values of the Ada project.

I created a patch to Alire ([generate_shell.zip](https://github.com/alire-project/alire/files/9410468/generate_shell.zip) ) that first has to be activated ( `[configuration] generate_shell = true`) and the generated file can then be read by the post-build actions.

Fabien-Chouteau:
That sounds like a good idea. Can you open a pull-request with the changes?

RREE:
As explained in the PR I don't think anymore it is a very useful feature.

Perhaps it would be more useful if `alr config --get/--list` showed the entries of `[configuration]` and `[configuration.values]`, too. Shell scripts can then query alr for the needed parameters.

mosteo:
Sounds sensible, @RREE. If eventually you have a use case please describe it here to help shaping the feature better.

Fabien-Chouteau:
> Perhaps it would be more useful if `alr config --get/--list` showed the entries of `[configuration]` and `[configuration.values]`, too. Shell scripts can then query alr for the needed parameters.

We have this bad issue that `alr config` and `[configuration]` (a.k.a. crate configuration) are two different things...

RREE:
The use case at thand is the following:
I port the examples one by one from [AVR-Ada](https://sourceforge.net/projects/avr-ada/) to the new [AVRAda_Examples](https://github.com/RREE/AVRAda_Examples) project. 

After generating the elf binary (with `alr build/exec`) you have to convert it to a hex file and eep (for the EEPROM part) with `avr-objcopy` before you can upload it to the device with `avrdude`. I am used to achieve that with generic rules in a Makefile. At least `avrdude` must know the name of the actual chip. The MCU name is stored in `alire.toml` as a configuration value `avrada_rts.AVR_MCU`.  I currently extract that value with the help of `awk`, it would be much more elegant if I could read it as output of `alr config --get avrada_rts.avr_mcu`. 

Most sample apps do not work on all AVR MCUs. Even the simple Uart_Echo requires an internal U(S)ART which is missing on the attiny85, once popular as Digispark. Here it would help if each example could set the configuration value in `alire.toml` with a command like `alr config --set avrada_rts.avr_mcu atmega328p`

[...]

PR #1156, opened 2022-08-24
generate config file for use in shells and Makefiles

RREE:
A pull request corresponding to issue #1155


RREE:
I am not sure anymore that this feature is useful. 

You typically want to access the config values from a crate that you depend on. The Ada and gpr config files are found since their search paths are expanded by Alire. There is no corresponding feature for shell scripts or Makefiles. It will be very difficult for a script to find out where the shell config file is located.

Fabien-Chouteau:
> You typically want to access the config values from a crate that you depend on. The Ada and gpr config files are found since their search paths are expanded by Alire. There is no corresponding feature for shell scripts or Makefiles. It will be very difficult for a script to find out where the shell config file is located.

That's true, so the config will only be usable within a crate really.

mosteo:
I'm closing this one as my understanding from #1155 is that it's no longer wanted. Also, is there much difference with doing

`alr printenv > script.sh`

?

PR #1158, opened 2022-08-25
Fix Ada VSC plugin for Gitpod

### LINKED ISSUE #1133: Gitpod ada extension deprecated
I think the Ada extension version mentioned in .gitpod.yml needs a bump.

If I open Alire in gitpod, I see under 'problems': user uploaded extensions are deprecated, and if I click on that, it takes me to .gitpod.yml with a strike through the version hash part:

adacore.ada@22.0.5 _:b56gZDvQ5KBm6u34IBzm1g==_

mosteo:
Fixes #1133

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -3,4 +3,4 @@ image:
 
 vscode:
   extensions:
-    - adacore.ada@22.0.5:b56gZDvQ5KBm6u34IBzm1g==
+    - AdaCore.ada


mosteo:
This is a trivial change that doesn't affect Alire itself and that I already tested, so I'm merging it directly. The obsolete version was causing the plugin to fail to load.

PR #1167, opened 2022-08-30
Allow missing configuration values until build

### LINKED ISSUE #1125: Unset configuration variables in CI
Some crates fail CI with errors like:
`Configuration variable 'cortex_m.core not set and has no default value.`

Should `alr` have a way to auto-initialise these variables for CI?

At the moment the crates failing to configure are:
```
cortex_m
lvgl_ada
lvgl_ada_simulator
lvgl_ada_examples
minisamd51_example
minisamd51_bsp
microbit_examples
microbit_bsp
nrf5x_hal
```

mosteo:
This way we can normally get/without crates without defaults, but at the time of building the values must be set.

Fixes #1125

### DIFF
--- a/src/alire/alire-crate_configuration.adb
+++ b/src/alire/alire-crate_configuration.adb
@@ -31,6 +31,8 @@ package body Alire.Crate_Configuration is
    Profile_Assign : constant Character := '=';
    Profile_Split  : constant Character := ',';
 
+   Must_Be_Set : constant UString := +"(variable without default still unset)";
+
    function Builtin_Build_Profile is new Typedef_From_Enum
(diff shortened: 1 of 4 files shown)


Fabien-Chouteau (commented):
The patch looks ok but I am not sure about the solution used here.

In particular I don't like that the config files will be generated but only partially.
Could we skip entirely the config generation on the get step?

mosteo:
Changed to avoid generating files in case of incompleteness. Also renamed a `Map` to `Var_Map` to be a bit more meaningful.

Fabien-Chouteau (approved):

PR #1173, opened 2022-09-06
Support openSUSE and SUSE distributions

serock:
The openSUSE and SUSE distribution use the zypper package manager. Support for zypper is included in this pull request.

### DIFF
--- a/src/alire/alire-origins-deployers-system-zypper.adb
+++ b/src/alire/alire-origins-deployers-system-zypper.adb
@@ -1,0 +1,117 @@
+with AAA.Strings; use AAA.Strings;
+
+with Alire.OS_Lib.Subprocess;
+with Alire.Errors;
+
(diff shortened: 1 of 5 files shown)


mosteo:
Thanks, @serock, this is a nice addition. 

If you have the time, you could also submit a docker image for SUSE testing, similar to the rest we have at https://github.com/alire-project/dockerfiles

These images must contain the default Ada compiler for SUSE, if there's one, and would be used to test the building of `alr` itself on SUSE. 

If there's no Ada compiler, then still we could use an image with the basic tools used by `alr`, to test submissions to the index, in particular SUSE-specific libraries.

If you can't do it, opening an issue about this would also be useful so we eventually get to it.

Fabien-Chouteau (approved):
Thanks @serock !

It would be great to enable Suse in our testing.

Fabien-Chouteau:
Awesome!

PR #1180, opened 2022-09-07
Fix grouping commands and simplify descriptions

### LINKED ISSUE #1138: Improvements to grouping of commands
I've been thinking about the grouping of the commands displayed by `alr help` a bit, and I think it could use some minor changes. First of all, for some of the commands it's not clear they aren't meant to be used by regular users. Second, the description of commands like 'with' and 'update' is a bit vague and confusing to new users. Third, I had put the 'publish' command in its own group so that in the future new commands like 'login' and 'logout' could be added to this group (this is what several other package managers have), but I think this isn't really applicable to Alire because users follow a different process involving GitHub. Thus the 'publish' command is all that's needed and there's no need to put it in a separate group.

So I propose to rename the 'Build' group to 'Crate' to indicate to the user that these are the commands that the user is likely to need when interacting with the crate. Some of the commands from 'General' and 'Index' have been moved to the 'Crate' group. Some of the commands are moved to a new group 'Internal' (name can be changed).
And the last change is that I have modified the description of the 'with' and 'update' commands.
For consistency, the first word of 'help', 'version', 'get', and 'init' commands have been modified to use the imperative mood ('Show' instead of 'Shows').

This will then be the new output:

```
   General
   help          Show help on the given command/topic
   config        List, Get, Set or Unset configuration options
   toolchain     Manage Alire-provided toolchains
   version       Show detailed version, configuration, and environment information

   Index
   get           Fetch a crate release
   index         Manage indexes used by current configuration
   search        Search a string in release names and properties
   show          See information about a release

   Crate
   build         GPRbuild current working release
   clean         GPRclean working release and manage cached releases
   edit          Start GNATstudio with Alire build environment setup
   exec          Run the given command in the alire project context
   init          Create a new working release with alire metadata, or generate metadata
   pin           Pin dependencies to exact versions
   printenv      Print the build environment variables
   publish       Help with the publication of a new release
   run           Launch an executable built by the release
   update        Update some or all dependencies to a newer version
   with          Add or remove dependencies

   Internal
   action        List or manually trigger action hooks
   dev           Developer helpers
   test          Tests the compilation of all or some releases
```

If agreed, I can submit a PR.

onox:
Fixes #1138

### DIFF
--- a/src/alr/alr-commands-show.ads
+++ b/src/alr/alr-commands-show.ads
@@ -21,7 +21,7 @@ package Alr.Commands.Show is
       Config : in out CLIC.Subcommand.Switches_Configuration);
 
    overriding function Short_Description (Cmd : Command) return String is
-     ("See information about a release");
+     ("Display information about a crate");
 
    overriding function Usage_Custom_Parameters (Cmd : Command) return String is
      ("[<crate>[allowed versions]] [--system] [--external[-detect]]"
--- a/src/alr/alr-commands-show.ads
+++ b/src/alr/alr-commands-show.ads
@@ -21,7 +21,7 @@ package Alr.Commands.Show is
       Config : in out CLIC.Subcommand.Switches_Configuration);
 
    overriding function Short_Description (Cmd : Command) return String is
-     ("Display information about a crate");
+     ("Display information about a crate version");
 
    overriding function Usage_Custom_Parameters (Cmd : Command) return String is
      ("[<crate>[allowed versions]] [--system] [--external[-detect]]"
--- a/src/alr/alr-commands.adb
+++ b/src/alr/alr-commands.adb
@@ -581,9 +581,9 @@ begin
 
    Sub_Cmd.Register ("Publish", new Publish.Command);
 
-   Sub_Cmd.Register ("Internal", new Action.Command);
-   Sub_Cmd.Register ("Internal", new Dev.Command);
-   Sub_Cmd.Register ("Internal", new Test.Command);
+   Sub_Cmd.Register ("Testing", new Action.Command);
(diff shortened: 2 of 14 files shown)


src/alr/alr-commands-show.ads
```diff
@@ -21,7 +21,7 @@ package Alr.Commands.Show is
       Config : in out CLIC.Subcommand.Switches_Configuration);
 
    overriding function Short_Description (Cmd : Command) return String is
-     ("See information about a release");
+     ("Display information about a crate");
```
mosteo: There's actually a somewhat of a fine distintion between crate and release. A release is a version of a crate, and the show command actually operates on a concrete version (the last one by default). Perhaps `crate version` is better if the idea is to avoid `release` as not to confuse users unnecessarily.

src/alr/alr-commands.adb
```diff
@@ -560,30 +560,31 @@ begin
...
+   Sub_Cmd.Register ("Crate", new Printenv.Command);
+   Sub_Cmd.Register ("Crate", new Run.Command);
+   Sub_Cmd.Register ("Crate", new Update.Command);
+   Sub_Cmd.Register ("Crate", new Withing.Command);
 
    Sub_Cmd.Register ("Publish", new Publish.Command);
 
+   Sub_Cmd.Register ("Internal", new Action.Command);
+   Sub_Cmd.Register ("Internal", new Dev.Command);
+   Sub_Cmd.Register ("Internal", new Test.Command);
```
mosteo: I'm not very thrilled by this `Internal` category. I'd go with `Testing` better.

mosteo (approved):

mosteo:
Thanks, onox.

Issue #1172, opened 2022-09-05
Homebrew on macOS


simonjwright:
I've been working on integrating Homebrew into Alire.

I've got it to a point where it works for me; I patched up the [ci-linux](https://github.com/simonjwright/alire/actions/runs/2988808227) and [ci-macos](https://github.com/simonjwright/alire/actions/runs/2988808226) workflows, the only reasons the latter failed were
```
-Ada -> Alire_Host_Distro: distro_unknown
+Ada -> Alire_Host_Distro: macos
```
and
```
-C -> Alire_Host_Distro: distro_unknown
+C -> Alire_Host_Distro: macos
```

Is this work of interest? If so, what branch should I base a PR on? (I would do a merge squash on the relevant commits I've made so far in my `homebrew-dev` branch.)

Further,

* Is `macos` OK for `Alire.Platforms.Distribution`? (actually `MacOS`). Obviously the test that failed above needs fixing!
  * That would be adding e.g. `on_macos()` to `testsuite/drivers/helpers.py`, and altering `distribution()` to use it. I see that `platform.system()` returns `Darwin`, I guess using `darwin` instead of `macos` would confuse people!
* Is there a way of running github actions for testing crates using this version of `alr` from my repo?
* What would the process be for amending external manifests, e.g. `sdl2`, `sdl2_ttf`, `sdl2_image`? We could invite Mac users to make/propose changes as needed?
* What would the process be for amending crates that could use this ability, e.g. `sdlada` (which is maintained by @mosteo but authored by Luke Guest), and ought to use `$HOMEBREW_PREFIX` to find include & library files (Homebrew don't recommend moving it from its default location, but the default location is different between x86_64 and aarch64!)

Fabien-Chouteau:
Thanks a lot @simonjwright !

 I commented on PR for code details.

> * Is `macos` OK for `Alire.Platforms.Distribution`? (actually `MacOS`). Obviously the test that failed above needs fixing!

As I said on the PR, the distribution should be `Homebrew`.

>     * Is there a way of running github actions for testing crates using this version of `alr` from my repo?

I am not sure, we have the Crate CI stuff but I don't know if you can use it (@mosteo ).
 
>     * What would the process be for amending external manifests, e.g. `sdl2`, `sdl2_ttf`, `sdl2_image`? We could invite Mac users to make/propose changes as needed?

[...]

PR #1185, opened 2022-09-09
Implement Homebrew as the system deployer for macOS.

simonjwright:
See Issue #1172.

It's not clear to me how we could deal with MacPorts as an alternate deployer, if anyone wants to do the work (won't be me).

I'm not sure about  `System_Package_For_Tool` in `src/alire/alire-utils-tools.adb`; I've treated macOS like all the other distributions so as to get a build, but I don't understand how/when this is used.


src/alire/alire-origins-deployers-system-homebrew.adb
```diff
@@ -0,0 +1,273 @@
...
+
+package body Alire.Origins.Deployers.System.Homebrew is
+
+   pragma Warnings (Off);
+
+   --  Ada.Strings.Unbounded is use-visible via Alire.Origins.
+   use GNATCOLL.JSON;
+
+   Homebrew_Prefix : constant String
+     := Ada.Environment_Variables.Value ("HOMEBREW_PREFIX", "");
```
Fabien-Chouteau: Across the project we use `Alire.OS_Lib.Getenv` to get environment variable . We should probably use `Ada.Environment_Variables.Value` but for consistency let's use keep `Alire.OS_Lib.Getenv` everywhere.

src/alire/alire-origins-deployers-system-homebrew.adb
```diff
@@ -0,0 +1,273 @@
...
+         if Length (Arr) /= 1 then
+            raise JSON_Issue with "JSON info length /= 1";
+         end if;
+         if Kind (Get (Arr, 1)) /= JSON_Object_Type then
+            raise JSON_Issue with "JSON info (1) not JSON_Object";
+         end if;
+         Map_JSON_Object (Get (Arr, 1), Info_Callback'Access);
+      end;
+
+   --  exception
```
Fabien-Chouteau: Dead code?

src/alire/alire-origins-deployers-system-homebrew.adb
Fabien-Chouteau: Is the `HOMEBREW_PREFIX` env var something common for Homebrew users? Or a requirement only for Alire users?

src/alire/alire-origins-deployers-system-homebrew.adb
```diff
@@ -0,0 +1,273 @@
...
+         else
+            return Version_Outcomes.Outcome_Failure
+              ("unexpected version format",
+               Report => False);
+         end if;
+      end Get_Version_From_String;
+
+   begin
+      Trace.Debug ("detect? " & This.Base.Package_Name);
+      if Homebrew_Present then
```
Fabien-Chouteau: Detection of the "distrib" should be done in the platform specific code. There's no need to do it here as far as I know.

src/alire/alire-origins-deployers-system-homebrew.adb
```diff
@@ -0,0 +1,273 @@
...
+   -- Detect --
+   ------------
+
+   overriding
+   function Detect (This : Deployer) return Version_Outcomes.Outcome
+   is
+
+      function Get_Version_From_String (Candidate : String)
+                                       return Version_Outcomes.Outcome;
+      function Get_Version_From_String (Candidate : String)
```
Fabien-Chouteau: The `Semantic_Versioning` library is already providing this kind of conversion. Is there a reason not to use it?

src/alire/alire-platforms.ads
```diff
@@ -40,6 +40,7 @@ package Alire.Platforms with Preelaborate is
                           Rhel, -- RedHat Enterprise Linux
                           Centos,
                           Fedora,
+                          MacOS,
```
Fabien-Chouteau: This entry should be `Homebew` and not `MacOS`, for the same reasons that we have `Msys2` and not `Windows`:
 - You can use Alire on MacOS without Homebew
 - You can have other system package managers on MacOS (Macports)

src/alire/os_macos/alire-platforms-current__macos.adb
```diff
@@ -9,7 +9,7 @@ package body Alire.Platforms.Current is
    ------------------
 
    function Detected_Distribution return Platforms.Distributions is
-      (Platforms.Distro_Unknown);
+      (Platforms.MacOS);
```
Fabien-Chouteau: This is where you detect the `HOMEBREW_PREFIX` env variable and return `Platforms.Homebew` if present, `Platforms.Distro_Unknown` otherwise.

testsuite/drivers/helpers.py
```diff
@@ -89,6 +93,8 @@ def distribution():
 
         return 'DISTRO_UNKNOWN'
 
+    elif on_macos():
```
Fabien-Chouteau, line 96: This will have to be changed according to the comments about MacOS vs Homebew distrib.

Fabien-Chouteau (changes requested):
Awesome @simonjwright !

I have a few comments, the main one being about the `MacOS` vs `Homebrew` distrib.

Thanks for your contribution,

src/alire/alire-origins-deployers-system-homebrew.adb
simonjwright: OK (but may become OBE depending on how we resolve the Homebrew/MacPorts issue)

src/alire/alire-origins-deployers-system-homebrew.adb
simonjwright: Dead comment, anyway! OK.

src/alire/alire-origins-deployers-system-homebrew.adb
simonjwright: No reason. The existing code will (try to) detect format errors, is it OK to let `Semantic_Versioning` raise an exception in this case? would said exception need to be handled? (it doesn't seem to be handled in other uses).

My code would report libtcl's version as 8.6.12, Semantic_Versioning reports 8.6.12+_1 - would that be a problem? (building my test code with this works fine).

Fabien-Chouteau: @mosteo I think we need your input here.

src/alire/alire-origins-deployers-system-homebrew.adb
simonjwright: OK

src/alire/alire-platforms.ads
simonjwright: OK

src/alire/os_macos/alire-platforms-current__macos.adb
simonjwright: OK

src/alire/alire-origins-deployers-system-homebrew.adb
simonjwright: The 'normal' way for users to set up to use Homebrew is to do something like
```
eval "$(/where/ever/bin/brew shellenv)"
``` 
in their `.bash_profile` or equivalent, where `brew shellenv` returns (on my aarch64 Mac)
```
export HOMEBREW_PREFIX="/opt/homebrew";
export HOMEBREW_CELLAR="/opt/homebrew/Cellar";
export HOMEBREW_REPOSITORY="/opt/homebrew";
export PATH="/opt/homebrew/bin:/opt/homebrew/sbin${PATH+:$PATH}";
export MANPATH="/opt/homebrew/share/man${MANPATH+:$MANPATH}:";
export INFOPATH="/opt/homebrew/share/info:${INFOPATH:-}";
```
Of course they could just set the `PATH` themselves. Which means no `HOMEBREW_PREFIX` etc; but then consider [this](https://github.com/simonjwright/tcladashell/blob/8510f86420d0b33bca68441592efc427a7509e5d/options.gpr#L36) and [this](https://github.com/simonjwright/tcladashell/blob/8510f86420d0b33bca68441592efc427a7509e5d/options.gpr#L63) in my tash crate (to be announced), 

The less-obvious issue is that Homebrew on `x86_64-apple-darwin` installs in `/usr/local` by default, but on `aarch64` it installs in `/opt/homebrew` (so as not to confuse people who have both architectures installed???) -- the `/usr/local` thing was my chief reason for rejecting Homebrew; you can change it, e.g. on my x86_64 Mac I have it in `~/homebrew`, which has the disadvantage that packages have to be rebuilt from source rather than being available in binary.

simonjwright: The bottom line is, we can probably get away without setting up these environment variables for `alr` itself (check whether `brew` is on the `PATH`), but a crate that depends on an external will need them.

Fabien-Chouteau: If `eval "$(/where/ever/bin/brew shellenv)"` is the normal way then using `HOMEBREW_PREFIX` sounds like a perfectly good solution to me :+1:

testsuite/drivers/helpers.py
simonjwright: This is for CI, and the CI runners have Homebrew installed (but not the environment variables! see [here](https://github.com/simonjwright/tcladashell/blob/8510f86420d0b33bca68441592efc427a7509e5d/.github/workflows/ci-macos.yml#L42)).

So can I punt the Macports issue for the moment and just return `'HOMEBREW'`?

Fabien-Chouteau: I think what we might want to do is have a CI run that use Homebrew (`eval $(brew shellenv)`) and one that doesn't. @mosteo  ?

src/alire/alire-origins-deployers-system-homebrew.adb
simonjwright: Actually, I suspect that this code will only get called at all if the `Distribution` is `Homebrew`? in which case the question is moot (but I would have to change the implementation of `Already_Installed `to use the `Installed_Version` found by `Get_Info`, because my lazy implementation relies on knowing `Homebrew_Prefix`).

And the tests for `Homebrew_Present` would be redundant.

src/alire/alire-origins-deployers-system-homebrew.adb
mosteo: The extra build part should not be a problem. We have to be careful if distributions use `-` in a sense different than semantic versioning pre-releases, though, to avoid mis-reporting versions.

These detectors shouldn't raise; in the worst case we prefer the external to be reported as undetected. But then, relaxed semantic versioning will accept anything that starts with a number as a version.

testsuite/drivers/helpers.py
mosteo: Sounds right to me, just as we test with no package manager on linux and with the supported ones too.

src/alire/alire-platforms.ads
Joebeazelman: Homebrew also supports Windows 10 subsystem and Linux. You might need to create a package matrix or perhaps another environment variable for the parent package manager, i.e. how was Alire installed regardless of platform. This might be a consideration if, for instance, some operation might need the information.

simonjwright: > Homebrew also supports Windows 10 subsystem and Linux. You might need to create a package matrix or perhaps another environment variable for the parent package manager, i.e. how was Alire installed regardless of platform. This might be a consideration if, for instance, some operation might need the information.

When you say "You might need to create ..." I hope you don't mean me? This sounds way outside my comfort zone

testsuite/drivers/helpers.py
simonjwright: I could certainly set this up.

I did think about checking whether `brew` is on the `PATH`, but users might want to use a different package manager, & if they had both this could lead to confusion; checking for `HOMEBREW_PREFIX` would mean they were OK with it.

Would have to change `testsuite/drivers/helpers.py` to make the same choice: e.g. like [this](https://github.com/simonjwright/alire/commit/50f4fe9ed5cd6aa037ea9a6558a984bde9bb2b46) and [this](https://github.com/simonjwright/alire/commit/913e5573e3d47557d2e750fe2e6e71f2202a3eec), successful run [here](https://github.com/simonjwright/alire/actions/runs/3077781843)

src/alire/alire-platforms.ads
mosteo: > When you say "You might need to create ..." I hope you don't mean me? This sounds way outside my comfort zone

I guess it was said as hypothetical future feature. In any case, Alire currently only supports one package manager per distro. But, certainly, one can use combos [os x distro] in the externals, although the need has not yet arisen.

Which makes me think that, as we have `macOS` as the OS, maybe it's more appropriate to use `homebrew` as the distro just like `msys2` is used on Windows. This would enable using other package manager on macOS if someone wants to come up with it.

simonjwright: Yes, that’s how I’m approaching it. I’m tidying up the update to this PR.

mosteo:
Looks good to me, outstanding effort, Simon. Since this PR has had quite some discussion, I'm going to wait for @Fabien-Chouteau feedback before merging.

src/alire/alire-origins-deployers-system-homebrew.adb
```diff
@@ -0,0 +1,219 @@
+with AAA.Strings; use AAA.Strings;
+
+with Alire.OS_Lib.Subprocess;
+with Alire.Errors;
+
+with GNATCOLL.JSON;
+
+package body Alire.Origins.Deployers.System.Homebrew is
+
+   pragma Warnings (Off);
```
Fabien-Chouteau: All warnings off for the entire file? You can target a specific warning, but disabling all of them is not ok.

Fabien-Chouteau (changes requested):

Fabien-Chouteau (approved):

Fabien-Chouteau:
Thanks @simonjwright, can you fix the conflicts with the OpenSUSE support that was just added? Sorry for that

simonjwright:
I did fix the conflicts, and hoped to see CI runs; maybe the reason why they didn’t happen is that Github thinks there are still conflicts?

Fabien-Chouteau:
GitHub says there's still one conflic:
```
            System.Zypper.Deployer'(Deployers.Deployer'(Base => From)
<<<<<<< homebrew-merged
                                    with others => <>),
         when Platforms.Homebrew =>
            System.Homebrew.Deployer'(Deployers.Deployer'(Base => From)
                                      with others => <>));
=======
                                    with others => <>));
>>>>>>> master
```

simonjwright:
Actually, Github says there are 3 conflicts, but they are in fact the *changes* that I want to make. I don’t see how I can fix this ...

Oh. Perhaps the problem is that I cherry-picked the openSuse commit from master? ... but that introduced conflicts which I fixed, and I don’t see any other trace of it than the commit (cb3ca3b) author ...

Oh². I can make it happen here. Baffled for the time being

mosteo:
Let me take a stab at this

mosteo:
Sorry for stepping in in your fixing, Simon. Merged now!

simonjwright:
> Sorry for stepping in in your fixing, Simon. Merged now!

Brilliant! Thanks ..

PR #1228, opened 2022-10-24
alr-commands-init.adb: make alr init interactive

Fabien-Chouteau:
Alr init will now queries the user for information such as license, description, tags, etc.

Non-interactive usage is still possible using existing switches.

### DIFF
--- a/src/alire/alire-utils.adb
+++ b/src/alire/alire-utils.adb
@@ -189,6 +189,7 @@ package body Alire.Utils is
 
    function Is_Valid_Tag (Tag : String) return Boolean is
      ((for all C of Tag => C in '0' .. '9' | 'a' .. 'z' | '-')
+      and then Tag'Length in 1 .. Max_Tag_Length
       and then Tag (Tag'First) /= '-'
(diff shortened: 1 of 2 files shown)


Fabien-Chouteau:
Something I forgot to mention in the commit is that non-interactive usage is still possible but it requires the -n switch.

Using only `alr init --bin tests` will not work automatically as it did before.

mosteo:
So our test suite works because it was already using `-n`. OK.

Looks good in general, but I tried `alr -nq init` and I still get a lot of output. What if for scripting we mute the by-default-interactions too?

Fabien-Chouteau:
I don't mind seeing the default value being used in the non-interactive logs. Do you mean a special case when both `-n` and `-q` are used?

mosteo:
You're right that without `-n`, in this case `-q` does not make sense. But if the user does use both `-n` and `-q`, it seems counter-intuitive to show the mock interactions. 

(But I realize we probably have other similar cases (e.g. `alr publish`) so I'll merge as-is and it can be a future enhancement.)

mosteo (approved):

PR #1253, opened 2022-12-01
Update getting-started.md

Fabien-Chouteau:
We don't want to imply that those info will be required at some point. In particular we don't want the use of Alire to require a GitHub account.


mosteo (approved):

mosteo:
Good point.

PR #1254, opened 2022-12-01
Alr.Commands.Search: unify "search" and "search --crates" behavior

### LINKED ISSUE #1250: `alr search` should include descriptions
I wanted to find how to install SPARK using Alire, so, after failing with `alr get spark`, I tried `alr search spark`. This produced two hits - neither of which are actually the SPARK toolchain - even though `gnatprove` has, in its description, "Automatic formal verification of SPARK code".

I suggest that it would be helpful if `alr search` included results from descriptions. (Given that the help says "Search a string in release names and properties", I'm surprised that it does not do so, currently.)

Fabien-Chouteau:
They now both look in names and descriptions, case insensitive.

### DIFF
--- a/src/alr/alr-commands-search.adb
+++ b/src/alr/alr-commands-search.adb
@@ -239,10 +239,12 @@ package body Alr.Commands.Search is
          while Has_Element (I) loop
 
             declare
+               use AAA.Strings;
+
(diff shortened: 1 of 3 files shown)


Fabien-Chouteau:
Ref #1250

mosteo (approved):

PR #1262, opened 2022-12-05
Test alr from project dir

### LINKED ISSUE #1260: Testsuite runs any `alr` in `$PATH`
When I run the testsuite on my computer, lots of tests fail. Some do because the output has additional spaces at the end of lines (probably some whitespace alignment when printing process output on my computer), some expect slightly different messages (eg. `Fetch a published crate` instead of `Fetches a crate release` or `last_build_profile=xxx=DEVELOPMENT` instead of `last_build_profile=DEVELOPMENT`) and so on.

When I change expected values in problematic tests, they work fine on my computer, but fail on CI for this reason.

If the tests are expected to work the same way locally and on CI, I can provide you with more details.

0rzech:
Fixes #1260.

### DIFF
--- a/testsuite/run.py
+++ b/testsuite/run.py
@@ -30,6 +30,16 @@ class Testsuite(e3.testsuite.Testsuite):
         # Some tests rely on an initially empty GPR_PROJECT_PATH variable
         os.environ.pop('GPR_PROJECT_PATH', None)
 
+        # Let's run `alr` from project's dir
+        alr_path = self._get_alr_path()
+        self._prepend_alr_to_path_env_var(alr_path)
+
+    def _get_alr_path(self):
+        project_root = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
+        return os.path.join(project_root, 'bin')
+
+    def _prepend_alr_to_path_env_var(self, alr_path):
+        os.environ['PATH'] = alr_path + os.pathsep + os.environ['PATH']
 
 if __name__ == '__main__':
     suite = Testsuite()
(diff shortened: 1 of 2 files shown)


testsuite/run.py
```diff
@@ -30,6 +30,16 @@ def set_up(self):
         # Some tests rely on an initially empty GPR_PROJECT_PATH variable
         os.environ.pop('GPR_PROJECT_PATH', None)
 
+        # Let's run `alr` from project's build dir
+        alr_path = self._get_alr_path()
+        self._prepend_alr_to_path(alr_path)
+
+    def _get_alr_path(self):
```
0rzech, line 37: Should we add these functions (`_get_alr_path` and `_prepend_alr_to_path`) here, or in `PythonScriptDriver`?

0rzech:
Should we add `_get_alr_path` and `_prepend_alr_to_path` to `Testsuite` or to `PythonScriptDriver` class?

mosteo (approved):

mosteo:
Looks good to me, thanks. @Fabien-Chouteau , do you think this is safe?

Fabien-Chouteau:
This can be also confusing if one wants to test another build, but it's a better default I guess.
I would prefer to have an option on the command line to specify another executable.

mosteo:
Makes sense. That can be added as a further improvement.

mosteo:
Thanks, @0rzech .

0rzech:
@mosteo @Fabien-Chouteau Not sure about Windows, but on Linux and MacOS running
```sh
PATH="${HOME}/path/to/alire/bin:${PATH}" ./run.py
```
should do the job when one wants to test some out-of-tree build of alr.

PR #1284, opened 2023-01-03
Update msys2 installer

Fabien-Chouteau:


Fabien-Chouteau:
It's the `pacman -Su`  command that is causing trouble. I don't know why yet and how to fix it.
https://github.com/alire-project/alire/blob/a5aa0e399b5d790c7cdc37e21b7c7f4d84e68b4f/src/alire/os_windows/alire-platforms-current__windows.adb#L330

Since there's also this problem with `1.2.1` we might have to publish a fix release.

mosteo:
This is vexing because IIRC I added that second update precisely to survive some other msys2 problems. Oh well.

mosteo:
I think I've got the solution. Push incoming.

mosteo:
The remaining failure is because the self-build is done with the latest release, which doesn't contain the fix. So I think we're good to go.

mosteo:
@Fabien-Chouteau, for some reason I cannot add you directly as reviewer. If you can take a look at the PR, thanks!

Fabien-Chouteau:
Because I am the author of the PR. Also I can't approve the PR.

mosteo:
Oh right...

mosteo (approved):

PR #1285, opened 2023-01-06
GPR Generator

jquorning:
Separate generation of GPR file into generic package


src/alr/alr-commands-init.adb
```diff
@@ -121,56 +125,113 @@ package body Alr.Commands.Init is
...
+
+         GPR_File.NL_Before_IS      := False;
+         GPR_File.NL_After_FOR_List := False;
+         --  GPR File Style
+
+         NL;
+         WC ("config/" & Lower_Name & "_config.gpr");
+         NL;
+         RB (Mixed_Name, GPR_File.Library);
+         NL;
```
mosteo: I'm not sure I like this kind of super-telegraphic style :/

mosteo:
The Windows failures should be fixed by a merge or rebase.

src/alr/alr-commands-init.adb
Fabien-Chouteau: Me neither. I don't find this version to be more readable than the existing one.

In the grand scheme of things, it could be a good idea to make a library to facilitate the generation of GPR files.
But for the limited use we have in Alire, I don't think this change make sense.

src/alr/gpr_generator.ads
```diff
@@ -0,0 +1,184 @@
+with Ada.Strings.Unbounded;
+
+generic
+   with procedure Put (Item : String) is <>;
+   with procedure New_Line is <>;
+package GPR_Generator
```
Fabien-Chouteau: This package would have to be either in the `Alr` or the `Alire` hierarchy.

src/alr/alr-commands-init.adb
jquorning: > I'm not sure I like this kind of super-telegraphic style :/

I believe it is called *Abstraction*, a feature supported by Ada 83 ;-)

JeremyGrosser: Seems more like obfuscation to me.

jquorning: > Seems more like obfuscation to me.

Then *Rapid Applcation Development* is obscure.. ;-)

0rzech: > I believe it is called Abstraction, a feature supported by Ada 83 ;-)

> Then Rapid Applcation Development is obscure.. ;-)

Neither abstraction nor RAD is about abbreviating subprogram names. 😉

The problem is, after a while `CO`,`EL`, `EX`, `FR`, `FU`, `LO`, `NL`, `PB`, `PU`, `QE`, `RB`, `RE`, `WC` will require jumping to their respective definitions, probably even for the author of the code, in order to understand what is happening.

The code seems to be optimized for typing time instead of comprehension time and I think that in general it's better to do the opposite.

jquorning: > Neither abstraction nor RAD is about abbreviating subprogram names. 😉

Well it might just be a *Ruby* thing then :-)

> The problem is, after a while `CO`,`EL`, `EX`, `FR`, `FU`, `LO`, `NL`, `PB`, `PU`, `QE`, `RB`, `RE`, `WC` will require jumping to their respective definitions, probably even for the author of the code, in order to understand what is happening.

Well, I guess a modern IDE should be able to help you out on this one :-)

> The code seems to be optimized for typing time instead of comprehension time and I think that in general it's better to do the opposite.

Give me a break! If I optimized for typing time I wouldn't have created the PR in the first place :-) ;-)

I tried to do some data-centric approach.
I'll figure out something ..

I'be back in a day or two...

I the meantime hang on with a little entertainment [tralalalulalaj](https://youtu.be/jDrUwP09w74)

0rzech: > Well it might just be a _Ruby_ thing then :-)

I'd still be surprised, but then again, I don't know Ruby ecosystem. 😉

> Well, I guess a modern IDE should be able to help you out on this one :-)

I expected this. 😜 But it misses the point, because I'd still be forced to jump to definitions or use some kind of tooltips instead of just reading `Comment`, `Expression` etc.

> Give me a break! If I optimized for typing time I wouldn't have created the PR in the first place :-) ;-)

Ha! 😄

> I tried to do some data-centric approach.

Well, SQL is data-centric and yet we have `SELECT`, `FROM`, `WHERE` instead of `SE`, `FR`, `WH`. 😉

> I'll figure out something ..
>
> I'be back in a day or two...
>
> I the meantime hang on with a little entertainment [tralalalulalaj](https://youtu.be/jDrUwP09w74)

Thanks and have a nice time! 🙂

0rzech: > Well it might just be a _Ruby_ thing then :-)

I'd still be surprised, but then again, I don't know Ruby ecosystem. 😉

> Well, I guess a modern IDE should be able to help you out on this one :-)

But this misses the point, because I'd still have to jump to definitions or use some sort of tooltips instead of just reading `Comment`, `Expression` etc. 😉

> Give me a break! If I optimized for typing time I wouldn't have created the PR in the first place :-) ;-)

Ha! 😄

> I tried to do some data-centric approach.

I think it's good to look at how data-centric languages do it. For instance, SQL has `SELECT`, `FROM`, `WHERE` and not `SE`, `FR`, `WH`.

> I'll figure out something ..
>
> I'be back in a day or two...
>
> I the meantime hang on with a little entertainment [tralalalulalaj](https://youtu.be/jDrUwP09w74)

Thanks and I wish you a nice time! 🙂

jquorning:
Follow the work on https://github.com/jquorning/vanilla_replacer_v1

PR #1291, opened 2023-01-14
Run testsuite on selfbuilds

mosteo:
We are currently testing that `alr` can build itself, but not that the result passes tests


mosteo:
This is not turning out as I hoped, will make another attempt at a later time.

PR #1297, opened 2023-01-19
Update README.md

Fabien-Chouteau:


mosteo:
This is redundant with #1298 now.

PR #1298, opened 2023-01-19
alire_common.gpr: remove default host os to avoid mistakes on macOS

Fabien-Chouteau:
With the default being Linux, macOS users will not notice any issue until they try to use the executable. Even then the issue will be obscure.

### DIFF
--- a/.github/workflows/ci-windows.yml
+++ b/.github/workflows/ci-windows.yml
@@ -10,6 +10,9 @@ on:
   release:
     types: [created]
 
+env:
+  ALIRE_OS: "windows"
+
 jobs:
 
   build:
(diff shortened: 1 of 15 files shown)


Fabien-Chouteau:
This might trigger some build issues...

mosteo:
Then I think it will be better to remove reliance on `OS` and having a default as well from `alr_env.gpr`. Better to make it explicit everywhere?

Note to self: this will require fixing the `setup_alire` action for the case of building a branch.

mosteo:
I've removed any guess attempts from `alr_env.gpr` too, and added a couple of bash functions for our use in tests/builds. Also reviewed a bit the README. Removed a couple of obsolete scripts and tweaked the rest.

I think we're good to go if you don't see any issue in my changes, @Fabien-Chouteau

mosteo (approved):

Fabien-Chouteau:
Looks good to me :+1: Thanks for picking that up

PR #1322, opened 2023-02-10
Install a local crate

mosteo:
This one adds the ability to install a release locally available.

It's a bit more involved that just doing a blind `gprinstall -f -r` because with the info at hand we have easy ways to catch obvious problems like conflicting executables, and we can present more accurate info on what's going on during install. 

Everything ends in `<prefix>/bin` in usage mode, which means that only executables and dynamic libraries are installed, bypassing all the problems we discussed.

### DIFF
--- a/doc/user-changes.md
+++ b/doc/user-changes.md
@@ -6,7 +6,30 @@ stay on top of `alr` new features.
 
 ## Release 1.3-dev
 
-### New subcommand `alr install
+### Installation of local crates
+
+PR [#1322](https://github.com/alire-project/alire/pull/1322)
+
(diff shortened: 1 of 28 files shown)


Fabien-Chouteau (commented):
I am not a huge fan of the `--this` option. Can't we just say this is the default?

mosteo:
I did it that way just like in many other commands the default is to show some info (`alr index`, `alr with`), but on 2nd thought I see this is probably expected to work more like `alr build` or `alr get` so yes, let's do that.

Fabien-Chouteau (approved):

PR #1332, opened 2023-02-21
Unicode documentation

mosteo:
I think I have now a clear understanding of the implications of `-gnatW8`. I've tried to document the practical issues to put the matter to rest. So this PR includes my findings in an AEP file for our internal record, and also a document intended to go into the website for people that come looking for information on that.

I've also added the option to disable `-gnatW8` just so if in some case it really doesn't make sense (not for the public index, but maybe in some legacy private codebase) you're not forced to disable the whole crate configuration system.

Next steps would be to implement the check for the index submissions and to fix the few crates that cause problems.

Besides you @Fabien-Chouteau, it would be great if other people familiar with the issue (@reznikmm, ...) can verify I'm not making some wrong assumptions or statements in these documents. The `doc/unicode.md` can probably be improved as I tend to err to the convoluted side.

### DIFF
--- a/doc/unicode.md
+++ b/doc/unicode.md
@@ -1,27 +1,52 @@
 # Unicode guidelines
 
-Unicode is the solution to work with character sets other than ASCII or, in the case of Ada, beyond Latin-1, which was the character set chosen for the `Character` type. There are a number of considerations for the use of Unicode in Ada and in Alire crates, detailed in this section.
+Unicode is the solution to work with character sets other than ASCII or, in the
+case of Ada, beyond Latin-1, which was the character set chosen for the
+`Character` type. There are a number of considerations for the use of Unicode
+in Ada and in Alire crates, detailed in this section.
 
 ## TL;DR
 
-Alire v2.0 and onwards expects sources to be UTF-8 (through the `-gnatW8` switch). Input/Output must be done using specialized crates (e.g. `vss`, `uxstrings`) or with `Wide_Wide_Text_IO` (for unencoded/UTF-32 strings) or with `Streams_IO/GNAT.IO` (for UTF-8-encoded strings). Avoid `Text_IO` unless you are sure of what you're doing (using Latin-1 strings). Some friction may happen with older sources not using UTF-8.
+Alire v2.0 and onwards expects sources to be UTF-8 (through the `-gnatW8`
+switch). Input/Output must be done using specialized crates (e.g. `vss`,
+`uxstrings`) or with `Wide_Wide_Text_IO` (for unencoded/UTF-32 strings) or with
+`Streams_IO/GNAT.IO` (for UTF-8-encoded strings). Avoid `Text_IO` unless you
+are sure of what you're doing (using Latin-1 strings). Some friction may happen
+with older sources not using UTF-8.
 
 Read on for the gory details.
 ## Alire assumptions and configuration
 
-By default, any crate initialized via `alr init` will have the `-gnatW8` switch in its build configuration, which presumes UTF-8 encoding of sources, and tweaks some internals of the compiler accordingly.
+By default, any crate initialized via `alr init` will have the `-gnatW8` switch
+in its build configuration, which presumes UTF-8 encoding of sources, and
+tweaks some internals of the compiler accordingly.
 
-This means that source files **must** use UTF-8 encoding when not using plain ASCII.
+This means that source files **must** use UTF-8 encoding when not using plain
+ASCII.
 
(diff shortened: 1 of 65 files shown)


src/alire/alire-utils-switches.ads
```diff
@@ -52,6 +53,11 @@ is
                              Ada83, Ada95, Ada05, Ada12, Ada2022,
                              GNAT_Extensions);
 
+   type Unicode_Support_Kind is
+     (No_Support,  --  Default behavior of GNAT, no switches set (Latin1)
```
Fabien-Chouteau: I don't think those names are meaningful enough.

doc/unicode.md
```diff
@@ -0,0 +1,164 @@
...
+    1. By default, Ada.Wide_Wide_Text_IO will use UTF-8 encoding if `-gnatW8` is in effect.
+    1. Most modern plain text formats expect UTF-8 encoding.
+    1. On Linux terminals this is the usual default.
+    1. On Windows 10:
+        1. You can enable UTF-8 once by running `chcp 65001` in your terminal.
+        1. You can enable UTF-8 in terminals system-wide.
+            - This will break programs assuming the default Windows 10 encoding.
+    1. On Windows 11, UTF-8 terminals are the default.
+1. You can use `GNAT.IO` or `Ada.Streams.Stream_IO` to output properly encoded strings, as they don't manipulate the bytes.
+1. You **cannot** use `Ada.Text_IO` to output UTF-8-encoded `String` variables, as Latin-1 encoding is expected.
```
Fabien-Chouteau, line 66: That sounds scary since every Ada will ask you to use `Ada.Text_IO` to print strings.

Fabien-Chouteau (commented):
The more I see about this issue the more I think this is not for Alire to fix. Something has to change either in the language or the compiler (maybe both).

doc/unicode.md
```diff
@@ -0,0 +1,121 @@
...
+--  Without -gnatW8 but with an UTF-8 file it will fail as the sequence will be 6 bytes long.
+
+Euro : constant String := "€";
+--  Without -gnatW8 and a UTF-8-encoded file, this results in a 3-byte UTF-8 string.
+--  With -gnatW8 and a UTF-8-encoded file, this results in an error as '€' is outside of Character.
+```
+
+## Unicode in identifiers
+
+By default, GNAT only allows Latin-1 identifiers. To enable full Unicode identifiers you can use `-gnatiw`. This is generally not recommended, as that may greatly difficult others to work with your code. Alire does not enable full Unicode in identifiers by default.
```
reznikmm: In presence of `-gnatW8` no other switches is required to have non-ASCII identifiers. So `-gnatiw` only required if there is no `-gnatW8`.

mosteo: Right, this was a leftover.

doc/unicode.md
mosteo: Yeah, it's the ugliest aspect of this issue. Unfortunately this works "by chance" when the terminal is UTF-8, the file is UTF-8, but no `-gnatW8` is given, as TIO lets "fake" Latin-1 strings which are UTF-8 get by unmodified. If the source is properly Latin-1-encoded then the output of TIO is again wrong [meaning it won't display properly in UTF-8 terminals].

mosteo:
> The more I see about this issue the more I think this is not for Alire to fix. Something has to change either in the language or the compiler (maybe both).

Can you get a feel about this from people in AdaCore? What are people doing in their codebases? Is this a usual thing in coding conventions? What I've seen in comp.lang.ada is that using `Ada.Text_IO` for Unicode has always been seen as a mistake and asking for trouble.

I don't think the language is going to change, it already admits full Unicode. It's more of a problem with legacy codebases and compiler defaults, unlikely to change either, I'd say.

We are in the crossfire of wanting code from multiple origins to always interact properly. Closed projects likely have policies about this and is a non-issue. That said, I tend to think we're in the proper spot to adopt this now and get done with it for good.

I will try to get more eyes/feedback on it.

src/alire/alire-utils-switches.ads
mosteo: What about `Source_Encoding`/`ASCII`/`Latin_1`/`UTF_8`? (ASCII/Latin_1 would be equivalent -- no `-gnatW8`)

doc/unicode.md
mosteo: Actually that was already removed in a later commit. Better if you check the changes as a whole instead of commit-by-commit.

stcarrez:
This `-gnatW8` option breaks a simple Ada program which uses `Ada.Directories` and prints file names.
See https://github.com/alire-project/alire/discussions/1334#discussioncomment-5205705

Asking people not to use `Ada.Text_IO` is definitely a very bad idea.

Unless the compiler is fixed, we will have many issues if you set this option by default.

Fabien-Chouteau (approved):

mosteo:
I will merge this one then, as I think it's better to have some rationale rather than just simply enabling `-gnatW8` silently, but improvements to the documentation are welcomed.

PR #1342, opened 2023-03-02
Don't enable color if TERM=dumb

### LINKED ISSUE #1256: Output when building from `emacs` contains escape codes
I suggest that `alr` should not emit escape codes as default, but turn on colours etc only when instructed. Essentially what I am trying to say is to have a shell script front-end. RFC?

This is output when building from emacs:

```
-*- mode: compilation; default-directory: "~/GitHub/jquorning/hello/src/" -*-
Compilation started at Thu Dec  1 19:01:59

alr build
^[[36m^[[1mⓘ^[[39m^[[22m Building ^[[1mhello^[[22m/^[[35m^[[1mhello.gpr^[[39m^[[22m...
^[[2mCompile
   [Ada]          hello.adb
Bind
   [gprbind]      hello.bexch
   [Ada]          hello.ali
Link
   [link]         hello.adb
^[[22mBuild finished successfully in ^[[1m1.16^[[22m seconds.


Compilation finished at Thu Dec  1 19:02:00
```

mosteo:
Fixes #1256

### DIFF
--- a/src/alr/alr-commands.adb
+++ b/src/alr/alr-commands.adb
@@ -1,6 +1,7 @@
 with Ada.Characters.Handling; use Ada.Characters.Handling;
 with Ada.Command_Line;
 with Ada.Directories;
+with Ada.Environment_Variables;
 with Ada.Text_IO; use Ada.Text_IO;
 
 with CLIC.TTY;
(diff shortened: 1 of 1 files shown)


src/alr/alr-commands.adb
```diff
@@ -461,6 +462,7 @@ package body Alr.Commands is
       if Alire.Platforms.Current.Operating_System /= Alire.Platforms.Windows
         and then not No_Color
         and then not No_TTY
+        and then Ada.Environment_Variables.Value ("TERM", "dumb") /= "dumb"
```
Fabien-Chouteau, line 465: Is `dumb` value a usual value for this variable?

mosteo: I'm not sure about the extent of its use but it do appears in `man 7 term` and the emacs shell seems to set it. Whatever IO lib is `cargo` using also reacts to it being set.

Fabien-Chouteau: More than enough for me :)

Fabien-Chouteau (approved):

PR #1343, opened 2023-03-10
Update getting-started.md

corinthian13:
Lines 12 - 15.

More clarity on how to add PATH entry for Alire ecosystem client installed on a user's local Linux machine.  
PATH value was missing the double quotes that _may_ be needed when file paths contain spaces, e.g. /home/me/ada stuff/alr/bin.

### DIFF
--- a/doc/getting-started.md
+++ b/doc/getting-started.md
@@ -14,7 +14,7 @@ This may be done for the duration of a terminal session by running the command b
 ```bash
 $ export PATH="<PATH_TO_EXTRACTED>/bin/:$PATH"
 ```
-Those wanting to keep this path permanently in their `PATH` environment may do so by pasting the above command into the .bashrc file of their user's account.
+Those wanting to keep this path permanently in their `PATH` environment may do so by pasting the above command into the `.profile` file of their user's account.
 
 Alire provides GNAT toolchains hosted on x86-64 for Linux. If those toolchains do not 
 work for you, or if you are on another host architecture like ARM, you have the option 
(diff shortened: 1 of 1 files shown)


doc/getting-started.md
```diff
@@ -9,9 +9,9 @@ repository](https://github.com/alire-project/alire/releases).
 
 On Linux, `Alire` is simply provided in an archive.
 
-Once the archive is extracted you have to add `alr` in the environment `PATH`:
+Once the archive is extracted you have to add `alr` in the environment `PATH`in the .bashrc file of the user's home directory:
```
Fabien-Chouteau: I am not sure everyone will want to put that in `.bashrc`. I personally prefer to set the PATH every time I want to use Alire.

corinthian13: Then I suppose it's a case of an _either ... or ..._ depending on the user's choice.
Personally I would see no danger in having this additional path entry and a lot of convenience in retaining it.

mosteo: This could be reworded as to indicate that it's an optional possibility and not a requirement for use.

corinthian13: Edited as per mosteo suggestion.

jquorning: I guess this belongs in a **setup-environment.md** document.

corinthian13: @jquorning 

Why not where it is right now, i.e. the "Getting Started" page ?
https://github.com/alire-project/alire/blob/master/doc/getting-started.md

jquorning: > Why not where it is right now, i.e. the "Getting Started" page ? 

I guess the environment is a big matrix of combinations of terminals, character sets, shells, configurations, etcetera not controlled by Alire. Especially on Linux. IMO Alire  should not try to adapt to all kinds of specialized use cases internally.

Fabien-Chouteau (approved):

mosteo (approved):

PR #1344, opened 2023-03-13
Introduce Characters package

jquorning:
Reduce use of Standard.ASCII

### DIFF
--- a/src/alr/alr.ads
+++ b/src/alr/alr.ads
@@ -24,6 +24,7 @@ package Alr with Preelaborate is
    use all type Simple_Logging.Levels;
    pragma Warnings (On);
 
+   package Characters renames Standard.ASCII;
    package Trace renames Simple_Logging;
    package TTY renames CLIC.TTY;
 
--- a/src/alr/alr.ads
+++ b/src/alr/alr.ads
@@ -24,7 +24,7 @@ package Alr with Preelaborate is
    use all type Simple_Logging.Levels;
    pragma Warnings (On);
 
-   package Characters renames Standard.ASCII;
+   package Characters renames ASCII;
    package Trace renames Simple_Logging;
    package TTY renames CLIC.TTY;
 
--- a/src/alr/alr-testing-console.adb
+++ b/src/alr/alr-testing-console.adb
@@ -6,7 +6,7 @@ with Stopwatch;
 
 package body Alr.Testing.Console is
 
-   Tab : constant Character := ASCII.HT;
+   Tab : constant Character := Characters.HT;
 
    ---------------
    -- Start_Run --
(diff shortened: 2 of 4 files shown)


mosteo:
What's the issue with the `Standard.ASCII` package?

jquorning:
The capitalization is bad. I kept it **ASCII** to avoid to much change.  
I think **ASCII** is declared `package ASCII renames Standard.Ascii`  
A better way would be to say `package Characters renames Ada.Characters.Latin_1`  

[Latin_1](https://gallatin.physics.lsa.umich.edu/LocalDocs/GRASP_Documentation/ada95lrm/rm9x-A-03-03.html)

mosteo:
It's `ASCII` here: http://www.ada-auth.org/standards/12rm/html/RM-J-5.html

Not a big point in my book anyway...

jquorning:
> It's `ASCII` here: http://www.ada-auth.org/standards/12rm/html/RM-J-5.html

It is in the [Annex J](http://www.ada-auth.org/standards/12rm/html/RM-J.html).
If ASCII is needed it can be redeclared in the project library with `package ASCII renames Standard.Ascii`, I guess.

See [10 Program Structure and Compilation Issues](https://www.adaic.org/resources/add_content/standards/12rm/html/RM-10.html). This is a major contribution of Ada.
I do not really know how a the program/project library is structured at its root.

Something like:

Program.Standard.ASCII
Program.ASCII

or

Standard.ASCII
Standard.Program

or something of that kind..

src/alr/alr.ads
```diff
@@ -24,6 +24,7 @@ package Alr with Preelaborate is
    use all type Simple_Logging.Levels;
    pragma Warnings (On);
 
+   package Characters renames ASCII;
```
jquorning, line 27: Is this the way to go?

mosteo:
> It is in the [Annex J](http://www.ada-auth.org/standards/12rm/html/RM-J.html).

Ah, I hadn't realized that was the one with the obsolescent features.

src/alr/alr.ads
mosteo: If the idea is to avoid `ASCII` because of obsolescence, I'd say `Ada.Characters.Latin_1` is preferable.

jquorning:
> > It is in the [Annex J](http://www.ada-auth.org/standards/12rm/html/RM-J.html).
> 
> Ah, I hadn't realized that was the one with the obsolescent features.

Do no not know all the details either :-)

jquorning:
> > It is in the [Annex J](http://www.ada-auth.org/standards/12rm/html/RM-J.html).
> 
> Ah, I hadn't realized that was the one with the obsolescent features.

We ought try do design a 2023 community package tree parallel to the Ada package tree..
I read somewhere that the *Ada* tree is both something and nothing. Designed for portability.
Regarding utf8/ascii/unicode/whatever we could have a package like *Root.Config* or a variable in *Root.Text_IO* like *Default_Encoding* for starters. (I do not know so much about utf8/charsets/encoding, it is a kind of a blur to me)

PR #1345, opened 2023-03-14
Introduce vector util for Indefinite_Vectors

jquorning:


jquorning:
Graph is a map, not a vector.

PR #1349, opened 2023-03-14
Actually use the cache location for shared releases

mosteo:
We fixed the cache path in #1265 but the deployer of binary releases was defaulting to the config location.

### DIFF
--- a/src/alire/alire-toolchains-solutions.adb
+++ b/src/alire/alire-toolchains-solutions.adb
@@ -1,108 +1,10 @@
 with AAA.Strings;
 
-with Alire.Config.Edit;
-with Alire.Directories;
 with Alire.Index;
-with Alire.Paths;
-with Alire.Platforms.Folders;
 with Alire.Root;
 with Alire.Shared;
 
-with CLIC.User_Input;
-
 package body Alire.Toolchains.Solutions is
-
-   -------------------
-   -- Migrate_Cache --
-   -------------------
-
-   procedure Migrate_Cache is
-      use Directories.Operators;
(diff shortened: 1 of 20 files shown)


Fabien-Chouteau (approved):
This might be worth adding a test.

mosteo:
I thought about it, but all of our tests are running with custom config to avoid getting contaminated by the user config. So in a way it's being tested, but the default isn't.

I can't think an easy way of running a test with default environment short of using something like chroot... Do you have any ideas?

Fabien-Chouteau:
I forgot about the custom config for the testsuite. I don't have a solution to offer...

mosteo:
I've got a solution for testing with default config but that will be in a separate PR as it's not trivial. 

In this one I've added a migration assistant to try and save some large downloads to people. It runs only once even if it fails, as the worst thing that can happen is that you need to redownload compilers to the new location, so I prefer not to risk it interfering unexpectedly.

src/alire/alire-toolchains-solutions.adb
```diff
@@ -1,17 +1,143 @@
...
 
+with CLIC.User_Input;
+
 package body Alire.Toolchains.Solutions is
 
+   -------------------
+   -- Migrate_Cache --
+   -------------------
+
+   procedure Migrate_Cache is
```
Fabien-Chouteau: I don't like having such a big piece of code for something that only has to be done once. Downloading dependencies is not that bad of an issue.

Fabien-Chouteau (changes requested):

Fabien-Chouteau (approved):

PR #1353, opened 2023-03-20
Typed variants of Alire.Config.Edit.Set

mosteo:
Mostly because it's too easy to mix Boolean with String now.

### DIFF
--- a/src/alire/alire-config-edit.ads
+++ b/src/alire/alire-config-edit.ads
@@ -35,11 +35,6 @@ package Alire.Config.Edit is
                           Value : Integer;
                           Check : CLIC.Config.Check_Import := null);
 
-   procedure Set_Float (Level : Config.Level;
-                        Key   : CLIC.Config.Config_Key;
-                        Value : Float;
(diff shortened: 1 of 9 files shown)


src/alire/alire-config-edit.ads
```diff
@@ -22,6 +23,23 @@ package Alire.Config.Edit is
                   Value : String;
                   Check : CLIC.Config.Check_Import := null);
 
+   --  Typed alternatives
+
+   procedure Set_Boolean (Level : Config.Level;
```
Fabien-Chouteau, line 28: Are these used somewhere?

Fabien-Chouteau (commented):
It would be better to merge the feature in CLIC before this in Alire in my opinion.

src/alire/alire-config-edit.ads
mosteo: Not yet, but there are old uses of `Set` that could be migrated

Fabien-Chouteau (approved):

PR #1356, opened 2023-03-24
Test the local release with `alr test`

mosteo:
This allows to replicate locally what should happen on `alr test <indexed crate>`.


Fabien-Chouteau (approved):
What is `alr test` doing if there is no test action defined?

mosteo:
> What is `alr test` doing if there is no test action defined?

`alr build --release`

mosteo:
> What is `alr test` doing if there is no test action defined?

`alr build --release`

PR #1368, opened 2023-04-06
Try to get the distribution ID from an environment variable

pmunts:
Rationale: Map AlmaLinux, CentOS, Rocky Linux, Oracle Linux et al to RHEL.  Or map Linux Mint to Debian.

### DIFF
--- a/src/alire/os_linux/alire-platforms-current__linux.adb
+++ b/src/alire/os_linux/alire-platforms-current__linux.adb
@@ -1,3 +1,5 @@
+with Ada.Environment_Variables;
+
 with AAA.Strings; use AAA.Strings;
 
 with GNAT.Regpat;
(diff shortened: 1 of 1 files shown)


src/alire/os_linux/alire-platforms-current__linux.adb
```diff
@@ -20,7 +22,22 @@ package body Alire.Platforms.Current is
...
-      elsif not GNAT.OS_Lib.Is_Regular_File (OS_Identity_File) then
+      end if;
+
+      --  Try to get the distribution ID from an environment variable
+      --  Rationale: Map AlmaLinux, CentOS, Rocky Linux, Oracle Linux
+      --  et al to RHEL.
+
+      begin
+         Cached_Distro := Alire.Platforms.Distributions'Value
+           (Ada.Environment_Variables.Value ("ALIRE_DISTRIBUTION"));
```
mosteo, line 33: This `ALIRE_DISTRIBUTION` should be defined in `Alire.Environment`

src/alire/os_linux/alire-platforms-current__linux.adb
```diff
@@ -20,7 +22,22 @@ package body Alire.Platforms.Current is
...
+         Distro_Cached := True;
+         return Cached_Distro;
```
mosteo, line 35: In case of success, I'd like to have here a `Trace.Debug` report that the distro is being overridden via environment.

mosteo (changes requested):
Thanks for the contribution, @pmunts. A couple of minor changes requested before inclusion.

src/alire/os_linux/alire-platforms-current__linux.adb
Fabien-Chouteau: @mosteo Do we have a naming convention for variables that are inputs to the `alr` command vs variables that are set/defined by the `alr` command? I think it will get confusing otherwise.

mosteo: We don't have and this is something I wanted to revisit because the only other input env. var. that we have is `ALR_` instead of `ALIRE_` (which as long as it affects the library part and not only `alr` I guess should be used).

As for outputs, we are setting `<CRATENAME>_ALIRE_PREFIX`. Someone suggested and maybe it's clearer that we did the other way around, `ALIRE_PREFIX_<CRATENAME>` or whatever convention we adopt for "outputs"

mosteo: And we are also setting `ALIRE=True`

mosteo:
Superseded by #1610.

PR #1382, opened 2023-05-23
Fix unpacking of tarfiles with softlinks to subdirectories

mosteo:
Without this is impossible to install e.g. gcc 13.


Fabien-Chouteau (approved):
Looks good to me.

How is this related to GCC 13?

mosteo:
> How is this related to GCC 13?

The tarfile for gcc13 that I've been using (the one from Maxim IIRC) contains a softlink from a parent directory to a child one, something like:

parent/bits --> parent/x86_64/bits
parent/x86_64/bits

And our current tree remover is unable to remove that. Also, the contents of "parent/bits" are visited without "parent/bits" itself being ever visited (because it's a softlink), and that confuses the copy process.

It's a safeguard in our side, but the ideal situation would be that no softlinks existed in the tarfile.

PR #1383, opened 2023-05-26
More granular sharing of dependencies with `alr with --shared`

mosteo:
I apologize because this is one of those PRs where I got carried away with little tweaks that obfuscate a bit the actual changes.

Despite the long list of file changes, there is not much new logic here because the feature was already latent in the deployment of toolchains. It's mostly exposing it to the user and enabling it in the solver.

To simplify a bit some corner cases, I have homogenized that all external/system crates are implicitly shared, as they don't exist individually in a workspace cache, so they generate a folder containing their manifest in the shared location, as the rest of the shared binary crates already did. This also required some testsuite tweaking.

Another visible change is that, now that we have `alr install` meaning something entirely different, I've changed the few instances where we referred to a compiler as "installed" to "shared", in line with how it is named internally and how the regular crates are presented to the user.


doc/user-changes.md
```diff
@@ -6,6 +6,21 @@ stay on top of `alr` new features.
 
 ## Release `2.0-dev`
 
+### Selective sharing of dependencies with `alr with --shared`
+
+PR [#1383](https://github.com/alire-project/alire/pull/1383)
+
+As a more granular alternative to PR
+[#1367](https://github.com/alire-project/alire/pull/1367), a single dependency
+can now be shared using the new `--shared` switch of the `with` subcommand:
```
Fabien-Chouteau: I think this needs more information about what sharing a dependency means.

src/alr/alr-commands-withing.adb
```diff
@@ -357,6 +407,20 @@ package body Alr.Commands.Withing is
                 & " specified with --branch. Use `alr update` to refresh the"
                 & " tracking pin contents.")
        .New_Line
+       .Append ("* Adding dependencies with shared storage:")
```
Fabien-Chouteau: Is `alr with --shared libhello` also adding `libhello` as a dependency or only setting the expected sharing?

I think a dedicated command for this feature might make more sense.

testsuite/tests/with/shared-switch/test.py
```diff
@@ -0,0 +1,82 @@
...
+os.chdir(startdir)
+init_local_crate("yyy")
+run_alr("with", "libhello")
+match_deploy_dir("libhello", global_cache)
+reset()
+
+# Check that global caching can be prevented locally
+os.chdir(startdir)
+init_local_crate("yyy")
+run_alr("with", "--shared=no_local", "libhello")
```
Fabien-Chouteau: If I understand correctly, `with --shared=no_local  libhello` means that for this crate/workspace `libhello` should be deployed locally and not use the global shared deployment.

If I am correct, I think the command line is confusing as `no_local` means the "local" deployment will be used.

Fabien-Chouteau (changes requested):

src/alr/alr-commands-withing.adb
mosteo: It's currently doing both. I'm also not really happy with it, but couldn't find at the time an entirely satisfying interface.

One possibility is reusing `alr config`, which already has the `--global` needed to differentiate global/workspace settings. So it could be:

```
alr config --share[=yes|no|reset] <crate> [--global]
```

I'm open to suggestions.

testsuite/tests/with/shared-switch/test.py
mosteo: Ah, you're right. Indeed, the idea is that the 4 combos are [yes|no] x [locally|globally], but now I see how it's misleading.

src/alr/alr-commands-withing.adb
Fabien-Chouteau: `alr share [--global|--local] [--yes|--no] <crate>` ?

Default to `--local` and `--yes`.

And a `--info` to show the local and global shares.

mosteo:
I'm closing this one after our discussion. I'll review locally if there are any bits worth saving and open another one in that case.

PR #1398, opened 2023-06-29
Improvements to publishing with automatic PR creation

mosteo:
Add a few steps to the assistant that will end in the PR being opened without having to upload the manifest manually.

From the user PoV, this PR boils down to using `alr publish --submit`.

This is still lacking tests as we may want to tweak the workflow and in any case it will require some more instrumentation to allow dry runs and local impersonation of the community index, and there's already quite some amount of new code. So I would submit the tests in a follow-up.

Having now used the GitHub REST API, it will be easy to improve on this core functionality to allow status tracking and maybe some limited interaction (like requesting a review once checks pass) also from the command line (something like `alr publish --request-review | --status | --cancel`), so for the common case it could all be managed through `alr`. 

This should both simplify submissions and also some people have claimed for automation of the process (people already involved in packaging for distros).

For example, I created https://github.com/alire-project/alire-index/pull/840 with this.


Fabien-Chouteau (approved):
That looks promising :)

Two comments:
- I would love to see a comment explaining the whole process somewhere in the code. It looks fairly complex between the clone, commit and API calls.
- I think this should be a question at the end of `alr publish` rather than a switch.

PR #1399, opened 2023-07-03
Bits from the `alr share` feat to be kept

mosteo:
Refactorings and a few minor bits that will come in handy later on

### DIFF
--- a/src/alire/alire-config-edit.adb
+++ b/src/alire/alire-config-edit.adb
@@ -17,6 +17,10 @@ package body Alire.Config.Edit is
    type String_Access is access String;
    Config_Path : String_Access;
 
+   --  Separately loaded configs to be able to ascertain where a config value
+   --  is coming from.
+   Level_DB : array (Level) of CLIC.Config.Instance;
+
    -----------------
    -- Set_Locally --
    -----------------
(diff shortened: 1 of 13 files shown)


src/alire/alire-config-edit.adb
```diff
@@ -17,6 +17,10 @@ package body Alire.Config.Edit is
    type String_Access is access String;
    Config_Path : String_Access;
 
+   --  Separately loaded configs to be able to ascertain where a config value
+   --  is coming from.
+   Level_DB : array (Level) of CLIC.Config.Instance;
```
Fabien-Chouteau: So there are 3 DBs now [this one](https://github.com/alire-project/alire/blob/5c17abb0dc6ed48158ebd51750761317063b5ccd/src/alire/alire-config.ads#L8) and the two here? It seems overkill.

There's an origin field for every config value, we might be able to use that. Or we remove the common DB and only use the Level_DB with a getter function that will first look in local then global.

Fabien-Chouteau (approved):

src/alire/alire-config-edit.adb
mosteo: I see, I had not seen the origin field. Will try to use that.

Fabien-Chouteau (approved):

Issue #1403, opened 2023-07-14
MacPorts on macOS


simonjwright:
Having added support for the package manager [Homebrew](https://brew.sh), I'm tentatively considering the same for [MacPorts](https://www.macports.org). Would there be interest? What are the timescales for 2.0?

mosteo:
> What are the timescales for 2.0?

Not long now. Ideally release candidate in early august.

simonjwright:
Implemented in fa6c67d.

PR #1404, opened 2023-07-17
Support MacPorts as external provider.

simonjwright:
In response to issue #1403 - this was a lot easier than I’d expected.

Github CI for macOS has Homebrew installed by default. There is [unofficial Macports support](https://github.com/GiovanniBussi/macports-ci), which I haven’t investigated yet.

### DIFF
--- a/src/alire/alire-origins-deployers-system-macports.adb
+++ b/src/alire/alire-origins-deployers-system-macports.adb
@@ -1,0 +1,114 @@
+with AAA.Strings; use AAA.Strings;
+
+with Alire.OS_Lib.Subprocess;
+with Alire.Errors;
+
+package body Alire.Origins.Deployers.System.Macports is
+
+   --  Ada.Strings.Unbounded is use-visible via Alire.Origins.
+
+   package Subprocess renames Alire.OS_Lib.Subprocess;
+
+   -----------------------
+   -- Already_Installed --
+   -----------------------
+
+   overriding function Already_Installed (This : Deployer) return Boolean
+   is
+      Installed : AAA.Strings.Vector;
+   begin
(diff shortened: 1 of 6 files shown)


mosteo:
I see that Homebrew gets priority over MacPorts. As a MacOS ignoramus, I seem to remember that Homebrew requires activation, so if a user wanted to use MacPorts having both, there would be no trouble, right?

mosteo (approved):

simonjwright:
> I see that Homebrew gets priority over MacPorts. As a MacOS ignoramus, I seem to remember that Homebrew requires activation, so if a user wanted to use MacPorts having both, there would be no trouble, right?

It would be quite unusual to have both - I do, for the moment, so as to be able to test this change!

Both HB & MP need root to install the tool; after that, MP needs root (sudo) to install packages, HB doesn’t.

HB needs you to source a shell setup script which puts `brew` on your PATH; it also sets up 3 environment variables, including `HOMEBREW_PREFIX` which is the one I check - not sure what happens if `brew` is on `PATH` but the env vars aren’t set. Note, on Intel HB is in `/usr/local`, on Apple silicon in `/opt/homebrew`.

MP just needs `port` to be on the `PATH`; normally in `/opt/local` but apparently you can even have multiple versions!! No idea how `alr` would tell if this (very advanced) usage was in place - not that it’d affect `alr` itself, but of course include and library paths would have to account for it.

So, I have both set up; if I want `alr` to use MP, I unset `HOMEBREW_PREFIX`.

mosteo:
I see, thanks for the thorough explanation. As long as there is a way to override the priority that's hardcoded, I think it's fine.

PR #1408, opened 2023-07-21
dev: use env for executable shebangs

atalii:
Use `/usr/bin/env bash` instead of `/bin/bash` so that the scripts can run in more exotic environments, i.e., NixOS. Thanks!


mosteo (approved):

mosteo:
Merged, thanks.

PR #1413, opened 2023-07-27
Fix nightly build

mosteo:
At some point I got ahead of myself and introduced a circularity where the nightly build must exist before a new one can be produced. So, the first time a nightly build failed because of some network glitch, it couldn't succeed anymore.


mosteo:
I'm merging this one without review as it's a bit urgent because the `alr_install` action depends on it, and we use it in several places.

PR #1419, opened 2023-08-03
New shared builds mechanism (mockup)

mosteo:
This introduces most infrastructure needed for the new build system. It's still disabled by default (although some tests check it already) because it's missing the hash computation, which will come in a follow-up PR. 

Instead of the real hash, the same unique fake hash is always used. Hence, enabling it now roughly works like the previous `dependencies.dir` worked, as there will be only one build config per release, except that there are already separate locations for "read-only" sources and build sources.

There are also some assorted TODOs through the code for future optimizations, but the functional part is complete (again, but for the hash computation).

A summary of the new modus operandi is in `alire-builds.ads`.


doc/user-changes.md
```diff
@@ -6,6 +6,35 @@ stay on top of `alr` new features.
...
+- `$HOME/.cache/alire/toolchains`: contains downloaded toolchains.
+
+The previous `$HOME/.cache/alire/dependencies` that contained both toolchains
+and binary releases is no longer in use.
+
+Users wanting to modify dependencies in tandem within a workspace are
+encouraged to use the pin system.
+
+If these default locations for shared dependencies must be relocated, this can
+be achieved by using a new configuration path (`ALR_CONFIG` or `-c` global
```
Fabien-Chouteau, line 34: It could be useful to have a separate option/env-var for the cache.

src/alire/alire-builds.adb
```diff
@@ -0,0 +1,135 @@
...
+   ----------------------------
+
+   function Sandboxed_Dependencies return Boolean
+   is (not Config.DB.Get (Config.Keys.Dependencies_Shared,
+                          Config.Defaults.Dependencies_Shared));
+
+   -------------------
+   -- To_Msys2_Path --
+   -------------------
+   --  Convert C:\blah\blah into /c/blah/blah. This is needed because otherwise
```
Fabien-Chouteau, line 26: We should try to avoid msys2 specific behaviors like that. And that probably means removing the dependency on an `rsync` executable.

src/alire/alire-builds.adb
```diff
@@ -0,0 +1,135 @@
...
+            --  Archive, no CVS folders, keep hard links, human units
+            & (if Log_Level > Detail then To_Vector ("-P") else Empty_Vector)
+            & (if Log_Level < Info   then To_Vector ("-q") else Empty_Vector)
+            --  Trailing '/' to access contents directly in the following paths
+            & To_Msys2_Path (Src / "")
+            & To_Msys2_Path (Dst / "")
+           );
+         --  TODO: this may take some time, and rsync doesn't have a way
+         --  to show oneliner progress, so at some point we may want to use
+         --  something like GNAT.Expect for our spawns, and spin on newlines
```
Fabien-Chouteau, line 95: Or the Spawn library https://github.com/adacore/spawn

Fabien-Chouteau (approved):

src/alire/alire-builds.adb
setton: either spawn or [GNATCOLL.OS.Process](https://github.com/AdaCore/gnatcoll-core/blob/master/src/os/gnatcoll-os-process.ads) - spawn is higher level and suited for integration in a main loop. GNATCOLL.OS.Process is lower level and leaves the user more things to do, but is an easier in-place replacement for GNAT.Expect. Both are superior to GNAT.Expect since they allow separating stdout and stderr, and allow controlling the environment of the spawned process without affecting the parent's environment.

src/alire/alire-builds.adb
mosteo: Actually rsync is probably overkill since the new folder should either exist or not at all, so a full copy is OK. I was for some reason too focused on using rsync. I take note for a follow-up PR.

doc/user-changes.md
mosteo: I'll create an issue to track this.

PR #1420, opened 2023-08-05
MacOS; export PATH variables for includes and libraries.

simonjwright:
Homebrew and MacPorts install include files and libraries in places where GCC won't look by default.

GCC will use these environment variables if set:

   C_INCLUDE_PATH     for C includes
   CPLUS_INCLUDE_PATH for C++ includes
   LIBRARY_PATH       for libraries

Both of the distribution managers place (symbolic links to) include files in ${top_level}/include and libraries in ${top_level}/lib.

   For Homebrew on Intel silicon, top_level is normally /usr/local.
   For Homebrew on Apple silicon, top_level is normally /opt/homebrew.
   For MacPorts,                  top_level is normally /opt/local

  * src/alire/alire-platforms-current.ads (Load_Environment): add note on macOS use.
  * src/alire/os_macos/alire-platforms-current__macos.adb (context): added Alire.Environment (was limited), Ada.Directories. (Brew_Access): new. (Homebrew_Present): if Brew_Access is not null. (Detected_Distribution): made into an expression function. (Containing_Containing_Dir): new, used in Distribution_Root. (Distribution_Root): reworked. (Load_Environment): if either distribution is present, arrange to export the environment variables to suit.

### DIFF
--- a/.github/workflows/ci-macos.yml
+++ b/.github/workflows/ci-macos.yml
@@ -33,17 +33,8 @@ jobs:
       with:
         python-version: '3.x'
 
-    - name: Run test script (without Homebrew)
+    - name: Run test script
       run: scripts/ci-github.sh
-      shell: bash
-      env:
-        BRANCH: ${{ github.base_ref }}
-        INDEX: ""
-
-    - name: Run test script (with Homebrew)
-      run: |
-        eval $(brew shellenv)
-        scripts/ci-github.sh
       shell: bash
       env:
         BRANCH: ${{ github.base_ref }}
--- a/src/alire/alire-platforms-current.ads
+++ b/src/alire/alire-platforms-current.ads
@@ -19,7 +19,8 @@ package Alire.Platforms.Current is
 
    procedure Load_Environment (Ctx : in out Alire.Environment.Context);
    --  Set environment variables from the platform. Used by Windows to
-   --  initialize msys2 environment.
+   --  initialize msys2 environment, and by macOS to initialize which,
+   --  if either, of the Homebrew or MacPorts environment.
 
    -----------------------
    --  Self identification
--- a/src/alire/os_macos/alire-platforms-current__macos.adb
+++ b/src/alire/os_macos/alire-platforms-current__macos.adb
@@ -1,19 +1,23 @@
+with Alire.Environment;
 with Alire.OS_Lib;
+with Ada.Directories;
 with GNAT.OS_Lib;
 
 package body Alire.Platforms.Current is
 
    --  macOS implementation
 
+   use type GNAT.OS_Lib.String_Access;
+
    --  Homebrew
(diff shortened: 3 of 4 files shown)


simonjwright:
The [problem with the macOS tests](https://github.com/alire-project/alire/actions/runs/5771022704/job/15644588450#step:5:1391) is this difference:
```
-Ada -> Alire_Host_Distro: distro_unknown
+Ada -> Alire_Host_Distro: homebrew
```
The reason is that I changed the detection of Homebrew from looking for `$HOMEBREW_PREFIX` to checking whether `brew` is on the `PATH`, and Github CI loads Homebrew automatically (into `/usr/local` on Intel silicon, which is why we now detect it).

Sorry about that. What should I do?

Fabien-Chouteau:
Hi @simonjwright, the testsuite also has a distrib detection system: https://github.com/alire-project/alire/blob/9b476d0ea62e8ce72a97c602ecd5f5e650724fc7/testsuite/drivers/helpers.py#L100

It should be adjusted to match the Alire one.

simonjwright:
Can this be reviewed, please?

.github/workflows/ci-macos.yml
```diff
@@ -33,22 +33,13 @@ jobs:
...
         INDEX: ""
 
-    - name: Run test script (with Homebrew)
-      run: |
-        eval $(brew shellenv)
-        scripts/ci-github.sh
-      shell: bash
-      env:
-        BRANCH: ${{ github.base_ref }}
-        INDEX: ""
```
mosteo, line 50: So I understand that now tests run by default without either hombrew or macports, unless we purposely activate one of those in the tests themselves?

mosteo (approved):

mosteo:
Sorry for the wait, I had a concatenation of busy days.

.github/workflows/ci-macos.yml
simonjwright: The tests now run _with_ Homebrew (`alr` now checks whether `brew` is in the path, rather than whether `HOMEBREW_PREFIX` is set; the extra in the 'with Homebrew' part of the script was to set it).

I didn’t make a 'with MacPorts' version of the tests because CI automatically installs Homebrew; I’m sure we could install MacPorts (but needs sudo), but not sure about uninstalling Homebrew. For my purposes, I could just `sudo rm $(which brew)` - would that work? would you like another PR to do that?

I could reverse the test order so that, if both are present, MacPorts takes precedence, then wouldn’t need to `rm brew`. Grr, but if needs must ...

mosteo: I see, thanks for the explanation. I'm going to merge this one for now. I'm not sure how critical is to test the three situations; it might be overkill (for example we don't test Windows without msys2 IIRC).

PR #1425, opened 2023-08-06
Use build profile in build hash

mosteo:
This would be the first realistic albeit incomplete step towards the shared builds. This adds the actual computation of the hash, but for the moment relying only on the build profile, so we can discuss any changes before adding the remaining factors.

To have some traceability, the inputs to the hash function are stored in `cache/builds/<crate>/alire/build_hash_inputs` and so it also can be used in tests to verify the hash computation.

### DIFF
--- a/src/alire/alire-builds-hashes.adb
+++ b/src/alire/alire-builds-hashes.adb
@@ -108,7 +108,8 @@ package body Alire.Builds.Hashes is
          Trace.Debug ("   build hashing: " & Rel.Milestone.TTY_Image);
 
          --  Build profile
-         Add ("profile", "profile",
+         Add ("profile",
+              Rel.Name.As_String,
               Root.Configuration.Build_Profile (Rel.Name)'Image);
 
          --  GPR externals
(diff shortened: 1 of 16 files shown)


src/alire/alire-builds-hashes.adb
```diff
@@ -0,0 +1,155 @@
...
+            Append_Lines (File,
+                          Lines,
+                          Backup => False);
+         end Write_Inputs;
+
+      begin
+         Trace.Debug ("   build hashing: " & Rel.Milestone.TTY_Image);
+
+         --  Build profile
+         Add ("profile",
```
Fabien-Chouteau, line 113: In the end we my not have "build_profile" in the hash since there is a separate object dir depending on the profile, at least in Alire generated gpr files.

Fabien-Chouteau (approved):

src/alire/alire-builds-hashes.adb
mosteo: Alright, I'm going to merge it as is for now, as it is trivial to remove later on.

Issue #1417, opened 2023-08-02
CI fails in Install FSF Toolchain


simonjwright:
See [this run](https://github.com/simonjwright/alire/actions/runs/5737861391/job/15550313359), which was a check on a proposed patch to address #1410.

For info, the master branch that I was running the PR against was up-to-date with alire-project/alire:master.

Also, I see that the docker workflow ran OK.

mosteo:
Is this reproducible? I've seem some iffy behavior from GH actions from time to time (this doesn't seem one of those, but just in case).

simonjwright:
I just re-ran the linux & macos workflows, with the same result in each case (and the same failure); starting at [line 200](https://github.com/simonjwright/alire/actions/runs/5737861391/job/15587992479#step:3:215),
```
Note: Installing gnat_native=12.2.1...
ERROR: directory tree rooted at "/home/runner/work/alire/alire/alire_prefix/gnat_native_12.2.1_11f3b811/lib64" could not be deleted
ERROR: alr encountered an unexpected error, re-run with -d for details.
Error: Process completed with exit code 1.
```

May I try a PR to the project repo? (I was hoping to fix any errors before the real PR)

simonjwright:
I decided to raise the PR against the alire repo, which only failed on macOS (because of a side-effect of this change); other workflows OK.

I recently tried a similar scheme (try a PR on my own fork), which failed similarly, and worked fine on the alire repo, so I don’t think I’ll be trying that again.

No idea why :-(

mosteo:
I'm starting to see this in my own workflows for different distros. I'll try to add some debug output to that deletion routine.

PR #1426, opened 2023-08-07
Debug when folder deletion fails

mosteo:
Extra output when a file/folder deletion fails, to try and debug occurrences like #1417.


src/alire/alire-directories.adb
```diff
@@ -253,6 +253,28 @@ package body Alire.Directories is
...
+      when E : others =>
+         Trace.Error ("Could not delete: " & Path);
+         Log_Exception (E, Error);
+         Trace.Error ("Contents follow: ");
+         declare
+            use AAA.Strings;
+            Output : Vector;
+            Code : constant Integer :=
+                     OS_Lib.Subprocess.Unchecked_Spawn_And_Capture
+                       ("ls", To_Vector ("-alRF"),
```
Fabien-Chouteau: I don't think it's good to rely on and external `ls` command here. On Windows it might not be available.

Fabien-Chouteau (changes requested):

mosteo:
I've added an alternative for Windows. Nonetheless, this is kind of a brute attempt to diagnose exactly the failures happening here... Other ideas welcome :S

mosteo:
I know what the root cause is now and can reproduce locally, working on the fix.

mosteo:
The CI error in this one will go away after #1427

Fabien-Chouteau (approved):

PR #1428, opened 2023-08-21
Add GPR externals and env. vars. to build hash computation

mosteo:
This turned out a bit uglier than anticipated because when hashing the environment being set by the crate, it may depend on the own crate dir (via `$CRATE_DIR`), which in turn depends now on the build hash being computed. 

So we have to break the circularity and skip these substitutions when we are still computing the build hash.


src/alire/alire-builds-hashes.adb
```diff
@@ -115,14 +119,45 @@ package body Alire.Builds.Hashes is
...
-         --  TBD
+         declare
+            Externals : constant Releases.Externals_Info := Rel.GPR_Externals;
+         begin
+            for Var of GPR.Name_Vector'(Externals.Declared
+                                .Union (Externals.Modified))
+              --  Externals modified but not declared are presumably for the
+              --  benefit of another crate. It's unclear if these will affect
+              --  the crate doing the setting, so we err on the side of
+              --  caution and include them in the hashing. Maybe we could make
```
Fabien-Chouteau, line 130: I think it's good to include them.

src/alire/alire-builds-hashes.adb
```diff
@@ -115,14 +119,45 @@ package body Alire.Builds.Hashes is
...
+         end;
 
          --  Environment variables
          --  TBD
 
          --  Configuration variables
          --  TBD
 
+         --  Dependencies recursive hash? Since a crate can use a dependency
+         --  config spec, it is possible in the worst case for a crate to
```
Fabien-Chouteau, line 151: Indeed, I think that's something I already do in my project.

Fabien-Chouteau (approved):

PR #1430, opened 2023-08-25
Include compiler version in build hash

mosteo:


src/alire/alire-builds-hashes.adb
```diff
@@ -147,6 +147,15 @@ package body Alire.Builds.Hashes is
          --  Configuration variables
          --  TBD
 
+         --  Compiler version. Changing compiler will result in incompatible
+         --  ALI files and produce rebuilds, so it must be part of the hash.
+         --  Incidentally, this serves to separate by cross-target too.
```
Fabien-Chouteau, line 152: The compiler is able to detect incompatible build artifact, but it's probably better to avoid the error message.

Also note that this will not handle all case for cross-target since a given cross compiler can support multiple instruction sets. For instance `arm-elf` can do Arm Cortex-M0, Cortex-M7, etc.

Fabien-Chouteau (approved):

src/alire/alire-builds-hashes.adb
mosteo: > Also note that this will not handle all case for cross-target since a given cross compiler can support multiple instruction sets. For instance `arm-elf` can do Arm Cortex-M0, Cortex-M7, etc.

Right; for those I understood that we're not going to do anything directly, but that the project file will separate the objects into folders?

I guess at some point we can use `libgpr` to extract the necessary info from the project file?

PR #1440, opened 2023-09-01
Add some hardcoded externals to hash input

mosteo:
These are used but never declared. I'm not 100% sure this is needed.


Fabien-Chouteau (approved):
In theory they produce different outputs, but it could re-trigger the builds of all the units so it's probably good to include them.

PR #1444, opened 2023-09-04
New `alr printenv --last-build`

mosteo:
This may be useful when no direct editing of manifests is convenient, as otherwise `printenv` will always use the manifest's configuration.

### DIFF
--- a/src/alr/alr-commands-printenv.adb
+++ b/src/alr/alr-commands-printenv.adb
@@ -1,3 +1,4 @@
+with Alire.Crate_Configuration;
 with Alire.Environment;
 with Alire.Platforms;
 
@@ -28,6 +29,11 @@ package body Alr.Commands.Printenv is
       end if;
 
       Cmd.Requires_Workspace;
+
+      if Cmd.Last_Build then
+         Cmd.Root.Set_Build_Profiles
+           (Alire.Crate_Configuration.Last_Build_Profiles);
+      end if;
 
       declare
          Context : constant Alire.Environment.Context :=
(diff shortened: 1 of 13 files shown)


src/alr/alr-commands-printenv.adb
```diff
@@ -89,6 +95,10 @@ package body Alr.Commands.Printenv is
                      Cmd.Cmd_Shell'Access,
                      "", "--wincmd",
                      "Use a Windows CMD shell format for the export");
+      Define_Switch (Config,
+                     Cmd.Last_Build'Access,
+                     "", "--last-build",
+                     "Use last build profiles instead of manifest profiles");
```
Fabien-Chouteau: I don't understand the use for this switch.

mosteo: With the new shared builds the actual paths change depending on the build profile. 

Now, `alr printenv` relies on the info in the manifests, always printing the same info. With this, one could do
```
alr build --profiles=*=validation
alr printenv --last-build
```
and get the paths for validation builds.

Fabien-Chouteau: I would consider this to be the default behavior. We already save the last build profile somewhere, right?

mosteo: Yes, they're saved after every build. That's what's being used to set the profiles here indeed.

We could make `alr printenv` understand the `--profiles` switch too, in case one wants to get a particular environment without forcing a build in advance.

mosteo:
Modified to make this the default behavior

Fabien-Chouteau (approved):

PR #1447, opened 2023-09-09
Auto-update index periodically

### LINKED ISSUE #1348: Improve user experience with regards to index update
We have another instance of a user having a  bad experience because their index is not up-to-date: https://old.reddit.com/r/ada/comments/11r2kl7/alire_inability_to_install_some_packages_from_the/

Some ways to deal with this issue involve automatic internet access to check if index updates are available. I don't think this is a good thing to do all the time.

But we have other options:
1. Print a message that says "try to update the index..." when a `get`/`search`/`with` command fails. This means no internet access, but also it could be easy to miss and won't work for new versions of existing crates.
2. Query the user before running `get`/`search`/`with`/`update` commands whether or not they want to update the indexes. This question would default to Yes and with the option to answer Always/Never. I think this would be a good compromise. The cost of an update on my machine is less than 1 second, which is reasonable when running the commands in question.

mosteo:
Default every 24h. The user is asked the first time. 

Fixes #1348

### DIFF
--- a/doc/user-changes.md
+++ b/doc/user-changes.md
@@ -6,6 +6,18 @@ stay on top of `alr` new features.
 
 ## Release `2.0-dev`
 
+### Automatic index updates
+
+PR [#1447](https://github.com/alire-project/alire/pull/1447)
+
+A new configuration option, `index.auto_update`, allows setting the refresh
+period of indexes. It defaults to 24 hours and the user will be asked the first
+time to allow automatic updates. Setting this option to 0 will also disable
+automatic updates.
+
+When enabled, updates may happen before executing commands that rely on
+indexes: `get`, `search`, `with`, etc.
+
 ### Deprecation of `dependencies.dir` in favor of `dependencies.shared`
 
 PR [#1419](https://github.com/alire-project/alire/pull/1419)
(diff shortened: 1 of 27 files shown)


Fabien-Chouteau (commented):
I see why you want to do this, we regularly have questions from people that never update the index. However I am a but worried about the consequences :)

What happens if there's no internet connection?

mosteo:
Actually I thought that you also wanted it :) given the discussion in #1348.

Yup, I forgot to test what happens in the offline case. We should fail the update gracefully with a warning and continue as before. We probably don't want to try again immediately but back-off for the configured period in any case. I'll do a bit more testing.

rowan-walshe:
I like the change. I was once on of those that had to ask why alr couldn't find a newer version of a crate that I could see in the index on GitHub.

One comment I have is that while `alr update --online` can be used to force update the index(es) before attempting crate updates, it feels like it should still call `Auto_Update_Index` if `--online` isn't set

mosteo:
Updated to ensure that an automatic update failure causes the regular backoff, and is handled as a warning.

Also added a test for it. I didn't find a simple solution to make the test locally, so that test uses an [online repo](https://github.com/alire-project/test-index) I created to that effect. There may be solutions around that by using a local git server, but that seemed too complex just to avoid going online, given that that's not a crucial situation for our testing.

Fabien-Chouteau (approved):

PR #1457, opened 2023-09-17
Print UTF-8 string using `GNAT.IO` in `Report_Jekyll`

### LINKED ISSUE #1456: `alr show --jekyll zeromq_ada` prints garbage from special character and truncates output
The development version of `alr` has this problem. `alr` v1.2.1 works fine.

```
$ alr show --jekyll zeromq_ada
---
layout: crate
crate: "zeromq_ada"
[...uninteresting...]
"binding"]
version: "4.1.5"
short_description: "An Ada-binding the Ã$
``` 

Input is "An Ada-binding the ØMQ library"

So it is failing in the Ø character.

mgrojo:
Fixes #1456

### DIFF
--- a/alire_common.gpr
+++ b/alire_common.gpr
@@ -81,12 +81,18 @@ abstract project Alire_Common is
    end Compiler;
 
    package Builder is
-      for Switches ("Ada") use ("-s", "-j0", "-g");
+      for Switches ("Ada") use
+        ("-s", -- Recompile if switches changed
+         "-j0" -- Full parallelism
+        );
    end Builder;
 
    package Binder is
       for Switches ("Ada") use
-        ( "-Es", "-g", "-static");
+        ("-Es",    -- Symbolic tracebacks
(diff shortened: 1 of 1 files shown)


mgrojo:
Another possibility is changing this renames to `GNAT.IO`, but I didn't know the implications. Maybe you think it's better because it could solve other UTF-8 problems.

https://github.com/alire-project/alire/blob/8243841325be7938a7f3e6aafa311b335cf2d9a1/src/alr/alr-commands.ads#L128-L129

mosteo:
Thanks for spotting this, @mgrojo . No matter the implications, the proper way to output UTF-8 is unencoded via Wide_Wide_Text_IO, or encoded directly through GNAT.IO. 

Text_IO should not be used and this renaming slipped through the cracks during #972, #1220, #1332.

So I'd ask you if it's not too much trouble to fix the rename instead of using explicitly GNAT.IO.

PR #1463, opened 2023-09-22
Remove ".git" after cloning, with option to disable

### LINKED ISSUE #1455: "un-git" the releases deployed from git repos
What I mean here is to **not** have releases deployed from git as valid git checkouts.

This idea starts with a discussion with a user that was confused by Alire's behavior. This person wanted to use `alr get` as a way to checkout a repository to work on it, and was surprised by the fact that `alire.toml` was modified (see why here: https://github.com/alire-project/alire/issues/1385#issuecomment-1716222070).

For me the disconnect is in the expectation that `alr get` gives when the source of the release is a git repo. Indeed you get a git checkout, but it's a release not a development branch.

What I want to suggest here is to change this behavior and not have releases as valid git checkout. The only way I found to do this is to remove the `.git` folder after `git clone` .

mosteo:
Also make clones shallow (they already were for pins).

### DIFF
--- a/src/alire/alire-config-builtins.ads
+++ b/src/alire/alire-config-builtins.ads
@@ -6,6 +6,15 @@ package Alire.Config.Builtins is
    -- Builtins --
    --------------
 
+   --  DEPENDENCIES
+
+   Dependencies_Git_Keep_Repository : constant Builtin := New_Builtin
+     (Key  => "dependencies.git.keep_repository",
+      Def  => False,
+      Help =>
+        "When true, git origins are a proper git repository after deployment. "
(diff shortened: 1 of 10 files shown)


Fabien-Chouteau (commented):
Thanks @mosteo,

What is going to happen for pins? Because it should not remove `.git` for pinned dependencies in my opinion.

mosteo:
> What is going to happen for pins? Because it should not remove `.git` for pinned dependencies in my opinion.

Those use their own cloning logic, and remain as (shallow) repositories.

Fabien-Chouteau (approved):

PR #1475, opened 2023-10-19
Separate AppImage to own workflow

mosteo:
This will give us advance warning in case something breaks between releases (as has happened for some obsolete actions) and will allow using an older base OS for the appImage, increasing back-compatibility.


mosteo:
I'm merging directly as this doesn't touch actual code.

PR #1479, opened 2023-10-20
implement -C/--chdir flag

### LINKED ISSUE #1478: add change directory option to Alire commands
Alire doesn't support changing directories as part of the command (similar to make -C and git -C). Consequently, we must cd into the directory to run Alire commands. It would be nice to be able to specify the directory to run the Alire command. This is especially useful for CI / other automation systems.

atalii:
Fixes #1478.

This is my first contribution to any actual code, so expect it to be rough. I imagine for a start we'd want tests for this, as well as actual error messages.

### DIFF
--- a/src/alr/alr-commands.adb
+++ b/src/alr/alr-commands.adb
@@ -148,7 +148,7 @@ package body Alr.Commands is
       Define_Switch (Config,
                      Command_Line_Chdir_Target_Path'Access,
                      "-C=", "--chdir=",
-                     "Run Alire in the given directory");
+                     "Run `alr` in the given directory");
 
       Define_Switch (Config,
                      Alire.Force'Access,
--- a/src/alr/alr-commands.adb
+++ b/src/alr/alr-commands.adb
@@ -59,6 +59,8 @@ package body Alr.Commands is
 
    Command_Line_Config_Path : aliased GNAT.OS_Lib.String_Access;
 
+   Command_Line_Chdir_Target_Path : aliased GNAT.OS_Lib.String_Access;
+
    --  Following aliased booleans are used by GNAT.Command_Line processing:
 
    Log_Quiet  : Boolean renames Alire_Early_Elaboration.Switch_Q;
@@ -142,6 +144,11 @@ package body Alr.Commands is
                      Command_Line_Config_Path'Access,
                      "-c=", "--config=",
                      "Override configuration folder location");
+
+      Define_Switch (Config,
+                     Command_Line_Chdir_Target_Path'Access,
+                     "-C=", "--chdir=",
+                     "Run Alire in the given directory");
 
       Define_Switch (Config,
                      Alire.Force'Access,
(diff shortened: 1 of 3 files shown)


src/alr/alr-commands.adb
```diff
@@ -143,6 +145,11 @@ package body Alr.Commands is
                      "-c=", "--config=",
                      "Override configuration folder location");
 
+      Define_Switch (Config,
+                     Command_Line_Chdir_Target_Path'Access,
+                     "-C=", "--chdir=",
+                     "Run Alire in the given directory");
```
mosteo: To refer to the command-line tool  proper, better use `alr` here.  (Backticks included as this helps with doc generation later on).

mosteo (changes requested):
This looks good to me, but indeed we would need a new test. For this feature it should be quite simple and I can guide you. First thing is, are you able to run the testsuite now? You can get started with the README in the `testsuite` folder.

src/alr/alr-commands.adb
atalii: 10b7716 - thanks!

atalii:
I got the test suite running pretty painlessly, and attempted to write a test in fda16e3. I copied `init__default-executable` and got rid of most of it so as only to test `--chdir`. Does that look good? I don't think I'd know if I'm missing something, but it seems to work. Thank you so much!

testsuite/tests/init/with-chdir/test.py
```diff
@@ -0,0 +1,20 @@
+"""
+Test "executable" only appears in --bin initializations
+"""
```
mosteo: Quite common to forget to update this description ;-)

mosteo (changes requested):
Looks good to me, just a minor thing to fix.

testsuite/tests/init/with-chdir/test.py
```diff
@@ -0,0 +1,20 @@
...
+from drivers.helpers import content_of
+
+test_dir = os.getcwd()
+
+# Binary crate
+run_alr("init", "--bin", "xxx")
+assert_match(".*executables = \[", content_of("xxx/alire.toml"))
+
+# Check that it builds and runs
+run_alr("--chdir=xxx", "run")
```
mosteo: Also, it would be good to have a check after this one that when the target directory doesn't exist, alr exits with error.

testsuite/tests/init/with-chdir/test.py
atalii: Thanks for catching that - fixed in cba2c11, which also tests -C behavior on a non-existent directory. Thanks!

mosteo (approved):

mosteo:
Thanks, @atalii

Fabien-Chouteau:
Hi @atalii, it would be nice to have a user "User-facing changes" entry for this: https://github.com/alire-project/alire/blob/master/doc/user-changes.md

PR #1508, opened 2023-11-15
Offer uninstalled compiler as fallback with warning

mosteo:
This one may be controversial.

Currently, we will not try to satisfy a `gnat` dependency with an uninstalled compiler (specific dependencies like `gnat_arm_elf` do not fall into this case, only the abstract `gnat` ones), instead reporting a missing dependency:

```
$ alr with gnat^12
Requested changes:

   ✓ gnat ^12 (add)
                                                                         
   New solution is incomplete.
   +↪ gnat ^12 (new,external)

Do you want to proceed?
[Y] Yes  [N] No  (default is No) 
```

This doesn't seem very intuitive, as nothing says that installing the proper compiler version would fix the issue. With this PR an installed compiler is still chosen first for the dependency but, when none fits, this will happen:
```
$ alr with gnat^12
Requested changes:

   ✓ gnat ^12 (add)
                                                                         
   +📦 gnat 12.2.1 (new,gnat_native,binary)
   
   ❗ The solution requires the download of
      a toolchain not currently installed.

Do you want to proceed?
```

### DIFF
--- a/src/alire/alire-solutions-diffs.adb
+++ b/src/alire/alire-solutions-diffs.adb
@@ -404,14 +404,18 @@ package body Alire.Solutions.Diffs is
                Trace.Log (Prefix, Level);
                Trace.Log
                  (Prefix & Icon (Missing)
-                  & " The solution requires the (potentially large) ");
+                  & " The solution requires the download of");
                Trace.Log
                  (Prefix
-                  & "  download of a toolchain not currently installed.");
+                  & "   a toolchain not currently installed.");
                return;
             end if;
          end loop;
       end Warn_Toolchain_Download;
+
+      --------------------------------------
+      -- Warn_Unsatisfiable_GNAT_External --
+      --------------------------------------
 
       procedure Warn_Unsatisfiable_GNAT_External is
       begin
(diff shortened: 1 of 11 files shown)


Fabien-Chouteau:
Hi @mosteo ,

> ❗ The solution requires the download of
>       a toolchain not currently installed.

I am not sure about this error message because it doesn't provide a way forward.
In which situations a user would get this message?

mosteo:
> Hi @mosteo ,
> 
> > ❗ The solution requires the download of
> > a toolchain not currently installed.
> 
> I am not sure about this error message because it doesn't provide a way forward. In which situations a user would get this message?

The way forward is simply accepting the solution, and the dependency will be installed normally as a toolchain (as happens currently for explicit compiler dependencies like `gnat_native`. The message could be rephrased, of course.

This situation happens when a solution (`alr with`, `alr get`) requires necessarily a compiler that isn't already installed.

Fabien-Chouteau:
> The way forward is simply accepting the solution, and the dependency will be installed normally as a toolchain (as happens currently for explicit compiler dependencies like `gnat_native`. 

Ok so right now Alire would just show an incomplete solution, but not download the compiler even if it knows how to. 
And with this change Alire will download the compiler?

> The message could be rephrased, of course.

To me this message is telling the user that something went wrong and there's nothing Alire can do about it.

So maybe we should remove the exclamation mark and the message could be `The solution requires a toolchain that is not installed yet. Alire with now proceed to download and install this toolchain.`

mosteo:
Yeah, that's the gist of it.

mosteo:
New message:
```

   +📦 gnat 12.2.1 (new,gnat_native,binary)
   
   🛈  The solution requires a toolchain that is
      not yet installed. Accepting the solution
      will download and install this toolchain.
```

Fabien-Chouteau (approved):

PR #1510, opened 2023-12-02
platforms: use $XDG_DATA_HOME for cache on linux/freebsd/macos

### LINKED ISSUE #1502: Should '.cache/alire' be '.local/alire'?
Some say the former is for short-lived info and might even be mounted as tmpfs.

atalii:
$XDG_CACHE_HOME may be, in both specification and application, mounted on a tmpfs, removed frequently, &c. It generally appears to be used for small amounts of unimportant data, which is not what Alire expects of it. This commit switches things over to $XDG_DATA_HOME, which is generally more reliable.

Fixes #1502. (It looks discussion on that issue reached consensus; no worries if there's still debate/concern!)

### DIFF
--- a/src/alire/alire-platforms-common.ads
+++ b/src/alire/alire-platforms-common.ads
@@ -32,13 +32,13 @@ private package Alire.Platforms.Common is
                                                 Default => ".")));
 
    ----------------------
-   -- XDG_Cache_Folder --
+   -- XDG_Data_Folder --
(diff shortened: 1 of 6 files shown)


mosteo:
Thanks, @atalii, this looks good, but I still have to catch up with #1502.

mosteo:
Superseded by #1528, thanks @atalii

PR #1519, opened 2023-12-18
Build appimage on Ubuntu 20.04

### LINKED ISSUE #1511: v2.0.0-beta1 binaries fail on Linux distributions with older libc.so.6
When attempting to run the AppImage binary on **openSUSE Leap 15.5**, it fails with the following messages:
```
./alr-2.0.0-beta1-x86_64.AppImage: /lib64/libc.so.6: version `GLIBC_2.34' not found (required by ./alr-2.0.0-beta1-x86_64.AppImage)
./alr-2.0.0-beta1-x86_64.AppImage: /lib64/libc.so.6: version `GLIBC_2.32' not found (required by ./alr-2.0.0-beta1-x86_64.AppImage)
./alr-2.0.0-beta1-x86_64.AppImage: /lib64/libc.so.6: version `GLIBC_2.33' not found (required by ./alr-2.0.0-beta1-x86_64.AppImage)
```
When attempting to run the x86_64-linux binary on **openSUSE Leap 15.5**, it fails with similar messages:

```
./alr: /lib64/libc.so.6: version `GLIBC_2.34' not found (required by ./alr)
./alr: /lib64/libc.so.6: version `GLIBC_2.32' not found (required by ./alr)
./alr: /lib64/libc.so.6: version `GLIBC_2.33' not found (required by ./alr)
```

Similar messages are seen on **Ubuntu 20.04 LTS** for both the AppImage and x86_64-linux binaries.

serock:
Besides changing the versions of Ubuntu and GNAT, the only other change is to get the build to use GCC 10 instead of GCC 9. The `$HOME/.local/bin` directory is already in the `PATH` environment variable before the `/usr/bin` directory. The changes create the `$HOME/.local/bin` directory and create a `gcc` symbolic, in that directory, that points to gcc-10.

Fixes #1511.

### DIFF
--- a/.github/workflows/ci-appimage.yml
+++ b/.github/workflows/ci-appimage.yml
@@ -18,7 +18,7 @@ jobs:
 
     # Oldest supported by Alire+GitHub to increase AppImage back-compatibility
     # Unfortunately we depend on the static elaboration model of recent GNATs
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-20.04
 
     steps:
     - name: Check out repository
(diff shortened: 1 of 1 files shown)


mosteo (approved):

mosteo:
Thanks, @serock

PR #1522, opened 2023-12-31
Report the name of an unreadable directory.

simonjwright:
While working on a new external, I tried adding my local alire-index before the community index for testing. The error reported was "Not a readable directory". After 2 hours investigation (gdb ...) I realised that I’d been trying to add `alr-index` rather than `alire-index`.

This patch (which is in the style of other `Outcome_Failure` messages) would probably have saved that time!


simonjwright:
Sorry, I’ve no idea how to fix this.

simonjwright:
Sorry, this is just not worth it. Abandoned.

PR #1525, opened 2024-01-05
Add missing with statements for Ada.Directories

rowan-walshe:
Adds two with statements for Alire.Directories.

A compiler bug was giving visibility of some packages in cases where it shouldn't have. After the bug is fixed in a future version, Alire will fail to compile without these with statements.


mosteo (approved):

mosteo:
Thanks, @rowan-walshe

PR #1530, opened 2024-01-16
Tag builds done through `alr build` with unique build string

mosteo:
This may help down the line with reports of people using development versions, to uniquely identify them.

A script is invoked before build to patch in the commit (and an extra flag if the working folder is dirty). 

After the build, "_or_later" is added to the build string, so we have a lower bound, but we are not mislead by manual builds that don't trigger the patching (via pure `gprbuild`).

So, the new versions would look like `2.0-dev+abcd1234` or `2.0-dev+abcd1234_or_later`.

Since this triggers rebuilding a bunch of files if using `alr build` to build alr itself, it can be disabled by exporting `ALR_VERSION_DONT_PATCH` in the environment to avoid those rebuilds during development.


Fabien-Chouteau (approved):

mosteo:
I will rewrite the python part in Ada to avoid having a hidden dependency on Python, so I'm leaving this unmerged for now.

Fabien-Chouteau (approved):

PR #1533, opened 2024-01-19
Fix redeploying of toolchains

mosteo:
This addresses a corner case that arises when toolchains go missing outside of alr. Should fix the current problems with the index checks.


mosteo:
I'm going to merge proactively to clean up the mess on the index submissions. If you find anything concerning with the PR we can address it in a follow-up.

PR #1538, opened 2024-01-21
Test for full publishing workflow

mosteo:
This test goes online and creates a PR against a test index on github, so we probably don't want to run it every time but locally and sporadically, or just before a release. For that reason it belongs to a new class of 'local' tests that are normally skipped unless enabled explicitly.

Goes on top of #1537

### DIFF
--- a/testsuite/README.md
+++ b/testsuite/README.md
@@ -57,6 +57,9 @@ no matter their value, or lack of one.
 - `ALIRE_DISABLE_NETWORK_TESTS`: when defined, tests that
   require non-local network use will be skipped.
 
+- `ALIRE_ENABLE_LOCAL_TESTS`: when defined, tests that are intended to be run
+  locally only by the Alire developer team will not be skipped.
+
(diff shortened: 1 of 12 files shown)


testsuite/README.md
```diff
@@ -57,6 +57,9 @@ no matter their value, or lack of one.
 - `ALIRE_DISABLE_NETWORK_TESTS`: when defined, tests that
   require non-local network use will be skipped.
 
+- `ALIRE_ENABLE_LOCAL_TESTS`: when defined, tests that are intended to be run
```
Fabien-Chouteau: I think it would be better to prefix this testsuite variables with `ALIRE_TESTSUITE_` instead of `ALIRE_`

Fabien-Chouteau (approved):

testsuite/README.md
mosteo: Indeed.

Fabien-Chouteau (approved):

PR #1539, opened 2024-01-21
Tag manual builds with commit in version

mosteo:
For builds that eventually become available for download, make the git commit appear in the version build.

Goes on top of #1530


mosteo:
Superseded by #1530

PR #1541, opened 2024-01-22
Move To_Portable to body and make it a regular function

### LINKED ISSUE #1498: using pins with a path on arm64 causes infinite loop
I'm not sure if this has been mentioned in other issues. I'm using an `arm64` linux container and I can never get pins to work properly.

As a basic example, I want to with a dependency in another folder.

```toml
[[depends-on]]
dependency = "*"

[[pins]]
dependency = { path = "path/to/dependency" }
```

This will cause alire to spin out of control. On a related note, I can't use the command `alr with dependency --use /path/to/dependency` to set this up because it will print an error:

```
error: Loading release from manifest: /workspaces/reproducer/alire/tmp/alr-kpcu.tmp:
error:    Failed to load /workspaces/reproducer/alire/tmp/alr-kpcu.tmp:
error:    invalid path name ""
```

I'm using this version of alire: `alr 1.2.1`.

I'm using this architecture: `aarch64 GNU/Linux`.

dalybrown:
This works around an issue I was encountering when using pins on an arm64 architecture. Truthfully, I don't know why this fixes the issue but it does. I didn't take the time to go back to the reference manual to fully understand why this would address the issue. Regardless, using a regular function as opposed to an expression fixed it (for now...). This closes #1498.

### DIFF
--- a/src/alire/alire-vfs.adb
+++ b/src/alire/alire-vfs.adb
@@ -11,8 +11,8 @@ package body Alire.VFS is
                               return String
    is
       Relative : constant Any_Path :=
-                   Directories.Find_Relative_Path (Parent => From,
-                                                   Child  => Path);
+         Directories.Find_Relative_Path (Parent => From,
+                                         Child  => Path);
    begin
       if Check_Absolute_Path (Relative) then
          return Path;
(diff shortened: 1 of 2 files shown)


mosteo (approved):

mosteo:
I don't think the RM is going to have answers about this, I think it's a plain bug in the compiler. Expression functions and regular ones should be equivalent (except for syntactic limitations related to aspects) from a user POV. Maybe there are differences related to build-in-place or inlining, but I would be very surprised if there were grounds for observable different behavior.

Anyway, if you eventually find something, I'd be curious to know.

PR #1550, opened 2024-01-28
Fix bugs causing spurious diffs on updates without changes

### LINKED ISSUE #1278: Don't ask if user wants to proceed if there are no changes
I ran `alr update` without changing the `alire.toml` file. The question, if the user wants to proceed is not needed then and should be skipped.
```
❯ alr update

   No changes between former and new solution.

Do you want to proceed?
Using default: Yes
```

mosteo:
Some types internally where using pointers instead of pointed objects for comparison.

Fixes unrelated bug in dealing with empty error messages (we never generate those, but some Ada 2022 'Image do).

Fixes #1278

### DIFF
--- a/alire_common.gpr
+++ b/alire_common.gpr
@@ -31,7 +31,14 @@ abstract project Alire_Common is
             "-gnatyu",   -- no unnecessary blank lines
             "-gnatyx",   -- no extra parens around conditionals
             "-gnaty-s"); -- relax fwd decl
-      when "disabled" => Style_Check_Switches := ();
+      when "disabled" => Style_Check_Switches :=
+           ("-gnat2022",
+            "-gnatx",
+            -- When no style checks are enforced, we are likely debugging,
+            -- commenting blocks, moving things around, etc. At these times we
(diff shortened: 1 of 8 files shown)


alire_common.gpr
```diff
@@ -31,7 +31,14 @@ abstract project Alire_Common is
             "-gnatyu",   -- no unnecessary blank lines
             "-gnatyx",   -- no extra parens around conditionals
             "-gnaty-s"); -- relax fwd decl
-      when "disabled" => Style_Check_Switches := ();
+      when "disabled" => Style_Check_Switches :=
+           ("-gnat2022",
```
Fabien-Chouteau, line 46: I don't think it's good to change language version when disabling style checks.
I fear this is going to have unexpected consequences that will make your life harder.

Fabien-Chouteau (approved):

alire_common.gpr
mosteo: Let's make it a separate scenario then. In any case all of our checks have checks enabled so it shouldn't slip through.

Fabien-Chouteau (approved):

PR #1566, opened 2024-02-16
New workflow to check GNAT CE builds

mosteo:
We still rely on CE2020 to be able to bootstrap some `alr` builds on macOS, and there seems to be some differences with all the FSF versions indexed by Alire.


Fabien-Chouteau (approved):

Fabien-Chouteau:
We should stop relying on GNAT CE at some point :)

mosteo:
Yup, after 2.0 we can rely on `alr` exclusively to install FSF versions without the current bootstrapping issues.

PR #1573, opened 2024-02-19
New `build --stop-after`

### LINKED ISSUE #1544: post-fetch action doesn't run when enclosed by `[[actions."case(distribution)"."..."]]`
This may not be an Alire issue and may be an issue with the crate: I couldn't tell by the documentation whether this case statement is supported for actions or not.

In the xmlada crate, there is a `post-fetch` action which is enclosed by `[[actions."case(distribution)"."..."]]`. This action does not run on deployment of the crate (whether as a dependency of another crate or by deploying the crate itself) nor does it run when `alr update` is executed. The crate must be built for this action to run. See: https://github.com/alire-project/alire-index/blob/stable-1.2.1/index/xm/xmlada/xmlada-24.0.0.toml#L23.

This is problematic because it is a dependency of many crates within the ecosystem. Consequently commands can't be executed by alire on this crate unless the crate is built. This causes issues in CI pipelines when we run jobs in parallel to the build (and we don't want to always have to build crates prior to running every job).

According to the docs, `post-fetch` actions should be run on deployment of the crate and on updates to the environment. This doesn't happen with this crate.

mosteo:
This allows stopping the build process at different points, which may help in debugging actions, ensuring up-to-date config without requiring updating dependencies nor building, etc.

Fixes #1544

### DIFF
--- a/doc/user-changes.md
+++ b/doc/user-changes.md
@@ -8,8 +8,6 @@ stay on top of `alr` new features.
 
 ### New switch `alr build --stop-after=<build stage>`
 
-### Enable shared dependencies by default
-
 PR [#1573](https://github.com/alire-project/alire/pull/1573)
 
 From `alr help build`:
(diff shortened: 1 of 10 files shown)


src/alr/alr-commands-build.adb
```diff
@@ -158,7 +217,21 @@ package body Alr.Commands.Build is
          & " (dependencies). Indirect builds through, e.g., '"
          & TTY.Terminal ("alr run") & "' will use the last '"
          & TTY.Terminal ("alr build") & "' configuration.")
-      );
+       .New_Line
+         .Append (TTY.Bold ("Build stages"))
+         .New_Line
+         .Append ("Instead of a full build, the process can be stopped early "
```
Fabien-Chouteau, line 223: Maybe explain when this could be useful. This is a debug tool, and people who are not experts may be in bad situations after using it.

src/alire/alire-builds.ads
```diff
@@ -27,6 +27,29 @@ package Alire.Builds is
    --  many more shared releases in the vault, finding toolchains could take
    --  much more time, hence the separate storage.
 
+   type Build_Stages is
```
Fabien-Chouteau: This looks like a nice process for builds, but I see it is separate from how the code will run.
We may add new stages that will not be added to this enum, and vice versa.

Maybe we can have it more explicit with a loop like this:

```ada
for Stage in Build_Stages loop
   case Stage is
      when Sync =>
      when Generation =>
      
      -  Etc,
end loop;
```

doc/user-changes.md
```diff
@@ -6,36 +6,24 @@ stay on top of `alr` new features.
 
 ## Release `2.0-dev`
 
+### New switch `alr build --stop-after=<build stage>`
+
 ### Enable shared dependencies by default
 
-PR [#1449](https://github.com/alire-project/alire/pull/1449)
```
Fabien-Chouteau, line 11: Why is PR 1449 being removed?

mosteo: Plain mistake on my part.

src/alire/alire-builds.ads
mosteo: I think I understand what you mean, but not the implementation detail. What is that loop intended to replace?

Maybe the enum can be renamed to `Stopping_Points` as to make explicit what they are intended to be used for.

Fabien-Chouteau: Maybe that's an oversimplification of the build process on my part, and the different steps in the build are probably spread across different locations in the code.

> Maybe the enum can be renamed to `Stopping_Points` as to make explicit what they are intended to be used for.

Indeed that would probably be better.

mosteo: Ah, I see. Yes, it's not as clear-cut as that, as there's some recursion involved through the dependency tree. I'll change the name then.

src/alr/alr-commands-build.adb
mosteo: I added an explanation after the steps, and a clarification that a regular `alr build` should leave things as intended.

Fabien-Chouteau (approved):

PR #1575, opened 2024-02-22
Alr.Commands.Search: change default behavior to include properties

Fabien-Chouteau:
This patch comes from a comment of a user looking for gnatpp in Alire. gnatpp is part of the libabalang_tools crate and a 1.x `alr search gnatpp` returns nothing. The reason that by default only the release name and descriptions are searched. This is quite a bad user interaction.

Here we change the default behavior to search in all properties of a release, including tags, executables, website, etc.

We also display the list of properties that match the pattern.

Searching in names and descriptions only is still possible with the --crates switch.

### DIFF
--- a/src/alire/alire-releases.ads
+++ b/src/alire/alire-releases.ads
@@ -308,6 +308,11 @@ package Alire.Releases is
    function Property_Contains (R : Release; Str : String) return Boolean;
    --  True if some property contains the given string
 
+   function Property_Contains (R : Release; Str : String)
+                               return AAA.Strings.Vector;
+   --  Return a vector with the names of propreties that contain the given
+   --  string.
+
    function Satisfies (R   : Release;
                        Dep : Alire.Dependencies.Dependency'Class)
                        return Boolean;
--- a/src/alire/alire-releases.ads
+++ b/src/alire/alire-releases.ads
@@ -309,9 +309,8 @@ package Alire.Releases is
    --  True if some property contains the given string
 
    function Property_Contains (R : Release; Str : String)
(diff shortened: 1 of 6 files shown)


src/alire/alire-releases.ads
```diff
@@ -308,6 +308,11 @@ package Alire.Releases is
    function Property_Contains (R : Release; Str : String) return Boolean;
    --  True if some property contains the given string
 
+   function Property_Contains (R : Release; Str : String)
+                               return AAA.Strings.Vector;
+   --  Return a vector with the names of propreties that contain the given
```
mosteo: typo in "propreties"

src/alr/alr-commands-search.adb
```diff
@@ -90,7 +132,8 @@ package body Alr.Commands.Search is
          Tab.Append ("external");
          Tab.Append (Alire.Index.Crate (Name).TTY_Description);
          Tab.Append (Ext.Image);
-      end List_Undetected;
+         Tab.Append (Match_Locations.Flatten (", "));
```
mosteo: These match locations could be a Set instead of a Vector, to avoid for example:
```
$ alr search hello
...
ppag    2.1.1    First steps with Gnoga       Executable, Executable, Executable, Executable, Executable, Executable, Executable, Executable, Executable, Executable, Executable, Executable, Executable, Executable
```

mosteo (approved):
Nice feature. I just saw some noisy behavior that could be improved.

mosteo (approved):

PR #1591, opened 2024-02-26
Reinstate lost style checks

mosteo:
It seems the changes in #1497 removed quite more style checks than intended. In regard to @simonjwright concerns in that PR, I think we can just disable that specific check once it is enabled by default in `-gnatyg`.


simonjwright:
Personally I’ve spent approaching 25 years  being completely happy with plain `-gnaty`. The mistake I made in #1497 was leaving out `-gnatyy` (standard style checks). Looks to me as though the problem could have been fixed just by including that.

We know `-gnatyg` now includes `-gnatyz` => "check parentheses not required by operator precedence rules", which craziness (IMO) was the prompt for #1497. I suppose we could achieve that by `-gnaty-z`? (after the `-gnatyg`, ofc, as you’ve done with `-gnaty-s`).

I would _really_ like `-gnaty-z`. Just don’t include `-gnatyo`!

alire_common.gpr
```diff
@@ -24,13 +24,12 @@ abstract project Alire_Common is
...
       when "enabled"  => Style_Check_Switches :=
            ("-gnatwe",   -- Warnings as errors
-            "-gnatyd",   -- no DOS line terminators
-            "-gnatyI",   -- no IN mode
-            "-gnatyO",   -- all overrides
-            "-gnatyS",   -- separate lines after THEN/ELSE
-            "-gnatyu",   -- no unnecessary blank lines
-            "-gnatyx",   -- no extra parens around conditionals
-            "-gnaty-s"); -- relax fwd decl
+            "-gnatyg",   -- Standard GNAT style = -gnatyydISuxz
```
Fabien-Chouteau: One potential risk with gnatyg is that its definition can change in the future, which can bring some unwanted style errors.

Why not use our default style check switches for Alire crates?

mosteo:
No love for `-gnatyo` here either. And also no problem with `-gnaty-z`.

alire_common.gpr
mosteo: I guess I prefer to make an sporadic fix if that happens rather than being unaware of recommended style changes.

mosteo:
But I'm seeing a different problem now which is that different compilers understand different switches, so it's better to go for a explicit list in the end, so I'll take Fabien's suggestion.

PR #1592, opened 2024-02-26
Don't run OS initialization during elaboration

mosteo:
As this makes impossible to tweak such initialization with `alr config`.

More precisely, it was now impossible to disable installation of msys2 via `alr config --set msys2.do_not_install`.

With this PR, msys2 installation is not run unconditionally during elaboration, but later on when we are sure the user is not running `alr config` (or `alr --version`)

For good measure, also check that unexpected msys2 installation is not attempted during test runs.

### DIFF
--- a/src/alire/os_windows/alire-platforms-current__windows.adb
+++ b/src/alire/os_windows/alire-platforms-current__windows.adb
@@ -27,21 +27,9 @@ package body Alire.Platforms.Current is
    ------------------
 
    function Detect_Msys2 return Boolean is
-      use AAA.Strings;
    begin
       --  Try to detect if Msys2's pacman tool is already in path
-      declare
-         Unused : Vector;
-      begin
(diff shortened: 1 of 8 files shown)


src/alire/os_windows/alire-platforms-current__windows.adb
```diff
@@ -27,21 +27,9 @@ package body Alire.Platforms.Current is
...
-      begin
-         Unused := OS_Lib.Subprocess.Checked_Spawn_And_Capture
-           ("pacman", Empty_Vector & ("-V"),
-            Err_To_Out => True);
-         return True;
-      exception when others =>
-            null;
-      end;
-
-      return False;
```
mosteo, line 44: This caused unintended stderr output when no `pacman` was found in path

Fabien-Chouteau (approved):

Issue #1607, opened 2024-03-01
Confusion between Alire Configuration (alr config) and Crate configuration


reznikmm:
**Describe the bug**
It seems configuration variable don't work.

**To Reproduce**
- `alr init --bin test; cd test`
- `alr config --set cortex_m.core m4f`
- `alr with cortex_m`
At this point alr 2.0 rc1 succeed, but alr 1.2.1 shows error:
> error: Configuration variable 'cortex_m.core not set and has no default value.

But the variable is there:
```
alr config --list --show-origin
LOCAL: last_build_profile=DEVELOPMENT
GLOBAL: toolchain.external.gprbuild=FALSE
GLOBAL: toolchain.assistant=false
LOCAL: cortex_m.core=m4f
GLOBAL: toolchain.use.gnat=gnat_arm_elf=13.2.1
GLOBAL: toolchain.use.gprbuild=gprbuild=22.0.1
GLOBAL: toolchain.external.gnat=FALSE
```
- `alr build`
```
warn: Configuration variable 'cortex_m.core' not set and has no default value.
error: Configuration variables without a default remain unset
```

**Expected Behavior**
I expect:
- No error messages if variable is set with `alr config --set cortex_m.core m4f`

**`alr` version**
<details>
<summary>alr version</summary>
APPLICATION
alr version:               2.0-rc1
libalire version:          2.0-rc1
compilation date:          2024-02-29 11:12:30
compiled with version:     13.2.0

CONFIGURATION
config folder:             /root/.config/alire
cache folder:              /root/.local/share/alire
vault folder:              /root/.local/share/alire/releases
build folder:              /root/.local/share/alire/builds
temp folder:               /root/test
force flag:                FALSE
non-interactive flag:      FALSE
community index branch:    stable-1.3.0
compatible index versions: ^1.1 & <=1.3.0
indexes folder:            /root/.config/alire/indexes
indexes metadata:          OK
index #1:                  (community) git+https://github.com/alire-project/alire-index#stable-1.3.0
toolchain folder:          /root/.local/share/alire/toolchains
toolchain assistant:       disabled
tool #1 gnat:              not configured
tool #2 gprbuild:          not configured
system package manager:    /usr/bin/apt
distro detection disabled: FALSE

WORKSPACE
root status:               VALID
root release:              test=0.1.0-dev
root load error:           none
root folder:               /root/test
current folder:            /root/test

SYSTEM
distribution:              UBUNTU
host-arch:                 X86_64
[...]

PR #1614, opened 2024-03-05
Remove `toolchain --install/--uninstall/--install-dir`

### LINKED ISSUE #1612: alr toolchain --install / --install-dir now obsolete
Hi @mosteo, as part of V2 breaking changes I think we mentioned removing the install part of the toolchain command in favor of the new `alr install` command.

Is there any reason to keep that in?

mosteo:
Fixes #1612


src/alire/alire-toolchains.adb
```diff
@@ -487,6 +487,17 @@ package body Alire.Toolchains is
...
+      --  Remove caching of external condition too for consistency
+
+      if CLIC.Config.Defined (Config.DB.all, Tool_Key (Crate, For_Is_External))
+      then
+         Trace.Debug
+           ("Unsetting " & Tool_Key (Crate, For_Is_External) & ": "
+            & CLIC.Config.Edit.Unset
+              (Config.Edit.Filepath (Level),
+               Tool_Key (Crate, For_Is_External))'Image);
+      end if;
```
mosteo, line 500: This is a bugfix for a corner case when the user selected None in the assistant after having selected a regular compiler, causing detection of the external compiler to be skipped. It was exposed when tweaking a test to remove `--uninstall`.

testsuite/tests/toolchain/custom_install_dir/test.py
```diff
@@ -1,44 +1,53 @@
 """
 Check folders when installing binary compiler crates at a custom location
+through `alr get`
```
Fabien-Chouteau, line 3: Why not `alr install --prefix` here?

doc/user-changes.md
```diff
@@ -6,6 +6,19 @@ stay on top of `alr` new features.
 
 ## Release `2.0-dev`
 
+### Deprecation of `toolchain --install/--uninstall/--install-dir`
+
+PR [#1614](https://github.com/alire-project/alire/pull/1614)
+
+Toolchain installation for use by Alire is still possible using
+`alr toolchain --select`. For the installation of toolchains outside of Alire
+management, `alr get` covers the `alr toolchain --install --install-dir` use
```
Fabien-Chouteau: Same as below, why not recommend `alr install --prefix` here?

Fabien-Chouteau (changes requested):
So I worked on the same thing in parallel :man_facepalming: 

In my commit I renamed `alr toolchain` to `alr default-toolchain` because I think it will remove some confusion on how Alire handles toolchains.

mosteo:
> So I worked on the same thing in parallel 🤦‍♂️

:-(
 
> In my commit I renamed `alr toolchain` to `alr default-toolchain` because I think it will remove some confusion on how Alire handles toolchains.

This will break a lot of workflows/actions out there, ours included... Do you think there's that much risk of confusion, given the `--select`?

testsuite/tests/toolchain/custom_install_dir/test.py
mosteo: Just because it is the closest to what the test was doing. We could add a new test that checks a similar idea but with `alr install`.

doc/user-changes.md
mosteo: Again, because for people that was using `--install-dir`, this is what they would find more similar. But if we want people to just use `alr install`, we can omit `alr get` altogether.

testsuite/tests/toolchain/custom_install_dir/test.py
Fabien-Chouteau: `alr toolchain --install` was  introduced for people that wanted to get a toolchain from the Alire index but not use it in the Alire context. 

I feel like `alr install` is a better option for this use case.

Fabien-Chouteau:
> > In my commit I renamed `alr toolchain` to `alr default-toolchain` because I think it will remove some confusion on how Alire handles toolchains.
> 
> This will break a lot of workflows/actions out there, ours included... 

That's why we are doing a major release ^^

> Do you think there's that much risk of confusion, given the `--select`?

I think we do see misconceptions from time to time.

Which makes me thing we could also take this opportunity to rename `alr config` into `alr settings` (see #1607 ).

testsuite/tests/toolchain/custom_install_dir/test.py
mosteo: Sure, I'll adapt the test.

mosteo:
> > > In my commit I renamed `alr toolchain` to `alr default-toolchain` because I think it will remove some confusion on how Alire handles toolchains.
> > 
> > 
> > This will break a lot of workflows/actions out there, ours included...
> 
> That's why we are doing a major release ^^

I feel this is a point where I must resist, and not so much for the brokenness but because I find `default-toolchain` unnecessarily long, a bit ugly, and itself a bit misleading as for example when several are installed, the output shows both default and non-default toolchains:
```
$ alr toolchain
CRATE         VERSION       STATUS    NOTES                                     
gnat_native   11.2.4        Available                                           
gnat_native   12.2.1        Available                                           
gnat_native   13.2.1        Default
``` 

I'd be less objecting to `alr toolchain --select-default` or even `alr toolchain --set-default`.

But I have no more arguments so if you're truly convinced this will be a net gain, I won't object further.

> Which makes me thing we could also take this opportunity to rename `alr config` into `alr settings` (see #1607 ).

Sounds good, I saw the exchange with Maxim.

I'm a bit uneasy about introducing such breaking changes so late in the RC period, in particular if we dont have a 2nd RC (which I didn't plan to have). What do you think of having the old name as an undocumented alias that emits a warning pointing to the new command?

mosteo:
The last push completes the removal. Any command renamings I would do in a separate PR.

Fabien-Chouteau (approved):

Fabien-Chouteau:
@mosteo I started working on the `alr settings` modification.

PR #1615, opened 2024-03-06
Breaking changes in `alr toolchain`

Fabien-Chouteau:
First, the `alr toolchain` command is renamed `alr default-toolchain` to better represent its intent.

Second, the `--install`, `--uninstall`, and `--install-dir` options are removed. These were introduced for people that wanted to get a toolchain from the Alire index but not use it in the Alire context. We now recommend using the `alr install --prefix=` command in this case.


Fabien-Chouteau:
See #1615

PR #1617, opened 2024-03-08
New `alr settings` command to replace `alr config`

Fabien-Chouteau:
This change is introduced to tackle the confusion between the configuration of the Alire commands and operations, and the configuration of crates.

`alr config` is still available and should work as before with the exception of a deprecation warning message.


Fabien-Chouteau:
The other options would be to rename crate configuration into crate settings.

It would be trivial to change this releases currently in the index.
There are 288 occurrences of `[configuration` in the community index releases.
34 releases are defining configuration variables.
24 releases are setting configuration values.

mosteo:
> The other options would be to rename crate configuration into crate settings.

I like better `alr settings` and crate config than the other way around (so this PR). Also it seems less of a headache than breaking manifest compatibility. Also the `config` dir we create...

Fabien-Chouteau:
> Also it seems less of a headache than breaking manifest compatibility. Also the `config` dir we create...

Indeed, that would be the most annoying.

Note that there is a couple things I didn't change in this PR:
 - `-c (--config=ARG)       Override configuration folder location` global switch
 - `.alire/config.toml` file
 - `ALR_CONFIG` scenario variable
 - The name of `Alire.Config.*` packages
 - Default folder `~/.config/`

mosteo:
> * `-c (--config=ARG)       Override configuration folder location` global switch

Ouch. Maybe we can say "configuration settings" and leave as is. I don't see potential for confusion here.

> * `.alire/config.toml` file

We could rename this one on the fly.

> * `ALR_CONFIG` scenario variable

Ouch again. I would now document `ALIRE_SETTINGS` and use the old one as fallback, with warning.

> * The name of `Alire.Config.*` packages

Internal, minor issue.

> * Default folder `~/.config/`

This one comes from XDG_CONFIG_HOME. On Windows we could start using AppData/Local or similar, like with the cache...

mosteo (approved):

PR #1619, opened 2024-03-09
Handle config switch errors on missing parameter

rocher:
Both `alr -c` and `alr --config` throw exceptions because a required parameter is missing.

This fix handle missing parameter of config switch in the same way as, for example, `alr -C` and `alr --chdir`: an error message informs the user that a required parameter is missing.


mosteo (approved):

mosteo:
Thanks, @rocher

PR #1643, opened 2024-03-14
Enforce usage of Sementic_Versioning when printing Alire/Alr version

Fabien-Chouteau:
To ensure a consistent formatting whatever the version. The version string is now private, only the semver type is visible.


Fabien-Chouteau:
We've got an issue on alire.ada.dev because `v2.0.0` is only reporting `v2.0` as version string.

mosteo (approved):

PR #1650, opened 2024-03-15
Rename option arguments to indicate expected type

rocher:
In commands with options that admit arguments, like `--prefix` in `alr install`, is useful to provide an argument name in the command help that gives a hint of the expected type.

For example, `alr install -h` shows the message:
```
OPTIONS
   --prefix=ARG   Override installation prefix (default is ...)
```

When changed to:
```
OPTIONS
   --prefix=DIR   Override installation prefix (default is ...)
```
the `DIR` name of the argument immediately says what is expected in this option. In cases where this is not so obvious, it helps even more.

For example, in `alr build -h`:
```
OPTIONS
   --profiles=LIST     Comma-separated list of <crate>=<profile> values (see description)
   --stop-after=STAGE  Build stage after which to stop (see description)
```
is a quick remainder of what is expected in that options.

### DIFF
--- a/src/alr/alr-commands-build.adb
+++ b/src/alr/alr-commands-build.adb
@@ -273,13 +273,15 @@ package body Alr.Commands.Build is
         (Config,
          Cmd.Profiles'Access,
          "", Switch_Profiles & "=",
-         "Comma-separated list of <crate>=<profile> values (see description)");
+         "Comma-separated list of <crate>=<profile> values (see description)",
+         Argument => "LIST");
 
       Define_Switch
         (Config,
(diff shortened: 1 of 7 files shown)


rocher:
Don't understand why it failed in `Ubuntu-latest gcc^10`  with:

```
446   Link
447      [link]         alr-main.adb
448   gcc: fatal error: '-fuse-linker-plugin', but liblto_plugin.so not found
449   compilation terminated.
450   gprbuild: link of alr-main.adb failed
```

Any idea?

mosteo (approved):

PR #1683, opened 2024-05-14
Pin a crate in a remote monorepo

### LINKED ISSUE #1552: pinning a crate in a remote monorepo (i.e., using --use=URL) does not setup GPR_PROJECT_PATH
I'm not sure if the existing behaviour is intended or not. 

When pinning to a remote create in a monorepo, you can provide the URL and the crate name and Alire will add the dependency. However, when you examine the environment of the enclosing crate, the GPR_PROJECT_PATH is set to the root of the monorepo, not to the root of the crate within the monorepo. Consequently, the build will not find the dependency's project file.

For example, if this is the folder structure of the remote repo:

/monorepo/dependency/alire.toml
/monorepo/dependency/project.gpr

Then when you pin the dependency crate, the GPR_PROJECT_PATH is set to /monorepo, not /monorepo/dependency.

I can update the alire.toml environment to add the path, but I think it should probably be done automatically.

dalybrown:
Fixes https://github.com/alire-project/alire/issues/1552.

### DIFF
--- a/src/alire/alire-user_pins.adb
+++ b/src/alire/alire-user_pins.adb
@@ -322,9 +322,6 @@ package body Alire.User_Pins is
                               else Origins.Short_Commit (+This.Commit))
          elsif Branch (This).Has_Element
          then "#" & TTY.Emph (+This.Branch)
-         else "")
-       & (if Subdir (This).Has_Element
-         then "#" & TTY.Emph (+This.Subdir)
          else ""));
 
    ----------
--- a/src/alire/alire-user_pins.adb
+++ b/src/alire/alire-user_pins.adb
@@ -16,6 +16,7 @@ package body Alire.User_Pins is
    package Keys is
       Branch   : constant String := "branch";
       Commit   : constant String := "commit";
+      Subdir   : constant String := "subdir";
       Internal : constant String := "lockfiled";
       Path     : constant String := "path";
       URL      : constant String := "url";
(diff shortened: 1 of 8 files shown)


dalybrown:
This is a work in progress. Haven't had much time to take a look at it. I started making some changes but I am not finished yet. I put this up to get some early feedback and ensure I'm on the right path and following conventions.

dalybrown:
It's getting a bit closer ... just need to sort out where the environment (project paths, etc.) are set. Code navigation doesn't work in VS Code for me anymore so it's difficult to figure out where everything is. I'll need some help here.

src/alire/alire-user_pins.adb
```diff
@@ -312,6 +322,9 @@ package body Alire.User_Pins is
                               else Origins.Short_Commit (+This.Commit))
          elsif Branch (This).Has_Element
          then "#" & TTY.Emph (+This.Branch)
+         else "")
+       & (if Subdir (This).Has_Element
```
dalybrown: I don't think this is structured properly right now: it will produce something like `url/root.git#branch#subdir`. Probably not what we want.

dalybrown:
@mosteo I haven't circled back to this pull request yet ... but I was a bit stuck and didn't know where to go from here. When you get a minute, I could use your help! Thanks!

mosteo:
Yup, I'll have a look after I'm back from AEiC.

mosteo:
Superseded by #1857. Thanks @dalybrown for the initial work.

dalybrown:
Well I obviously didn't get around to finishing it - sorry! I definitely overestimate the amount of time i have to poke away at these things lately...

Thanks again for this - I was literally just jumping in here to see the status of this and pick it back up (after ... many many months).

mosteo:
No reason to be sorry, you laid the groundwork and I needed this anyway. Hopefully the current implementation fits your needs.

PR #1696, opened 2024-06-14
Ensure index repos can be deleted on Windows

mosteo:
As indexes are git repos, they suffer from the old deletion problem when some internal files are marked read-only by git

### DIFF
--- a/src/alire/alire-version.ads
+++ b/src/alire/alire-version.ads
@@ -16,7 +16,7 @@ private
    --  be replaced by `alr build` with the current commit, and appended with
    --  "_or_later" after build.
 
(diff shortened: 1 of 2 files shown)


src/alire/alire-version.ads
```diff
@@ -16,7 +16,7 @@ private
    --  be replaced by `alr build` with the current commit, and appended with
    --  "_or_later" after build.
 
-   Current_Str : constant String := "2.1-dev";
+   Current_Str : constant String := "2.1-dev+bd25511f_or_later";
```
Fabien-Chouteau: Do you want to check that in?

Fabien-Chouteau (approved):

PR #1700, opened 2024-06-16
bugfix: improvements to temp file name creation

mosteo:
Still seeing, albeit less frequently, flaky test runs due to tmp files, so plugging a couple of remotely probable holes.


mosteo:
This is essentially a refactor with a couple of extra checks.

mosteo:
I've seen the problem of duplicated temp file names, so that's one fewer to worry about.

Fabien-Chouteau (approved):

PR #1704, opened 2024-06-20
fix: test runs with recent Docker images

mosteo:
I bumped all of our Docker images and that has resulted in some new bugs being caught by the testsuite


novaTopFlex:
What types of bugs are you expecting to see from the Docker images for Alire, an Ada project?

mosteo:
> What types of bugs are you expecting to see from the Docker images for Alire, an Ada project?

We detect versions of system packages, and some new non-semver versions were confusing our detector.

PR #1705, opened 2024-06-23
Add OpenBSD support.

ibara:
Add OpenBSD support.

I am running the latest OpenBSD/amd64 -current with tip-of-tree GNAT:
```
GNAT 15.0.0 20240622 (experimental)
Copyright (C) 1996-2024, Free Software Foundation, Inc.
This is free software; see the source for copying conditions.
There is NO warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

The test suite (run in a venv) looks like this:
```
INFO     Summary:
_  PASS         220
_  FAIL         46
_  SKIP         5
```

Thanks.


mosteo (approved):

mosteo:
Thanks, @ibara. Do you think we could use e.g. https://github.com/marketplace/actions/cross-platform-action to run our testsuite on OpenBSD?

ibara:
> Thanks, @ibara. Do you think we could use e.g. https://github.com/marketplace/actions/cross-platform-action to run our testsuite on OpenBSD?

Probably. Honestly, I've not tried it so I am not sure how to set it up. Also, OpenBSD does not ship a package of GCC 15 (it ships a package of GCC 11.2.0, soon to be 11.4.0). And OpenBSD also does not ship a package of gprbuild I built it from bootstrap and used that to build alire.

I can get you a package of those things, but it would be for OpenBSD -current, not 7.5 release, if that works for you.

mosteo:
Mmm I see. GNAT 11 shouldn't be a problem, but the lack of gprbuild would be, unless it's simple to build.

It doesn't make much sense to have a complicated or involved setup if users are unlikely to go through the same process. The idea of testing these platforms is not only to check that `alr` works, but also that the default user experience also works. Also, I prefer to test a platform that is more likely to be the one chosen by users and that also has some guaranteed lifetime, e.g., stable releases. 

Given that any Alire would-be-user would need to build its own gprbuild, could you elaborate a bit about the difficulty of that?

ibara:
> Given that any Alire would-be-user would need to build its own gprbuild, could you elaborate a bit about the difficulty of that?

It was very simple. I'll write up some directions for you shortly.

mosteo:
I'm merging these changes, as the testing support can be added at a latter date.

ibara:
Thanks. Sorry, I've been traveling for the US holiday but hope to get you something early next week.

PR #1707, opened 2024-06-28
Update token permissions in publishing.md

LionelDraghi:
Doc update : when generating the PAT, checking “public_repo” is needed, but also "workflow".


mosteo (approved):

mosteo:
Thanks, @LionelDraghi, this was something I had had reports about and hadn't yet come around to fixing.

PR #1712, opened 2024-07-02
Create FUNDING.yml

mosteo:
Someone asked about supporting the project on Telegram, so this will increase visibility of options.


.github/FUNDING.yml
```diff
@@ -0,0 +1,5 @@
+# This is a YAML file, hence this line is a comment
```
Fabien-Chouteau: You can probably remove the comments.

Fabien-Chouteau (approved):

PR #1719, opened 2024-07-12
Made error message more specific.

bracke:
Changed the error message which complains about a tag being too long, to actually tell which tag it is.


Fabien-Chouteau:
This seems like a good improvement.

PR #1721, opened 2024-07-20
Fix Bash completions

pdietl:
Fix Bash completions                  
                                                                                                                                                                                                                                                                                                                                                            The previous completion script was including extra completions that                                                                                                                              were not actually commands.       
                                                                                                                                                                                                                                                                                                                                                                This commit also moves the completion file and renames it such that                                                                                                                              on Ubuntu one can simple download the release zip, unpack it, cd                                                                                                                                 into the unpacked directory, and run `cp -r bin share /usr/local/`                                                                                                                                to install with bash-completions in the proper place to get                                                                                                                                      automatically sourced

### DIFF
--- a/.github/workflows/ci-linux.yml
+++ b/.github/workflows/ci-linux.yml
@@ -66,7 +66,9 @@ jobs:
 
     - name: Package binaries
       if: (github.event_name == 'release')
-      run: zip alr-bin-linux.zip bin/alr LICENSE.txt
+      run: |
+        echo 'sudo cp -r bin share /usr/local/' > INSTALL.md
+        zip alr-bin-linux.zip bin/alr share/bash-completion/completions/alr LICENSE.txt INSTALL.md
 
     - name: Retrieve upload URL for the release
       if: (github.event_name == 'release')
--- a/share/bash-completion/completions/alr
+++ b/share/bash-completion/completions/alr
@@ -13,8 +13,43 @@ else
 fi
 alr settings --global --set index.auto_update 0
 
-# Commands/Topics: all line-first words not starting with capital letter, after # COMMANDS
-_alr_commands=$(alr | grep COMMANDS -A 99 | awk '{print $1}' | grep -v '[[:upper:]]' | xargs)
+# The help output is roughly structured into sections like so:
+# SECTION_NAME_1
+#   Subsection
+#   word    description
+#
+# Note that the section names are in ALL_CAPS and the subsection names start
+# with a capital letter. The good stuff is the lowercase words in the COMMANDS and ALIASES sections
+
+# First, we collect all commands (words not starting with an uppercase letter) in the COMMANDS section.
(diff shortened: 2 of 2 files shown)


.github/workflows/ci-linux.yml
```diff
@@ -66,7 +66,9 @@ jobs:
 
     - name: Package binaries
       if: (github.event_name == 'release')
-      run: zip alr-bin-linux.zip bin/alr LICENSE.txt
+      run: |
+        echo 'sudo cp -r bin share /usr/local/' > INSTALL.md
```
mosteo, line 70: I'd rather have a proper INSTALL.md file a bit more fleshed out, where this command is not just given but also explained. Some people may prefer to use another prefix not requiring sudo.

share/bash-completion/completions/alr
```diff
@@ -13,8 +13,43 @@ else
...
+_alr_commands+=' '
+
+# Now do the same, but for the ALIASES section.
+
+_alr_commands+=$(alr | awk '
+    $0 == "ALIASES", ($0 ~ /^[A-Z]+$/ && $0 != "ALIASES") {
+        if ($0 == "ALIASES" || $1 ~ /^[A-Z]/ || !NF)
+          skip
+        else
+          print $1
```
mosteo: Could the common bits here be refactored to avoid duplication? This is what I get from AI, untested:
```
get_section_commands() {
    local section=$1
    alr | awk -v section="$section" '
        $0 == section, ($0 ~ /^[A-Z]+$/ && $0 != section) {
            if ($0 == section || $1 ~ /^[A-Z]/ || !NF) skip
            else print $1
        }
    '
}

_alr_commands=$(get_section_commands "COMMANDS")
_alr_commands+=' '
_alr_commands+=$(get_section_commands "ALIASES")
```

mosteo (changes requested):

.github/workflows/ci-linux.yml
pdietl: Should I make this Linux-specific or even distro specific? I believe that where to put Bash completion files is a bit distro-dependent.

share/bash-completion/completions/alr
pdietl: Yes I believe this will work.

pdietl: Changed in next commit.

.github/workflows/ci-linux.yml
mosteo: Can this be found automatically? Otherwise I don't think we want to manually maintain such a thing. We can explain in the INSTALL.md file that the given structure is for Debian and should be tweaked if needed. Also yes, I think the INSTALL.md should make clear this is only for Unix-like, and that there's a Windows installer.

mosteo:
I want to eventually get this one in

pdietl:
It looks like someone already added completions

PR #1722, opened 2024-07-24
Fix changing branch of pins with 'git+ssh://' and 'xyz+https://' urls

Seb-MCaw:
Currently, specifying the branch of pinned remote repositories fails for `git+ssh://` urls. The requested branch is fetched correctly when alr build is first run (or after manually clearing the cache), but subsequent changes to the `branch =` component of the pin in `alire.toml` always results in the default branch being checked out. The same bug is also observed with `[^+]*\+https:` urls.

`git+ssh://` urls actually [appear to be deprecated](https://github.com/git/git/blob/ad850ef1cf1b4af1c1fa3ed8f79e48b8273aa557/connect.c#L725), but we should probably support them anyway.

The issue manifests in [the `Update` procedure](https://github.com/alire-project/alire/blob/master/src/alire/alire-user_pins.adb#L196). This detects if the url has changed and, if so, it clears the cached copy and does a fresh `git clone` (presumably on the basis that a different url probably points to a different version of the repository, which may have diverged sufficiently that an update will fail).

For some reason, this fresh `git clone` always checks out the default branch ([calls `Checkout`](https://github.com/alire-project/alire/blob/master/src/alire/alire-user_pins.adb#L210), rather than `Checkout (Branch => Branch)`). I'm not sure whether this is intended behaviour, but it is not ordinarily a problem as [`Deploy`](https://github.com/alire-project/alire/blob/master/src/alire/alire-user_pins.adb#L128) always seems to be called several times. As such:
- The first call identifies that the url has changed, and consequently does a fresh `clone` of the default branch
- The second (and subsequent) call(s) considers the url to be unchanged (since it is the same as the first call), and therefore triggers the normal update procedure (which does checkout the correct branch)

This fails for `git+ssh://` and `[^+]*\+https:` urls because the [`Repo`](https://github.com/alire-project/alire/blob/master/src/alire/alire-vcss-git.adb#L164) function (which constructs the url when performing the `clone`) [removes the prefix](https://github.com/alire-project/alire/blob/master/src/alire/alire-vcss.adb#L55). A change to the url is detected by comparing the appropriate line in `git config --list` ([here](https://github.com/alire-project/alire/blob/master/src/alire/alire-vcss-git.adb#L328)) with the string defined in `alire.toml` ([here](https://github.com/alire-project/alire/blob/master/src/alire/alire-user_pins.adb#L470)). `git config` returns the form passed to the `clone` command (without the prefix), while the string in `alire.toml` retains the full `git+ssh://` form. The net effect is that a pin with a `git+ssh://` url will **always** be treated as having changed (from `ssh://` to `git+ssh://`), and the default branch will always be used (for **every** call to `Deploy`).

This pull request resolves the issue by substituting `git+ssh://` &rarr; `ssh://` and `xyz+https://` &rarr; `https://` when reading `alire.toml`.

### DIFF
--- a/src/alire/alire-user_pins.adb
+++ b/src/alire/alire-user_pins.adb
@@ -7,6 +7,9 @@ with Alire.Utils.User_Input;
 with Alire.Utils.TTY;
 with Alire.VFS;
 
+with Ada.Strings.Unbounded;
+with AAA.Strings;
+
 with GNAT.OS_Lib;
 
 package body Alire.User_Pins is
@@ -465,6 +468,8 @@ package body Alire.User_Pins is
          -----------------
 
          function Load_Remote return Pin is
+            use Ada.Strings.Unbounded;
+            use AAA.Strings;
             Result : Pin :=
                        (Kind       => To_Git,
                         URL        => +This.Checked_Pop (Keys.URL,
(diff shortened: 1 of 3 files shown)


mosteo:
> For some reason, this fresh `git clone` always checks out the default branch ([calls `Checkout`](https://github.com/alire-project/alire/blob/master/src/alire/alire-user_pins.adb#L210), rather than `Checkout (Branch => Branch)`).

AFAIR, this is because in the past some git versions found in the wild wouldn't have yet the option to check out a particular branch. I think it is safe to use this now when needed.

mosteo (approved):

mosteo:
Thanks for the comprehensive test

PR #1731, opened 2024-08-05
Build on Mac OS X ARM64

reznikmm:
As discussed in https://github.com/alire-project/setup-alire/pull/74

### DIFF
--- a/.github/workflows/ci-macos.yml
+++ b/.github/workflows/ci-macos.yml
@@ -11,12 +11,20 @@ on:
     types: [published]
   workflow_dispatch:
 
+env:
+  ARCH: x86_64
+
 jobs:
 
   build:
     name: CI on macOS
 
-    runs-on: macos-12
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [macos-12, macos-14]
+
+    runs-on: ${{ matrix.os }}
 
     steps:
     - name: Check out repository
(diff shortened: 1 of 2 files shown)


.github/workflows/ci-macos-arm64.yml
```diff
@@ -0,0 +1,95 @@
...
+    types: [published]
+  workflow_dispatch:
+
+jobs:
+
+  build:
+    name: CI on macOS
+
+    runs-on: macos-14
+
```
mosteo: Do you think we could avoid some duplication by having a single `ci-macos.yml` with a few conditional steps?

mosteo (commented):
See comment about merging both macos workflows

.github/workflows/ci-macos-arm64.yml
reznikmm: Ok. I've rewritten it in a single file as you advised.

reznikmm:
I've tried a testing release. It creates `alr-2.0.2-bin-aarch64-macos.zip`. Is it appropriated name?

reznikmm:
ping

mosteo:
Yes, I think that's the proper name as we use internally aarch64 over arm64.

Thanks for the rework.

mosteo (approved):

reznikmm:
You are welcome. If you have suggestion for next step(s) to release `alr` on Mac OS X ARM64, don't hesitate to tell me.

mosteo:
I think with this we are set. I don't want to make a release right now because I have limited time available in case of surprises, but in a couple weeks we can go ahead.

PR #1739, opened 2024-08-15
Solver refactoring

mosteo:
This reworks the recursive solver we have into an A*-like sequential one. The advantages are: cleaner implementation, better maintainability, no need to start the search several times to find a good incomplete solution (once we find an incomplete one, it will be decent enough and it means no complete one exists), simpler options for usage.

For users there should be no noticeable changes.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -57,9 +57,6 @@
 [submodule "deps/dirty_booleans"]
 	path = deps/dirty_booleans
 	url = https://github.com/mosteo/dirty_booleans
-[submodule "deps/compare_to_case"]
-	path = deps/compare_to_case
-	url = https://github.com/mosteo/compare_to_case
 [submodule "deps/den"]
 	path = deps/den
 	url = https://github.com/mosteo/den
--- a/.gitmodules
+++ b/.gitmodules
@@ -63,6 +63,3 @@
 [submodule "deps/cstrings"]
 	path = deps/cstrings
 	url = https://github.com/mosteo/cstrings
-[submodule "deps/lml"]
-	path = deps/lml
-	url = https://github.com/mosteo/lml_ada.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -57,9 +57,15 @@
 [submodule "deps/dirty_booleans"]
 	path = deps/dirty_booleans
 	url = https://github.com/mosteo/dirty_booleans
+[submodule "deps/compare_to_case"]
+	path = deps/compare_to_case
+	url = https://github.com/mosteo/compare_to_case
 [submodule "deps/den"]
 	path = deps/den
 	url = https://github.com/mosteo/den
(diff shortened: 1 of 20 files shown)


.gitmodules
```diff
@@ -57,6 +57,9 @@
 [submodule "deps/dirty_booleans"]
 	path = deps/dirty_booleans
 	url = https://github.com/mosteo/dirty_booleans
+[submodule "deps/compare_to_case"]
+	path = deps/compare_to_case
+	url = https://github.com/mosteo/compare_to_case
```
Fabien-Chouteau: I am not too fond of having a dependency project with only one function in it.

src/alire/alire-solver.ads
```diff
@@ -2,71 +2,48 @@ with Alire.Dependencies;
...
-      --  releases. All possible solutions and incomplete subsets are
-      --  eventually explored.
-
-     );
-   --  Allow the solver to further explore incomplete solution space. Each
-   --  value takes more time than the precedent one. All_Incomplete can take
-   --  a veeery long time when many crates/releases must be considered. TODO:
-   --  All these policies can go away once we move from a recursive solver to
-   --  a non-recursive priority-based one.
+   --  When looking for releases within a crate, which one to try first. The
```
Fabien-Chouteau, line 22: So what is the strategy for Alire?

Fabien-Chouteau (approved):
I can't say I fully understand the code but it looks well organized and documented.

I am a bit worried about the dependency on a growing number of  very small libraries.

.gitmodules
mosteo: I can see the concern. I was operating from the idea of not having in Alire logic unrelated to its function, but what is the minimum reusable code that makes worthwhile moving it into a new crate is certainly debatable.

I will move that function for now into `Alire.Utils`.

src/alire/alire-solver.ads
mosteo: Alire defaults to newest first, unless the global switch `--prefer-oldest` is used, which affects `get`, `search` and `show`. Notably, `alr update` (and automatic updates after manual manifest edition) don't honor it, so I've had for a time the idea of moving it to a setting, so the behavior is globally consistent.

src/alire/alire-meta.ads
```diff
@@ -6,8 +6,8 @@ package Alire.Meta with Preelaborate is
 
    package Working_Tree is
 
-      Commit  : constant String := "unknown";
-      Changes : constant String := "unknown";
+      Commit  : constant String := "c3dade93403aaacec59097e450a22ce83e0ceb0f";
+      Changes : constant String := "dirty";
```
Seb-MCaw, line 10: Did you intend to include this?

mosteo: No, thanks for the catch!

PR #1745, opened 2024-08-28
Add support for private indexes to "alr publish"

### LINKED ISSUE #1553: Mandatory maintainers-logins field in manifest should not be limited to GitHub logins
I'm not sure the history of why this field is a prerequisite to publishing. However, I don't think it should be limited to just GitHub logins since some of us use (or... want to use) a private index not hosted on GitHub.

If the idea is to limit who can modify the crate, perhaps instead of hard coding rules to match GitHub logins, it can automatically detect which system (GitLab, GitHub, others) and use rules to match logins for that. Alternatively, provide some way to generate publishing information from the `alr publish` commands that allows you to omit this field. A third option is to look for a code owners file in the repo and respect those rules on who can update the crate.

Seb-MCaw:
Closes #1553

Currently, the `publish` command always requires the user to provide (or have already configured) the username for a GitHub account, automatically forks the community index, and performs various other checks which are not necessarily applicable to private indexes. It also automatically raises a pull request into the community index, but this can be disabled with the `--skip-submit` switch.

This PR:
- `alr publish`
  - extends the effect of the `--skip-submit` switch to disable the requirement for a GitHub username to be provided and skip the automatic forking.
  - adds the `--for-private-index` switch, which disables those checks only applicable to the community index:
    - The "maintainers-logins" field of the manifest need not be a valid GitHub username, or exist at all
    - URLs which appear private are permitted
- `alr init`
  - makes the `maintainers-logins` field optional (configuration of `user.github_login` was already optional).
  - slightly changes the UI to clarify the function of `alr config`.
- General
  - changes validation to allow manifests with no `maintainers-logins` field, or whose `maintainers-logins` field has values which are invalid GitHub usernames (but empty strings are still considered invalid)


src/alr/alr-commands-init.adb
```diff
@@ -616,11 +624,30 @@ package body Alr.Commands.Init is
...
       --  Query User info
+      if User_Not_Already_Configured then
+         TIO.New_Line;
+         AAA.Text_IO.Put_Paragraph
+           ("Alire needs some user information to prepare the crate for "
+            & "eventual submission to an index, which will be interactively "
+            & "requested now.");
+         TIO.New_Line;
+         TIO.Put_Line
+           ("You can edit this information at any time with 'alr config'");
```
mosteo: `alr config` is now `alr settings`

doc/publishing.md
```diff
@@ -313,15 +313,32 @@ This will be shown as:
...
-by the dependency solver. This is only necessary for the first release in a
-crate that uses the `"provides"` feature.
+Having a local or private index may be useful sometimes, be it for local
+testing, or for private crates not intended for publication.
+
+There is no practical difference between the community index and a private index
+stored locally on disk or on your own infrastructure. An index must be an
+accessible path or URL at which an `index.toml` file can be found, either at the
+root or in any immediate subdirectory. This file specifies the version of the
+index, containing one line with the form `version = "0.0.0"`. No files should be placed in the same
```
mosteo: The `0.0.0` might be misleading; I'd rather mention that supported index versions can be found with `alr version`.

doc/publishing.md
```diff
@@ -313,15 +313,32 @@ This will be shown as:
...
+location as `index.toml` except for the manifests of published crates.
```
mosteo: I'd clarify that the manifests have to be in their corresponding subdirectories, not besides the `index.toml`

testsuite/drivers/alr.py
```diff
@@ -282,7 +304,8 @@ def index_version():
     return index_branch().split('-')[1]
 
 
-def init_local_crate(name="xxx", binary=True, enter=True, update=True):
+def init_local_crate(name="xxx", binary=True, enter=True, update=True,
+                     with_maintainer_login=False):
```
mosteo, line 308: This is because the new default for `alr init` is not to have a github username?

testsuite/drivers/helpers.py
```diff
@@ -293,3 +294,80 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+    The mocked version of git will be placed in mock_git_dir, which will be
+    temporarily added to PATH.
+    """
+
+    def __init__(self, substitution_dict, mock_git_dir):
+        self._substitution_dict = substitution_dict
+        self._mock_git_dir = mock_git_dir
+
+    def __enter__(self):
+        # Mocking on Windows would require git.exe wrapper
```
mosteo, line 340: What are the difficulties about having it for Windows too?

src/alire/alire-publish.adb
```diff
@@ -580,23 +618,40 @@ package body Alire.Publish is
             then
                raise Early_Stop;
             end if;
+         elsif Context.Options.For_Private_Index then
+            --  We are publishing to a private index, the location of which is
+            --  unknown, so we can only give generic instructions on where to
+            --  place the file.
+            Put_Info
+              ("Please upload this to the index in the "
```
mosteo: "this" --> "this file"

mosteo (changes requested):
Just minor things. Thanks for the comprehensive test!

Fabien-Chouteau (approved):

src/alr/alr-commands-init.adb
Seb-MCaw: I was going to raise a separate PR for this, as there are a couple of other instances, but might as well do it here.

testsuite/drivers/alr.py
Seb-MCaw: Essentially, yes.

To be more precise, by default (i.e. when pressing &#9166; with a blank input) `alr init` previously added the line `maintainers-logins = ["github-username"]` to the manifest (which is [a real user](https://github.com/github-username) by the way). It now omits the `maintainers-logins` field altogether (since it is no longer required). In both cases, `user.github_login` was/is left unset, so apologies the PR description was somewhat misleading in this respect.

Since some tests run `alr publish` (without `--for-private-index`) on the resulting crate, which fails in the absence of `maintainers-logins`, the `with_maintainer_login` option restores the old behaviour for those tests which need it.

testsuite/drivers/helpers.py
Seb-MCaw: In Windows, non-executable scripts cannot in general be called directly, so placing a `git`, `git.py`, `git.bat` or similar script on `PATH` does not have the desired effect (with or without modifications to `PATHEXT`).

To be clear, there's no reason it can't be done, but it seemed more trouble than it's worth for a test that ought not be particularly platform dependent. If you want, I can look into `gprbuild`ing a binary wrapper instead of writing it to a Python script.

testsuite/drivers/alr.py
mosteo: OK, it's as I understood it but thanks for the clarification.

> which is [a real user](https://github.com/github-username) by the way

Well, now this is funny...

testsuite/drivers/helpers.py
mosteo: Alright, I see. 

> a test that ought not be particularly platform dependent

I keep being surprised by how quirky Windows can be in the most unexpected ways, but let's not complicate this test any further. If evidence arises of issues on Windows we can do it then.

mosteo (approved):

mosteo:
The publishing code is not one of the most straightforward in Alire, and the testing is particularly unfriendly, so congrats on this one.

PR #1760, opened 2024-09-18
Add test for building on an air-gapped system

Seb-MCaw:
We are currently interested in the possibility of using Alire on an air-gapped system. The idea would be to fetch a crate and its dependencies online, then transfer to an offline system/network before building.

`alr get` and `alr build` seem to already implement the necessary functionality, so this PR simply adds a test to verify the process works as expected.


testsuite/tests/workflows/air-gapping/test.py
```diff
@@ -0,0 +1,90 @@
...
+
+from drivers.alr import run_alr
+from drivers.helpers import init_git_repo, on_windows
+from drivers.asserts import assert_eq, assert_match
+
+
+# Mock git, curl, gprbuild etc. with dummy scripts
+if on_windows():
+    print('SKIP: command mocking unavailable on Windows')
+    sys.exit(0)
```
mosteo: Rather than hardcoding this detection in the test, look at the test skels examples for using the `control` field in the yaml file.

testsuite/tests/workflows/air-gapping/test.py
```diff
@@ -0,0 +1,90 @@
...
+            "sys.exit(1)"
+        ]))
+    os.chmod(executable, 0o764)
+os.environ["PATH"] = f'{os.getcwd()}{os.pathsep}{os.environ["PATH"]}'
+os.chdir("..")
+
+
+# 'dependencies.shared=false' is required to ensure dependencies are packaged
+# inside the workspace (at ./alire/cache/dependencies)
+run_alr("settings", "--global", "--set", "dependencies.shared", "false")
```
mosteo: Likewise, you can configure the test in the yaml to only run in sandboxed mode.

testsuite/tests/workflows/air-gapping/test.py
```diff
@@ -0,0 +1,90 @@
...
+os.chdir(f"hello")
+p = run_alr("build", quiet=False, complain_on_error=False)
+assert_match(".*Mocked command called", p.out)
+assert_match(r'.*Command \["gprbuild", .*\] exited with code 1', p.out)
+
+# Disable gprbuild mocking and run `alr build` to build the crate (with git
+# mocking still enabled to check it doesn't try to fetch anything else)
+os.remove(os.path.join("..", "path-dir", "gprbuild"))
+p = run_alr("build", quiet=False)
+assert_match(".*Build finished successfully in .* seconds", p.out)
```
mosteo: You probably don't need this line as any error will cause the previous line to raise, but if you want it as a double-check it's OK.

testsuite/tests/workflows/air-gapping/my_index/crates/hello/HEAD
```diff
@@ -0,0 +1 @@
+ref: refs/heads/master
```
mosteo: I remember having some troubles in the past because of bare repositories within the parent repository. If it's the same to you, I'd rather prefer that the crates are .tgz files and you use `tar` as the mock command.

mosteo (changes requested):
The test is essentially OK but please fix these minor comments.

@Fabien-Chouteau, I guess you need to review this one too.

Fabien-Chouteau:
Hi Alejandro,

> @Fabien-Chouteau, I guess you need to review this one too.

You don't need my reviews for @Seb-MCaw 's contributions :)

testsuite/tests/workflows/air-gapping/test.py
Seb-MCaw: I hadn't noticed `skip_unix` was an option; thanks for pointing that out.

testsuite/tests/workflows/air-gapping/test.py
Seb-MCaw: Definitely unnecessary, given the check that follows.

testsuite/tests/workflows/air-gapping/my_index/crates/hello/HEAD
Seb-MCaw: Sure, no problem.

mosteo:
> You don't need my reviews for @Seb-MCaw 's contributions :)

OK, noted.

mosteo (approved):

mosteo:
Merged, thanks!

Issue #1769, opened 2024-09-30
Building Crates in Nix


atalii:
[Nix](https://nixos.org/) has some somewhat aligned goals to Alire, but goes about them at the level of a package manager rather than language-level tooling. Nix can be used to wrap some build tools (traditionally autotools, but there's good integration with cmake, meson, cargo, go, &c) to enable and provide reproducible packaging.

I would like to be able to package Alire crates in Nix. To do so, I'm testing with the following derivation:

```nix
pkgs.stdenv.mkDerivation {
  pname = "test-crate";
  version = "0.1.0";
  src = ./.;

  nativeBuildInputs = with pkgs; [ gprbuild gnat alire ];

  buildPhase = ''
     alr build
  '';

  installPhase = ''
    cp -r bin $out/bin
  '';
}
```

If you run this, you run into these two issues first:

1. $HOME is set to `/homeless-shelter`. For Alire not to complain, we have to `unset HOME` prior to `alr build`. Alternatively, Alire could automatically fallback to a `/tmp` directory if $HOME doesn't exist and can't be created --- would that be a good thing for it to do? I'd be happy to send a PR.
2. The build environment doesn't contain git. Of course, Alire quits execution as soon as it finds that out. We could just install git in `nativeBuildInputs` so that it's available, but then we run into a more fundamental issue: Nix doesn't allow derivations to access the internet\*, and the first thing Alire tries to do is fetch the index.

<sub>*Fixed-output-derivations allow for internet access in exchange for a hash of the expected output to ensure reproducibility. Nix wrappers of, e.g., Go and Cargo each use this to build a 'vendored' version of their packages, and then perform an offline build.</sub>

The broader issue, however, appears to be that **there is no existing mechanism for offline and sandboxed builds.** Thus, two things primarily are required to support Nix:

1. A way to vendor crate dependencies alongside the code.
2. A way to build offline, assuming deps are vendored.

There's also a secret third extra difficulty in that Nix can't run binaries downloaded from the internet due to dynamic linking issues. This is obviously very much at odds with Alire's toolchain mechanism. I would think that part of an offline build involves using the native or local toolchain, but that doesn't strike me as a complete solution.

---

Is this impression correct? Am I understanding the problem space well; i.e., is this all required to build Alire crates with Nix? Do you think I'm missing anything or making it more complicated than it needs to be?

[...]

PR #1770, opened 2024-10-01
fix: use tmp dir if $HOME isn't writable

atalii:
All the tests pass on my machine; let me know if there are any issues, if an additional test would be warranted, &c.

Context for this PR is in #1769 - this makes crate builds in Nix a bit more ergonomic by eliding the need to `unset HOME`. Hopefully it'll only be a few patches before they're trivial :)

Thanks!


mosteo (approved):

mosteo:
Looks good to me. I will add a test for it before merging.

PR #1779, opened 2024-10-14
Fix Spellcheck

### LINKED ISSUE #1776: spellcheck does not fail when there are typos
[example](https://github.com/alire-project/alire/actions/runs/11298298089/job/31426947613)

BTW it looks like the action is only triggered by `**.md` files. Is this on purpose?

Seb-MCaw:
Closes #1776 

The spellcheck check now has what I'm guessing is the intended behaviour:

- The check fails if typos are detected
- Typos [trigger annotations](https://github.com/alire-project/alire/actions/runs/11323681673) (PR comments would presumably be preferable, but I believe this would require a `pull_request_target` workflow, which [has non-trivial security implications](https://securitylab.github.com/resources/github-actions-preventing-pwn-requests/))
- Only `.md` (_i.e._ documentation) files are checked

Let me know if I have guessed incorrectly.

### DIFF
--- a/.github/workflows/spellcheck.yml
+++ b/.github/workflows/spellcheck.yml
@@ -8,11 +8,12 @@ jobs:
   spellcheck:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v1
-    - name: misspell
+    - uses: actions/checkout@v4
+    - name: Run misspell
       uses: reviewdog/action-misspell@v1
       with:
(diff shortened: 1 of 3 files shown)


mosteo:
Your assessment is correct. 

The `pull_request_trigger` considerations I think would be safe in this instance, but it doesn't seem important enough to open that can of worms.

mosteo (approved):

PR #1782, opened 2024-10-14
Add support for Git repos with multiple remotes configured

Seb-MCaw:
A Git repo may have multiple remotes configured, often with different branches tracking different remotes. For example, development may be done on a private repository, with release/production versions available on a public repository.

Currently Alire assumes there is at most one remote, and therefore ends up [using whichever is first alphabetically](https://github.com/alire-project/alire/blob/e20a5cbe0bb29c59c68361321c8e0bc04a81b377/src/alire/alire-vcss-git.adb#L416) when there is more than one. In practice, this is only relevant to `alr publish`ing, since dependencies are always fetched as a plain `git clone`, which does indeed yield a repo with only one remote.

This PR makes this assumption more explicit where applicable, and adds a subprogram to `Alire.Publish.Local_Repository` which attempts to guess the appropriate remote when there is more than one.


Seb-MCaw:
This is basically ready for review, but it's left as a draft because there is some minor overlap with #1736; I'll make a few more changes here if/when that one is merged.

Seb-MCaw:
This one is ready for review too, @mosteo. You can ignore my previous (deleted) comment.

testsuite/tests/publish/multiple-remotes/test.py
```diff
@@ -0,0 +1,407 @@
...
+
+    Also returns the commit's hash and attaches the tag `f"tag_{commit_name}"`
+    thereto.
+    """
+    with open(path, "x") as f:
+        f.write(content)
+    run(["git", "add", path])
+    run(["git", "commit", "-m", f"Commit {commit_name}"])
+    run(["git", "tag", f"tag_{commit_name}"])
+    return git_head()
```
mosteo: Should this be in helpers.py named as `git_commit_file`, for general availability?

mosteo (changes requested):
Just a minor refactor.

Seb-MCaw:
I'm assuming [this](https://github.com/alire-project/alire/actions/runs/11554342812/job/32157351611?pr=1782#step:17:1108) is transient.

mosteo:
> I'm assuming [this](https://github.com/alire-project/alire/actions/runs/11554342812/job/32157351611?pr=1782#step:17:1108) is transient.

Yes it is.

mosteo (approved):

mosteo:
Merged, thanks.

PR #1815, opened 2025-01-06
Make source archive download command configurable

Seb-MCaw:
At the moment, the biggest obstacle to using Alire for private development is that source archives cannot be fetched from servers which require user authentication (this is already possible for private Git repositories, since the `git` command implements all authentication schemes commonly used for Git repositories).

Unfortunately, there is a much wider variety of commonly used schemes and protocols for authenticated file downloads than for Git repos, so it does not seem feasible to support all of them natively. This PR therefore makes the command used for downloads (globally) configurable with the `alr settings` key `origins.archive.download_cmd`, without changing the default behaviour.

This PR also:
- Prevents the `editor.cmd` setting being read from the local crate's `settings.toml` (this seemed to me unexpected behaviour, and possibly something of a security vulnerability)
- Adds support for mocking commands when testing on Windows
- Fixes validation of the `distribution.override` setting
- Fixes the deprecated (and ignored) `fail_on_error` in the `spellcheck` workflow
- Adds a first draft of a new section to the documentation to better signpost support for private crates/indexes

---
##### PR creation checklist
- [x] `doc/user-changes.md` has been updated.


Seb-MCaw:
Ready for review @mosteo.

src/alire/alire-settings-edit.adb
```diff
@@ -338,6 +346,31 @@ package body Alire.Settings.Edit is
...
+   is (Valid_Builtin (Key, Value, Global => False));
+
+   -------------------------
+   -- Valid_Builtin_Check --
+   -------------------------
+
+   function Valid_Builtin_Check (Lvl : Level) return CLIC.Config.Check_Import
+   is (case Lvl is
+       when Global => Valid_Global_Builtin'Access,
+       when others => Valid_Local_Builtin'Access);
```
mosteo: I'd rather have here the explicit case instead of `others`.

testsuite/tests/workflows/edit/test.py
```diff
@@ -14,30 +15,45 @@
...
 # Use echo as an editor to check command line args
-run_alr('settings', '--set', 'editor.cmd', 'echo arg1 ab${GPR_FILE}ab arg3')
-p = run_alr('edit', '--project=project1.gpr')
-assert_match("arg1 abproject1.gprab arg3", p.out)
+set_setting('editor.cmd', 'echo arg1 ab${GPR_FILE}ab arg3')
+p = run_alr('edit', '--project=project1.gpr', quiet=False)
+assert_match(r"arg1 abproject1\.gprab arg3", p.out)
+
+# Verify that an `editor.cmd` value in a crate's local `settings.toml` is
+# ignored with a warning (otherwise this would offer an inconspicouous vector
```
mosteo: s/inconspicouous/inconspicuous/

testsuite/drivers/alr.py
```diff
@@ -640,18 +640,40 @@ def external_compiler_version() -> str:
...
+    Unset a key with `alr settings`
+
+    Sets the value globally unless `local` is `True`.
+    """
+    if local:
+        run_alr("settings", "--unset", key)
+    else:
+        run_alr("settings", "--global", "--unset", key)
+
+def set_setting(key: str, value: str, local: bool = False):
```
mosteo: For consistency with other `alr_blah`, I'd rename those to `alr_settings_set` and `alr_settings_unset`.

mosteo (changes requested):
A nice set of improvements, I identified just a couple of minor changes.

mosteo (approved):

PR #1819, opened 2025-01-10
feat: setting to override trusted sites list

### LINKED ISSUE #814: Adding trusted sites
Dear all,

we are very close to fully adopting Alire for our projects. External dependencies are pulled from the Alire index.

What we still want to do is also split our internal dependencies into different repos. Currently, all our code lives in one repo.

Due to IP we can not host the repos externally on the trusted sites that are currently allowed by Alire. Is there a way we can add our in-house Git server to the list of trusted sites?

Kind regards,
Felix

Seb-MCaw:
Closes #814.

The `alr publish` and `alr index --check` commands currently require that Git origins use a URL with a host found in a hard-coded list of [trusted sites](https://github.com/alire-project/alire/blob/79a83539bb1cded99a03f3bb89d0f6ad26b6869a/src/alire/alire-publish.adb#L52), motivated by [concerns regarding SHA1 hash collisions](https://github.com/alire-project/alire/blob/79a83539bb1cded99a03f3bb89d0f6ad26b6869a/doc/publishing.md?plain=1#L121). Those using their own hosting arrangements (with private indexes) need to be able to configure this list.

This PR makes the list configurable, except when using `alr publish` without the `--for-private-index` switch.

##### PR creation checklist
- [x] `doc/user-changes.md` has been updated, if applicable.


alire.toml
```diff
@@ -56,8 +56,9 @@ url    = "https://github.com/mosteo/aaa"
 commit = "ddfeffe2d6c8f9d19161df7b31d16d37bef4ba71"
 
 [pins.ada_toml]
-url    = "https://github.com/mosteo/ada-toml"
-commit = "da4e59c382ceb0de6733d571ecbab7ea4919b33d"
+url    = "https://github.com/pmderodat/ada-toml"
+commit = "e760110ad2b5b776a44dace31b8421532e429fbb"
+
```
Seb-MCaw, line 61: This is the version of `ada-toml` we are using for our internal mirror of Alire. I encountered some unexpected test failures on Windows when upstreaming these changes, and this discrepancy turned out to be the culprit.

If there is some reason updating the dependency is a problem, I'm happy to look into finding a workaround. Otherwise, I presume you will want to update your fork at https://github.com/mosteo/ada-toml/ and continue using that? (I can also open a PR to update [`clic`'s dependency](https://github.com/alire-project/clic/blob/56bbdc008e16996b6f76e443fd0165a240de1b13/alire.toml#L23) to use `ada_toml = "~0.2|~0.3|~0.4|~0.5"` if you want.)

Seb-MCaw:
Ready for review @mosteo.

alire.toml
mosteo: There's no particular reason we used that commit, I've been updating it whenever we have needed some new feature. My fork has been used only when no appropriate release by the upstream `ada_toml` repo was available. In short, it's OK to change this one.

src/alire/alire-publish.adb
```diff
@@ -1084,13 +1075,40 @@ package body Alire.Publish is
                     else No_Steps));
    end Directory_Tar;
 
+   -------------------
+   -- Trusted_Sites --
+   -------------------
+
+   function Trusted_Sites (Ignore_Setting : Boolean) return Vector is
```
mosteo: Thinking in positive, I'd prefer if `Ignore_Setting` were `For_Community` with the same meaning.

src/alire/alire-index.adb
```diff
@@ -129,7 +129,7 @@ package body Alire.Index is
       for Crate of All_Crates.all loop
          for Rel of Crate.Releases loop
             if Rel.Origin.Kind in Origins.VCS_Kinds then
-               if not Publish.Is_Trusted (Rel.Origin.URL) then
+               if not Publish.Is_Trusted (Rel.Origin.URL, False) then
```
mosteo: Better use the explicit parameter name with `=> False`

src/alire/alire-publish.ads
```diff
@@ -47,11 +47,17 @@ package Alire.Publish is
...
-   function Is_Trusted (URL : Alire.URL) return Boolean;
-   --  According to our whitelist
+   procedure Print_Trusted_Sites (For_Private_Index : Boolean);
+   --  Print our list of allowed sites to host git releases.
+   --
+   --  If For_Private_Index is True, the list is that configured with the
+   --  'origins.git.trusted_sites' setting, otherwise it is the hardcoded
+   --  Community_Trusted_Sites list.
+
+   function Is_Trusted (URL : Alire.URL; Ignore_Setting : Boolean)
```
mosteo: Likewise about `Ignore_Setting`, either `For_Community` or `For_Private_Index` or even some enum:
```
type Index_Providers is (Community_Index, Private_Index);
```

mosteo (commented):
A minor refactor and a general comment:

I find using `' '` or whitespace in general to mean any domain is a bit hackish and won't be nice when printing settings with `alr settings`. What do you think about using some impossible domain, like `'*'` or `"..."` instead? (Probably the latter to avoid interactions with shell expansions at the time of setting.)

Seb-MCaw:
> I find using `' '` or whitespace in general to mean any domain is a bit hackish ...

Agreed. `'*'` was my first thought, but will definitely cause problems, and an empty list was the best alternative I could come up with. `'...'` is a good idea, though; thanks for the suggestion.

mosteo (approved):

mosteo:
Merged, thanks.

Joebeazelman:
This feature needs to be documented for private indexes. I'm getting an error because dev.azure.com isn't a recognized one.

Seb-MCaw:
> This feature needs to be documented for private indexes.

It is already documented at the end of the second paragraph [here](https://alire.ada.dev/docs/#publishing-to-a-localprivate-index), but it could stand to be better signposted; I have opened #1985.

PR #1820, opened 2025-01-12
Universal builds for macOS

simonjwright:
For macOS (Darwin), some users are running on Intel silicon (`x86_64`), some on Apple silicon (`aarch64`). The purpose of this PR is to generate universal binaries of `alr` for macOS, so that users won't need to specify which version they need -- the loader will select the appropriate binary.

To do this, in both the `nightly.yml` and the `ci-macos.yml` workflows  a new job `build-macos-universal` runs after the `build` job has completed (which implies completion of its runs on `macos-13` (`x86_64`) and `macos-14` or `macos-latest` (`aarch64`)).

This job dowloads & unpacks the `x86_64` build to `x86_64/`, and the `aarch64` build to `aarch64/`. It then runs `lipo` to create the universal binary `alr`. The new binary has to be marked as executable, don't know why.

`alr-<release>-bin-universal-macos.zip` is created and uploaded to the release directory.

Notes:
* All three binaries (`aarch64`, `x86_64` and `universal` are created, mainly because I didn’t know how otherwise to preserve the outputs of the two parallel build matrix runs. This could cause confusion, which should be minimised when we update the ada-lang-io and alire.ada.dev sites
* I didn’t know how to get the metadata for a build (`RELEASE.dat` etc), so I copied it from the `aarch64` build.


simonjwright:
Updated from master branch.

simonjwright:
> The purpose of this PR is to generate universal binaries of `alr` for macOS, so that users won't need to specify which version they need -- the loader will select the appropriate binary.

Badly phrased, I think - "the loader will select the appropriate architecture from the single binary executable" might be clearer

.github/workflows/nightly.yml
```diff
@@ -113,3 +114,53 @@ jobs:
...
+      - name: download x86_64
+        uses: robinraju/release-downloader@v1
+        with:
+          fileName: '*-x86_64-macos.zip'
+          latest: true
+          preRelease: true
+          out-file-path: 'x86_64'
+          extract: true
+      - name: check x86_64 attributes
+        run: xattr -vr x86_64
```
mosteo: Is this a debug leftover? Or missing in `ci-macos`?

.github/workflows/nightly.yml
```diff
@@ -113,3 +114,53 @@ jobs:
...
+      - name: download aarch64
+        uses: robinraju/release-downloader@v1
+        with:
+          latest: true
+          preRelease: true
+          fileName: '*-aarch64-macos.zip'
+          out-file-path: 'aarch64'
+          extract: true
+      - name: check aarch64 attributes
+        run: xattr -vr aarch64
```
mosteo: Likewise

.github/workflows/ci-macos.yml
```diff
@@ -113,3 +113,67 @@ jobs:
...
+          out-file-path: 'aarch64'
+          extract: true
+
+      - name: Create universal binary
+        run: |
+          mkdir bin
+          lipo x86_64/bin/alr aarch64/bin/alr -create -output bin/alr
+          chmod +x bin/alr
+          cp aarch64/LICENSE.txt .
+          zip alr-bin-macos.zip bin/alr LICENSE.txt
```
mosteo, line 150: No `-universal-` in the name like in the nightly one?

.github/workflows/nightly.yml
simonjwright: See above: will remove.

.github/workflows/ci-macos.yml
simonjwright: The [upload step](https://github.com/simonjwright/alire/blob/e6b7401f1c6e70f69a7b5284f7b915a37e7ceb93/.github/workflows/ci-macos.yml#L169-L178), copied from :150 above, uses `alr-bin-macos.zip ` at line 176 and does the renaming.

.github/workflows/nightly.yml
simonjwright: Yes, it's a debug thing -- well, really investigative; since the download is via a url, I was worried that the `com.apple.quarantine` attribute might get applied to the download contents, which would mean that when a user downloads the .zip via curl they would still have to remove the attribute even though curl doesn't apply it itself. Result of investigation was that there's no quarantine attribute; will remove this step.

.github/workflows/ci-macos.yml
simonjwright: Resolved in commit 9fbdb2e.

mosteo (approved):

.github/workflows/ci-macos.yml
mosteo: Thanks, I missed the renaming in the upload step.

PR #1847, opened 2025-02-12
fix: remove `libgcc_s_seh-1.dll` dependency

Seb-MCaw:
It seems 134d1157 added a dependency on `libgcc_s_seh-1.dll` (via the new `den` crate) on Windows, without which all commands instantly (and silently) fail. This DLL is packaged along with GNAT but, since `alr` should be able to auto-install the toolchain on a fresh system, this behaviour seems like a bug.

This PR adds the `-static` linker switch, removing the external dependency on `libgcc_s_seh-1.dll`.

##### PR creation checklist

### DIFF
--- a/testsuite/tests/misc/empty-path/test.py
+++ b/testsuite/tests/misc/empty-path/test.py
@@ -1,0 +1,26 @@
+"""
+Verify that `alr` doesn't depend on anything being on `PATH` for basic commands
+(e.g. `.dll` files on Windows).
+"""
+
+
+import os
+from drivers.alr import run_alr
+
+
+# Check `alr --version` succeeds and prints something.
+p = run_alr("--version")
+assert len(p.out) > 0
+
+# Completely clear `PATH`.
+#
+# Setting it to an empty string causes a buffer overflow on some platforms
+# so we set it to a nonexistent directory.
+os.environ["PATH"] = os.path.join(os.getcwd(), "nonexistent")
+
+# Check `alr --version` still succeeds and prints something.
+p = run_alr("--version")
+assert len(p.out) > 0
+
+
+print("SUCCESS")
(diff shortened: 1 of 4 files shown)


mosteo (approved):

mosteo:
Yes, indeed, and not long ago I added a check in our GH actions to check that Alire works even after the toolchain that built it is removed. I don't really understand why it didn't catch this problem though.

Seb-MCaw:
> I don't really understand why it didn't catch this problem though.

I believe the GitHub Windows runners are using Git Bash, which puts a MinGW distribution on `PATH` by default, so it is available at `/mingw64/bin/libgcc_s_seh-1.dll` even without GNAT installed.

testsuite/tests/misc/empty-path/test.py
```diff
@@ -0,0 +1,26 @@
...
+from drivers.alr import run_alr
+
+
+# Check `alr --version` succeeds and prints something.
+p = run_alr("--version")
+assert len(p.out) > 0
+
+# Completely clear `PATH`.
+#
+# Setting it to an empty string causes a buffer overflow on some platforms
```
Seb-MCaw, line 17: This probably technically qualifies as a bug, but anybody with an empty `PATH` probably has bigger problems than `alr` failing, so I presume we can safely ignore it. Let me know if you want me to open an issue, though.

mosteo: You mean `alr` is the one having the buffer overflow?

Seb-MCaw: [Yes, the subsequent `alr --version` command fails](https://github.com/alire-project/alire/actions/runs/13291811450/job/37114259324#step:4:1515).

mosteo: Ouch. Then yes, even if it's an impractical corner case, I think we should track it.

PR #1848, opened 2025-02-13
Add test duration measurement

mosteo:
Enable time tracking in e3 testsuite

### DIFF
--- a/scripts/ci-github.sh
+++ b/scripts/ci-github.sh
@@ -114,6 +114,6 @@ $run_python ../scripts/python/check_finalize_exceptions.py ../src
 echo ............................
 
 echo Running test suite now:
-$run_python ./run.py -E || { echo Test suite failures, unstable build!; exit 1; }
+$run_python ./run.py --show-time-info -j $(nproc) -E || { echo Test suite failures, unstable build!; exit 1; }
 cd ..
 echo ............................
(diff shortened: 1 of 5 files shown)


mosteo:
@Seb-MCaw I'm seeing [this test](https://github.com/alire-project/alire/actions/runs/13306824145/job/37159604562?pr=1848#step:7:1706) taking almost 3 minutes, do you think this is expected? I see it's a long one so it could be normal.

Running it locally I had to kill it after ~10 minutes, so something seems suspicious.

Seb-MCaw:
Yes it's always been rather slow, sorry.

I've never seen it take as long as 10mins on Linux, so that would indeed be suspicious (my local machine runs the whole `master` branch testsuite in less than 2mins). It's less surprising if you're on MacOS, where the compiler (which the test is calling out to repeatedly) [seems a lot slower in general](https://github.com/alire-project/alire/actions/runs/13306824145/job/37159603202?pr=1848#step:7:1705).

Let me know if this is an issue, as I'm sure it could easily be improved; perhaps calls to `gprbuild` could be intercepted by a `MockCommand`.

mosteo:
Nevermind my time, that was an unrelated issue. That test takes a bit less than 2 mins in my machine, so it seems everything is as expected. I too had noticed that macOS, particularly x64, is much slower. Thanks!

Issue #1831, opened 2025-02-03
[RFC] Proposal for integrated test suite in Alire


AldanTanneo:
> **Disclaimer**: this is a very rough draft of a feature, in order to collect more thoughts and ideas and eventually implement something useful for the community at large.

The goal is to provide a very simple test functionality in Alire, explicitly targeted towards small community crates. That means: very little setup, **NO** extra dependencies.

The default test runner would be easily replaceable, however, to allow bigger projects to use more complex test drivers.

## CLI changes

* Retire current `alr test` subcommand, rename it `alr index-test` (**breaking change**; Alire 3.0?)
* Use the `alr test` command for running actual crate tests
* `alr test --generate`: create basic file structure for tests with default test runner. This is run only once, when setting up the project.

## Manifest changes

* New section: `[tests]`

Default keys:

```toml
runner = "alire" # or custom runner command, for ex:
                 # runner = "./run.sh"
                 # runner = ["testsuite.py", "<extra args>"]

location = "tests" # run tests command in this sub-folder
```

* The test runner command is executed in the alire execution environment; that means it can use binary dependencies specified in the `tests` sub-folder's `alire.toml`, if it exists (recommended). This opens the possibility to provide more advanced test runners in the alire index.

## Default test runner

**Goals:** be extremely simple, but also very easy to replace

Default dir structure (`tests` is a sub-folder of the main crate folder):

```
tests
|_ src
|  |_ tests-[name1].adb
|  |_ tests-[name2].adb
|_ common
|  |_ tests.ads
|  |_ [test helper files]
|_ alire.toml
|_ tests.gpr
```

All tests are sub-procedures of a main `Tests` package, optionally with test helper code located in `common`. Each test file contains a single procedure that will be run in its own process (potentially in parallel - leverage process isolation).

Default `alire.toml` and `tests.gpr` are created by `alr test --generate`; the GPR file imports a config file specifying the `Main` targets, that is recreated when the tests to compile and run change (ie. from test filtering or from adding/removing tests). When running `alr test`, those main files are compiled and executed.

Sub-crate's `alire.toml` can contain test only dependencies, environment overrides, etc.

A test is considered PASSING when its exit code is 0, FAILING otherwise. This catches assertion fails (if they return non-zero exit codes, like in AUnit) and unhandled exceptions. (Maybe provide default `Assert` functions that exit with a non-zero code? or simply reuse AUnit's Assertions -> leave choice to the user with `alire.toml` test dependencies)

[...]

PR #1850, opened 2025-02-14
first draft of alire test runner

AldanTanneo:
This is a demo implementation of what a minimal Alire test runner could look like, following the proposed design in #1831.

For now, it only takes two arguments: a `-j[obs=]N` arg that specifies how many tests should be run in parallel (defaults to 0, aka how many cores your machine has) and `--dir=<dir>`, specifying which sub-directory holds your tests (defaults to `tests`).

### To test it:

Have a repository with a `tests` folder containing a binary crate. In the GPR file, replace the `Main` setting with
```ada
for Main use [Crate_Name]_List_Config.Test_Files;
```

and at the top, add

```ada
with "config/[crate name]_list_config.gpr";
```

Each `.adb` file in `src/` should contain a _single_ main procedure. You can add extra source directories (like `common`...), they will not be considered test folders, and you can put extra test code in there, to be used from the actual test procedures.

A test fails when its exit code is not zero (for instance, from an unhandled exception).

Finally, simply run `alr test2` with the alr from this branch.

### DIFF
--- a/src/alr/alr-commands-test2.adb
+++ b/src/alr/alr-commands-test2.adb
@@ -1,0 +1,57 @@
+with Alire.Directories;
+with Alr.Test_Runner;
+
+package body Alr.Commands.Test2 is
+   overriding procedure Execute
+     (Cmd : in out Command; Args : AAA.Strings.Vector)
+   is
+      use Alire.Directories;
+      use GNAT.Strings;
+
+      Subfolder : constant String :=
+        (if Cmd.Directory.all /= "" then Cmd.Directory.all else "tests");
+   begin
+      Cmd.Requires_Workspace;
+      Cmd.Set (Alire.Roots.Load_Root (Cmd.Root.Path / Subfolder));
+      Cmd.Requires_Workspace (Sync => True);
+
+      declare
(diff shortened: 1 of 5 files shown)


mosteo:
Thanks for providing this prototype, César. I'll play with it in the coming days.

AldanTanneo:
There is a follow up to this draft in [this branch](https://github.com/AldanTanneo/alire/tree/feature/test_runner_action), adding the test runner as a valid option in test actions, as well as a default test action in `alr init`:

```toml
[[actions]]
type = "test"
directory = "tests"
command = "alire_test_runner" # built-in minimal test runner
```

The modified `alr init` also creates a skeleton `tests` crate in a sub-directory.

I did not add this to the branch, because I don't think this is the right solution. Having a test runner running as an action makes it impossible to provide arguments to it, makes test output inconsistent and not really convenient, and has to integrate with the whole remote test machinery.

My opinion is still that the current test command should be moved to `index-test`, and even that test actions could be removed entirely for a better mechanism (cc @Fabien-Chouteau). I will try to implement my proposed designed in another separate branch.

PR #1862, opened 2025-02-21
feat: check executable existence prior to spawn

### LINKED ISSUE #1856: Improve error message when executable not found during spawns
See e.g. report here: https://github.com/alire-project/alire/issues/1852

mosteo:
Fixes https://github.com/alire-project/alire/issues/1856

##### PR creation checklist

### DIFF
--- a/TODO
+++ b/TODO
@@ -1,1 +1,0 @@
-Bump aaa/lml/yeison (wait for --format show to be merged and rebase on top of it)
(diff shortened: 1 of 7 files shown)


TODO
```diff
@@ -0,0 +1 @@
+Bump aaa/lml/yeison (wait for --format show to be merged and rebase on top of it)
```
mosteo: Bumping without waiting `aaa` requires bumping `lml` which requires adding `yeison` and that's a chore already done in #1851.  And `aaa` must be bumped because there is a `spawn` with the same issue inside it; not used by `alr`, but just in case.

Fabien-Chouteau (approved):

PR #1868, opened 2025-02-24
feat: proper `alr printenv` output even when sync needed

### LINKED ISSUE #1866: `alr printenv` includes unwanted preamble
**Checklist**

- [x] I've searched for similar issues and found none.
- [ ] I've included the output of `alr version`.
- [x] I've included complete steps to reproduce my issue.

**Describe the bug**

(originally reported by @manthonyaiello)

If the crate has not already been set up (e.g. by a prior call to `alr build`), `alr printenv` may print things like dependency information, toolchain installations etc. before printing the desired environment. This will cause issues if the output is executed using `eval "$( alr printenv )"` or similar.

**To Reproduce**

1. `git clone https://github.com/alire-project/hello`
2. `cd hello`
3. `alr printenv`
4. Observe `ⓘ Synchronizing workspace...` etc. before the actual environment is printed.

**Expected Behavior**

`alr printenv` should either handle the setup silently, or exit with an error if this is not possible. It appears this is already what happens with `alr -q printenv`, but I see no reason why the quiet switch should need to be specified explicitly.

**`alr` version**

Compiled from ad435d9

mosteo:
Fixes https://github.com/alire-project/alire/issues/1866

##### PR creation checklist

### DIFF
--- a/src/alire/alire-utils-user_input.adb
+++ b/src/alire/alire-utils-user_input.adb
@@ -75,8 +75,8 @@ package body Alire.Utils.User_Input is
       --  not want to miss any messages.
 
       if Alire.Log_Level < Simple_Logging.Debug then
-         Trace.Debug ("Enabling silent running");
-         --  If we are running with -vv, it's too late to remain silent anyway
+         Trace.Detail ("Enabling silent running");
+         --  If we are running with -v, it's too late to remain silent anyway
 
          Alire.Log_Level := Simple_Logging.Error;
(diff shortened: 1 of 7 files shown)


mosteo:
Hi Fabien, this is the last thing I had in the pipeline for 2.1. If you don't have anything that you want to go in, I'm ready for releasing.

Fabien-Chouteau (approved):

PR #1874, opened 2025-02-28
feature: test runner and command implementation

AldanTanneo:
Implement the test command design from #1831 

TBA: documentation and change list

##### PR creation checklist


AldanTanneo:
This is a follow-up PR to #1850, implementing the test runner with a dedicated `alr test` command. The previous test command is moved to `alr index-test`.

The runner can be configured with a `[test]` section in `alire.toml`, with the following default values:

```toml
[test]
runner = "alire"
directory = "tests"
jobs = 0
```

Running `alr init` now creates an additional subcrate named `tests` with an example test included, and the structure expected by the default alire test runner. This can be disabled with the `--no-test` option on `alr init` (implied by `--no-skel`).

doc/catalog-format-spec.md
```diff
@@ -397,7 +397,7 @@ static, i.e. they cannot depend on the context.
      solution.
 
    - `test`: the command is run on demand for crate testing within the Alire
```
Fabien-Chouteau, line 399: We might want to mark the test actions as obsolete given the new `alr test` command.
As far as I can tell all the test actions in the community index crates can easily be replaced by defining a custom runner for the new test command.

src/alire/alire-properties-tests.adb
```diff
@@ -0,0 +1,116 @@
+package body Alire.Properties.Tests is
+   overriding
+   function To_TOML (S : Settings) return TOML.TOML_Value is
```
Fabien-Chouteau, line 8: In terms of style, Alire follows the GNAT rule of having "sub-program boxes" before the body of a subprogram.
There's a shortcut in GNATstudio to make the box, I don't know if there's one in the VS Code plugin.

For instance here:
```
   -------------
   -- To_Toml --
   -------------
```

src/alire/alire-properties-tests.adb
```diff
@@ -0,0 +1,116 @@
...
+              From.Descend (Raw, "values");
+            Res   : Settings;
+            Val   : TOML_Value;
+         begin
+            if Local.Pop (TOML_Keys.Test_Runner, Val) then
+               if Val.Kind = TOML_String then
+                  if Val.As_String = "alire" then
+                     Res.Runner := (Kind => Alire_Runner);
+                  else
+                     Res.Runner :=
```
Fabien-Chouteau: I wonder if it would be better to only accept "Alire" as the valid value here. The custom runner should always be an array of strings, even if there is just one string in the array.

This will allow us to clearly distinguish between built-in and custom runners and, maybe, in the future, add different built-in runners without risking a conflict with existing manifests.

src/alire/alire-properties-tests.adb
```diff
@@ -0,0 +1,116 @@
...
+                        Command =>
+                          AAA.Strings.Empty_Vector.Append (Val.As_String));
+                  end if;
+               elsif Val.Kind = TOML_Array then
+                  declare
+                     Cmd : AAA.Strings.Vector := AAA.Strings.Empty_Vector;
+                  begin
+                     for I in 1 .. Val.Length loop
+                        if Val.Item (I).Kind /= TOML_String then
+                           Raise_Checked_Error
```
Fabien-Chouteau: Maybe use `Local.Checked_Error` here?

src/alire/alire-properties-tests.adb
```diff
@@ -0,0 +1,116 @@
...
+               end if;
+            else
+               Local.Checked_Error ("runner field required");
+            end if;
+
+            if Local.Pop (TOML_Keys.Test_Folder, Val) then
+               if Val.Kind /= TOML_String then
+                  Local.Checked_Error ("directory must be a string");
+               end if;
+               Res.Directory := Val.As_Unbounded_String;
```
Fabien-Chouteau, line 113: We have special type for relative paths and absolute paths, it makes sense to use a path type here and check if the string value is actually a valid relative path.

src/alire/alire-properties-tests.adb
```diff
@@ -0,0 +1,116 @@
...
+            else
+               Res.Directory := UStrings.To_Unbounded_String ("tests");
```
Fabien-Chouteau: There's already a definition of default value for `directory` and `jobs`. To avoid duplication, you should initialize `Res` with `Alire.Properties.Tests.Default` and remove the `else` branch here and below for jobs.

src/alire/alire-test_runner.adb
```diff
@@ -0,0 +1,232 @@
...
+          (Process_Id,
+           String,
+           "<" => Cmp);
+
+      Running_Tests : Map.Map := Map.Empty_Map;
+      Output_Files  : Map.Map := Map.Empty_Map;
+
+      procedure Spawn_Test (Test_Name : String) is
+         Exe_Name : constant String := Test_Name & Alire.OS_Lib.Exe_Suffix;
+         Filename : constant String := "output_" & Test_Name & ".tmp";
```
Fabien-Chouteau: This is going in the current working directory?

src/alr/alr-commands-test-runner.ads
```diff
@@ -0,0 +1,36 @@
+with AAA.Strings;
```
Fabien-Chouteau: What's the difference between this command and `alr test`?

AldanTanneo: This commands only runs the test runner in the current directory, without reading the test runner config in `alire.toml` and switching to any subfolder. It _could_ be useful if someone wants to run the test runner in a shell script.

testsuite/tests/test/crate-init/test.yaml
```diff
@@ -0,0 +1,4 @@
+driver: python-script
+indexes:
+    basic_index: 
+        in_fixtures: true
```
mosteo: This test seems to need no index, so it should use a configuration similar to the example in `testsuite/skels/no-index`

testsuite/tests/test/crate-init/test.py
```diff
@@ -0,0 +1,16 @@
...
+import os.path
+from drivers.alr import run_alr
+from drivers.asserts import assert_match
+
+run_alr("init", "--lib", "xxx")
+os.chdir("xxx")
+
+p = run_alr("test", complain_on_error=False)
+assert_match(".*\[ FAIL \] tests-example_test.*", p.out)
+# default test after init always fails
```
mosteo: Do we really want that all new crates come with a failing test? This is a chore we'll be imposing on all of our users.

testsuite/tests/test/default-success/test.py
```diff
@@ -0,0 +1,23 @@
...
+
+with open("./tests/src/tests-example_test.adb", "w") as f:
+   f.write("""procedure Tests.Example_Test is
+begin
+   null;
+end Tests.Example_Test;
+""")
+
+p = run_alr("test")
+assert_match(".*\[ PASS \] tests-example_test.*", p.out)
```
mosteo: At some not distant point I'd like to see timing of tests just like we have in e3-testsuite. All these lines will need modification; saying it now so maybe in the future there's less maintenance to do.

This is the kind of output we get with e3:

`PASS     00m00s help__equivalent_help`

testsuite/tests/test/filtering/test.py
```diff
@@ -0,0 +1,41 @@
+"""
+Filter test runs
+"""
```
mosteo: More explicit test purposes will help the reader/reviewer ;-)

mosteo:
I haven't yet dived into the details of the code, so here are only first general impressions after testing this PR and looking at the new entries in the testsuite.

First, this will be great. It will make writing simple tests so much more convenient. So, on that front, I'm happy to see it taking shape.

A word of caution: this PR is starting to do too many things at the same time. I'd advice a more piecemeal approach. I'll summarize at the end what I'd like to see here before the runaway gets bigger.

This PR already introduces two new commands, besides the new `test` behavior: `test-runner` and `index-test`. I'm not thrilled about this proliferation,  when `test --alire-runner` (instead of `test-runner`) and `test --from-index` (for the old remote behavior) can do. (And, this is personal, but I don't like two-word commands, if we can avoid them. And, in general, I'd like to keep the number of commands as contained as reasonably possible, in particular for closely related infrequent actions.)

Just to be clear, the `alr test <crate>` behavior is not intended to test the index, or do our custom test of a crate, but to do whatever the author of the crate wanted for testing. So it's conceptually equivalent to `alr get` followed by `alr test`. That's to frame my previous and following comments.

For the sake of initial clarity in this PR, it is OK to temporarily have all the old behavior under `index-test`, but before merging, I'd like to see a more conservative approach: `alr test` by itself should indeed do the new testing, but with fallback to the old build (informing about it) if no tests are found. `alr test <crate>` can be now `--from-index` and incompatible with filtering.

(Side note: since filtering only applies to the builtin runner, maybe those should simply be considered general arguments for whatever runner is invoked.)

It is OK and welcomed to move the tests for legacy alr test behavior to its own category (`test-legacy` rather than `index-test`). Now there's some confusion because `index-test` tests are testing the local crate with the existing behavior, which is unrelated to the index.

Running `alr test` in a crate without `tests` dir causes an unmanaged error:
```
$ alr test
no primary runner defined, using the builtin test runner
error: directory "tests" does not exist
error: alr encountered an unexpected error, re-run with -d for details.
error: error location: 0x5dbb38d3dec2 ada__directories__set_directory at ???
```

This is were the old behavior should kick in, with a warning.

In summary:
- Please don't add any more features here; keep those for follow-ups (filtering, for example could be separate, but since it is already in, OK).
- Rename `testsuite/tests/index-test` to `testsuite/tests/test-legacy`.
- Replace `test-runner` with `test --alire-runner` (this is another thing that could have gone in a follow-up PR).

Then there is the issue of making test actions obsolete. Having two mechanisms is certainly not nice. Having a dedicated `[test]` is reasonable given the importance of testing, although we are losing the ability to have several different test mechanisms at the same time, so maybe `[[test]]` should be allowed too (or be the only option). The error above talks about a "primary" runner, are there secondary ones? I think this point merits some more discussion.

mosteo:
To save you some effort, I think I'd prefer to entirely drop the remote testing behavior rather than moving it elsewhere, as it doesn't make much sense in `alr` (I mean the `--full`, `--redo`, etc). What should remain as a fallback is our current default test of building the crate, to be able to have a smooth transition for our online checks at `alire-index` and alire-crate-ci.ada.dev.

src/alire/alire-properties-tests.ads
```diff
@@ -0,0 +1,93 @@
...
+
+   function Jobs (S : Settings) return Natural;
+
+   function Default return Settings;
+
+private
+
+   type Settings is new Properties.Property with record
+      Runner    : Runner_Type;
+      Directory : UString;
```
mosteo: s/UString/Unbounded_Relative_Path/

src/alire/alire-properties-tests.ads
```diff
@@ -0,0 +1,93 @@
...
+
+   overriding
+   function To_Yaml (S : Settings) return String;
+
+   function From_TOML
+     (From : TOML_Adapters.Key_Queue) return Conditional.Properties;
+
+   function Runner (S : Settings) return Runner_Type;
+
+   function Directory (S : Settings) return UString;
```
mosteo: s/UString/Unbounded_Relative_Path/

src/alire/alire-test_runner.ads
```diff
@@ -0,0 +1,13 @@
+with Alire.Roots;
+
+with AAA.Strings;
+
+package Alire.Test_Runner is
+   procedure Run
+     (Root   : in out Alire.Roots.Root;
+      Filter : AAA.Strings.Vector := AAA.Strings.Empty_Vector;
+      Jobs   : Natural := 0;
+      Fails  : out Integer);
```
mosteo: Why not a function for the single result?

src/alr/alr-commands-init.adb
```diff
@@ -331,6 +332,78 @@ package body Alr.Commands.Init is
...
+         TIO.Close (File);
+
+         if not Create (+Full_Name (Test_Directory / "tests.gpr")) then
+            Trace.Warning ("Could not create project file 'tests/tests.gpr'");
+            return;
+         end if;
+         Put_Line ("with ""config/tests_config.gpr"";");
+         Put_Line ("with ""config/tests_list_config.gpr"";");
+         Put_New_Line;
+         Put_Line ("project Tests is");
```
mosteo: I wonder if naming the test crate/project always `tests` can come and bite us later, or should we do like `xxx_config` and use `xxx_tests`. @Fabien-Chouteau , can you think of some scenario where two different testsuites come into the same scope? I don't right now, but...

src/alr/alr-commands-init.adb
```diff
@@ -331,6 +332,78 @@ package body Alr.Commands.Init is
...
+         pragma Style_Checks ("M100");
+
+         Test_Directory.Make_Dir;
+         if not Create (+Full_Name (Test_Directory / "alire.toml")) then
+            Trace.Warning ("Could not create test crate skeleton");
+            return;
+         end if;
+         Put_Line ("name = 'tests'");
+         Put_Line ("description = ''");
+         Put_Line ("version = '0.0.0-test'");
```
mosteo, line 363: We start at `0.1.0-dev` for our generated crates, could be the same here.

src/alr/alr-commands-init.adb
```diff
@@ -331,6 +332,78 @@ package body Alr.Commands.Init is
...
+         Put_New_Line;
+         Put_Line ("[[depends-on]]");
+         Put_Line (To_String (Info.Name) & " = '*'");
+         Put_New_Line;
+         Put_Line ("[[pins]]");
+         Put_Line (To_String (Info.Name) & " = { path = '..' }");
+         Put_New_Line;
+         Put_Line ("[build-profiles]");
+         Put_Line (To_String (Info.Name) & " = 'validation'");
```
mosteo: I wonder if this should be `"*" = validation` so any eventual dependencies are also built in validation mode. On the one hand, we're not testing those dependencies; on the other, one could want for the whole system to be validated. I don't really know what are the industry practices in this regard. @Fabien-Chouteau ?

src/alr/alr-commands.adb
```diff
@@ -44,6 +45,7 @@ with Alr.Commands.Search;
 with Alr.Commands.Settings;
 with Alr.Commands.Show;
 with Alr.Commands.Test;
+with Alr.Commands.Test.Runner;
```
mosteo: `Alr.Commands.Test` is redundant with `Alr.Commands.Test.Runner`.

mosteo (commented):
I have done a first pass through the code, looking good, I just left a few minor observations and a couple general questions.

src/alr/alr-commands.adb
AldanTanneo: Is it? `alr test` and `alr test-runner` are two different commands.

Update: caught up on the rest of the reviews, will remove the `test-runner` command and use a flag on `alr test` (maybe in another PR).

AldanTanneo:
> * Please don't add any more features here; keep those for follow-ups (filtering, for example could be separate, but since it is already in, OK).
> * Rename `testsuite/tests/index-test` to `testsuite/tests/test-legacy`.
> * Replace `test-runner` with `test --alire-runner` (this is another thing that could have gone in a follow-up PR).

Will do :)

> Then there is the issue of making test actions obsolete. Having two mechanisms is certainly not nice. Having a dedicated `[test]` is reasonable given the importance of testing, although we are losing the ability to have several different test mechanisms at the same time, so maybe `[[test]]` should be allowed too (or be the only option). The error above talks about a "primary" runner, are there secondary ones? I think this point merits some more discussion.

The 'primary' test runner mention is because at some point I thought about making test actions run alongside the main test runner defined in the crate. I will change the wording of that error message.

src/alr/alr-commands-init.adb
AldanTanneo: The user can rename this to anything as long as they follow the test folder structure; maybe adding a `--test-crate-name <something>` (not a good name for a flag, but for the sake of argument) to customize the behaviour of `alr init`?

src/alr/alr-commands-init.adb
AldanTanneo: My reasoning was that the test crate version would probably not change as it is only relevant in the context of the main crate. But that's open to change.

AldanTanneo:
> To save you some effort, I think I'd prefer to entirely drop the remote testing behavior rather than moving it elsewhere, as it doesn't make much sense in alr (I mean the --full, --redo, etc). What should remain as a fallback is our current default test of building the crate, to be able to have a smooth transition for our online checks at alire-index and alire-crate-ci.ada.dev.

To be clear: I can delete the `index-test` command and related tests, as long as I add the fallback to building the crate when no test runner is declared? Or is it when no test runner is declared AND the `tests` folder does not exist?

AldanTanneo:
> (Side note: since filtering only applies to the builtin runner, maybe those should simply be considered general arguments for whatever runner is invoked.)

This is the intended behaviour indeed, and it was an oversight that command arguments were not passed to the test runner command if it is an external runner :)

src/alr/alr-commands-init.adb
Fabien-Chouteau: Tests are not supposed to be published in the index so I don't think this will be a problem. As @AldanTanneo said, people can always rename the crate/project.

mosteo: I actually found a real use case (by me). It's a situation where I'm simultaneously developing a few pinned crates. I currently have a master crate that pins all those crates, and actually I pin their test projects (the ones I'm currently doing by hand and would be replaced with this) rather than the main crate so all tests are kept in sync with their main libraries. I gain visibility of the main libraries through the pins in their test crates, so I edit the main libs but also the tests when I add new features or tweak things.

I think this would be impossible if all of them are named `tests`. You can see for real in this repo where I maintain Ada 2012 and Ada 2022 variants. I had to rename the tests of one of them: https://github.com/mosteo/yeison/blob/devel/dev/alire.toml

Having an extra parameter to tweak the initialization could do, but I'd rather have a default that I won't regret later. Renaming a crate is a chore involving the gpr and all files due to package renaming (which GNAT studio never did properly for me).

(ETA: the folder can always be `tests`, it's the crate inside that would be `mycrate_tests`.)

src/alr/alr-commands.adb
mosteo: > Is it? alr test and alr test-runner are two different commands.

Yes, withing a child gives you visibility of all its parents. It's a minor nitpick, but still you never know when a future GNAT with all warnings enabled is going to bite you. (Actually I thought this was already caught by some GNAT flag.)

AldanTanneo: I didn't know that, thanks!

The `Test.Runner` command will be gone in the commits I'm currently preparing, so this will be solved anyway :)

testsuite/tests/test/crate-init/test.py
AldanTanneo: I didn't want users to inadvertently commit the 'example_test' file, so that's why I made it fail by default. I considered not putting an example test at all, although I like the self-documenting effect...

mosteo:
> To be clear: I can delete the `index-test` command and related tests, as long as I add the fallback to building the crate when no test runner is declared?

That was the idea, but I'm still ambivalent about it. For now, let's just have the fallback in `test` and the remote things in `index-test` and we can deal with that afterwards if necessary.

src/alire/alire-test_runner.adb
AldanTanneo:
```suggestion
         Filename : constant String :=
            Root.Working_Folder / ("output_" & Test_Name & ".tmp");
```

@mosteo is the `alire` folder acceptable for temporary test output files? If nothing goes catastrophically wrong they should always be cleaned up.

mosteo: Yes, it's actually the [preferred location for our temp](https://github.com/alire-project/alire/blob/f5c4bb3fb4116b1b88bd564d7b3a64e7fc981e64/src/alire/alire-directories.adb#L600) files when it is locatable (actually we use `alire/tmp`).

Seb-MCaw:
This looks great.

A few very subjective comments I have already discussed with @AldanTanneo directly, but stated here for the record:

- I agree we should allow `[[test]]` (or `[[tests]]`?).
- Personally, I don't like that `runner = "alire"` and `runner = ["alire"]` have completely different effects, or that actions use the `command` key but a test uses `runner` to mean the same thing. My vote is therefore to use different keys (i.e. `runner = "alire"` but `command = ["cmd", "arg1", "arg2"]`) to highlight the conceptual difference between a built-in runner and an external command.
- I'm not a fan of invisible default behaviour, and I think including a `[test]` section in the default manifest would be somewhat self-documenting.
- Backwards compatibility (by treating an `[[actions]]` with `type = "test"` as equivalent to `[[tests]]`) would be nice, but only if it's not too much work.

src/alr/alr-commands-init.adb
Fabien-Chouteau: It would say that it's highly recommended **not to** build dependencies you don't own or develop into validation mode.

Validation turns on all warnings and treats them as errors. New warnings can be introduced from one version of the compiler to another, which means you cannot guarantee that the code will actually build.

testsuite/tests/test/default-success/test.py
AldanTanneo: Currently there are only two files with a specific regex to match the output. I think it's best to leave it as is and change it when needed? I'm wary of adding support for a feature in the future and then implementing it in a different way that would require changes anyway

src/alr/alr-commands-init.adb
AldanTanneo: It's easy to edit this or add build overrides, and building the tested crate in validation mode seems a sensible default, so I will leave this as-is for now.

AldanTanneo:
> * I agree we should allow `[[test]]` (or `[[tests]]`?).

Allowing several test runners makes us lose the ability to pass command line arguments to them, which imo is more valuable. If there are several test runners, the tests are not much more than an enumeration of commands, and would be better served by a shell script.

> * Personally, I don't like that `runner = "alire"` and `runner = ["alire"]` have completely different effects, or that actions use the `command` key but a test uses `runner` to mean the same thing. My vote is therefore to use different keys (i.e. `runner = "alire"` but `command = ["cmd", "arg1", "arg2"]`) to highlight the conceptual difference between a built-in runner and an external command.

that's fair; in the next couple commits I already removed the possibility to specify a command as a single string (commands are always arrays and builtins are always strings).

> * I'm not a fan of invisible default behaviour, and I think including a `[test]` section in the default manifest would be somewhat self-documenting.

The behaviour I'm going with is the one @mosteo recommended above: if no runner is defined, build the crate. `alr init` will include a small `[test]` section with `runner = "alire"`.
 
> * Backwards compatibility (by treating an `[[actions]]` with `type = "test"` as equivalent to `[[tests]]`) would be nice, but only if it's not too much work.

Same comment as above for `[[tests]]` vs `[test]`... I don't think the tradeoff is worth it, and test actions still exist under the `index-test` command, even though it will probably end up being removed. In the current alire index, I don't think I saw a single crate (or maybe one) using more than one test action.

AldanTanneo:
Rebased and added commits, addressed a lot of comments (but not all)

I don't understand the test failure however:

```
before (last 100 chars): b'values:\r\nERROR:    test:\r\nERROR:    values:\r\nERROR:    forbidden extra entries: maintainers-logins\r\n'
```

src/alr/alr-commands-init.adb
AldanTanneo: Addressed in [this commit](https://github.com/alire-project/alire/pull/1874/commits/92167510c43a0d27289fe8491ded88e55ff29150)

src/alire/alire-properties-tests.adb
AldanTanneo: Used a path type, did not implement manual check however; maybe check at runtime if the folder exists before doing anything?

src/alire/alire-properties-tests.adb
AldanTanneo: Only accepts `"alire"` as a string now, however check @Seb-MCaw's comment for distinction between a `command` and `runner` field? It would probably be better for consistency.

Seb-MCaw:
Ah, I see. I had completely missed the implications for command line arguments. In that case, a single `[test]` indeed seems the best approach, since wrapping with a shell script should require minimal effort.

> I don't understand the test failure however

I wrote that test, so it's probably my fault; I'll have a look.

Seb-MCaw:
Yes, the problem is [here](https://github.com/alire-project/alire/blob/92167510c43a0d27289fe8491ded88e55ff29150/testsuite/tests/publish/private-indexes/test.py#L72) where I just lazily append to the manifest. This now ends up under the `[test]` table where it is invalid.

You should just be able to change (at line 67)
```python
    run_alr("init", "--bin", "xxx")
    os.chdir("xxx")
```
to
```python
    init_local_crate()
```
with the necessary `from drivers.alr import init_local_crate` (which is the approach I should have used in the first place).

testsuite/tests/test/default-success/test.py
mosteo: OK, that's reasonable

mosteo:
Hi @Seb-MCaw, welcome to the party :)

> I agree we should allow [[test]] (or [[tests]]?).

I read @AldanTanneo's reply and I agree that passing arguments to the testsuite is handy; I often use it with alr's testsuite. So we must preserve that. I was leaning towards `[test]` too, but read on.

> Personally, I don't like that runner = "alire" and runner = ["alire"] have completely different effects [...]

Agreed.

> I'm not a fan of invisible default behaviour, and I think including a [test] section in the default manifest would be somewhat self-documenting.

I too prefer this, it's good return in exchange for a few lines. It's also a way of clearly separating the legacy behavior. But I also just realized that this will break new crates for users of `alr<3.0`, as they won't be able to load the manifest, which is something we normally aren't willing to do. Alas, a major release is a major release and we can take the opportunity to implement measures for the future.

> Backwards compatibility (by treating an [[actions]] with type = "test" as equivalent to [[tests]]) would be nice, but only if it's not too much work.

Preserving legacy behavior of test actions in absence of `[test]` should be trivial, see later.

Moving onto @AldanTanneo's reply.

>  If there are several test runners, the tests are not much more than an enumeration of commands, and would be better served by a shell script.

There's a legitimate use case I've found in our index, see at the end.  And, even if it is a matter of running a few different commands, it is nice not having to write script cases for multiple OSes, or introducing some portable dependency like Python. As an Ada programmer, I don't like having to force Alire users to use other technologies when a little effort on our part is all it takes.

> The behaviour I'm going with is the one @mosteo recommended above: if no runner is defined, build the crate.

Actually (maybe I worded it poorly) what I wanted is to keep the old behavior when no `[test]` or legacy test action is defined, which is a build, but with extras. `alr test` (legacy) generates a log inside `alire` that we are using in places, as it captures all output from the build, including actions and errors, and it also generates a `junit` compatible XML report (this one is less important).

So instead of calling `Roots.Build`, when there's no `[test]`, you can just divert to `Index_Test.Execute` to preserve all of the old behavior.

> Same comment as above for [[tests]] vs [test]... I don't think the tradeoff is worth it, and test actions still exist under the index-test command, 

I insist, test actions aren't only for testing the index, and so `index-test` is ill-named. Test actions are run by `alr test` when invoked without a crate argument and until this PR goes in, its main function is the same function as this PR: testing the local crate. It makes no sense running `alr index-test` to run actions of the local crate. This is why all the old behavior cannot simply be lumped into a new `index-test` (although, as I said in my previous post, let's leave it as that for now).

> In the current alire index, I don't think I saw a single crate (or maybe one) using more than one test action.

There's only one AFAIK, `hirtos` ([here](https://github.com/alire-project/alire-index/blob/stable-1.3.0/index/hi/hirtos/hirtos-2.0.0.toml)). But the use case it's compelling: it's running two different test crates for different configurations. And this is something that won't be possible with the new runner, as one test crate implies one configuration for all the tests. So we should carefully consider the possibility of needing several test crates even with the builtin runner.

Actually, test actions became necessary for crates that don't have a complete configuration (which the client crate must supply; in this case, the test crate), or that require a cross-compiler which cannot be specified either in the main crate, as the library is compiler-independent (I suspect this is the main use case behind test actions). I guess it isn't that far-fetched wanting to test several crate configurations, particularly in embedded.

What about this:

- Use `[[test]]` instead of `[test]` (or allow both if there's only one).
- Have an optional field, `name`, or `id` or somesuch.
- Allow running only one entry with `alr test --id=blah`, and extra arguments are passed on.
- Allow running all tests with `alr test`, and arguments can't be passed if there's more than one entry.

This would work for the common use case of a single entry without any extra complications for the user, but allow multiple test crates, or multiple test runners, and running things in isolation which I guess is the desired effect during development (a few tests in a particular runner), or running everything.

AldanTanneo:
Something I suggested to Fabien but didn't end up implementing:

What if test actions were to be run alongside the runner in `[test]`, explicitely as '**secondary**' test runners. The main one, in the `[test]` section, would be the **primary** test runner, and the one you could pass arguments to when using `alr test`.

Passing arguments to secondary test runners would be done with `alr test --action=<action name> [args...]` (syntax to be refined, but essentially have an extra flag on `alr test`).

This would preserve the previous default behaviour: when no test runner is defined, build the crate, (and/or) run the test actions; I'd go towards build AND run as it would be more consistent with the behaviour when a test runner _is_ present, which is: run the primary runner, then run the actions (order to be refined? pre-test, post-test? I'm worried of complexity creep)

mosteo:
I think that's overcomplicating, specially if we mark test actions as obsolescent. In particular, I don't think test actions and test runners should coexist in a new crate. It's one or the other; this way the legacy behavior for old releases can be preserved without doing anything, and new crates will have a single test by default and use the new functionality. We can err if both test actions and test runners are defined.

Do you see some problem with my last proposal? It's similar to what you say and I think it preserves all we want: simplicity in the general case, generality for when several test crates are needed, and a way to pass arguments in every case for interactive use.

testsuite/tests/test/crate-init/test.py
mosteo: I see your point but, to me, always having to remove or rename+edit that test looks painful. Maybe if we had some way of having a XFAIL status?

src/alr/alr-commands-init.adb
mosteo: Yes, the doubt wasn't about the tested crate but about dependencies, and Fabien gave the rationale for leaving it as is.

testsuite/tests/test/crate-init/test.py
AldanTanneo: I think I will just change this to PASS, so it's less of a pain point. Unless we want to go all the way to add annotations to test files, like a magic `--  XFAIL` comment on the first line, or a filename suffix.

mosteo: No magic for now, we can always revisit :)

AldanTanneo:
Game plan:

- `for Test_Setting of All_Settings loop [run test runner like before]`
- arg forwarding: `(if All_Settings.Length > 1 then Empty_Vector else Args)` (with a warning and a help message when several runners are present and the user tries to pass extra arguments)
- `if All_Settings.Is_Empty then Execute_Legacy (Cmd)`
- add support in `Properties.Test` for several runners... probably a small change but that part of the code base is a bit arcane to me, I'd appreciate some pointers :)

mosteo:
> * add support in `Properties.Test` for several runners... probably a small change but that part of the code base is a bit arcane to me, I'd appreciate some pointers :)

I'd mimic what's done for actions, starting at the `From_TOML` in `Alire.Properties.Actions.Runners`. If it proves too confusing you can leave that part to me.

AldanTanneo:
I must still point out that to me, having several test runners goes against the design idea. It's basically remaking test actions. The point is to have a minimal default, that can be disabled and switched for a better option the minute it doesn't fit the user's needs.

Also, the `[test]` section can be used in combination with `'case(os)'` annotations and the like, making it easy to have scripts on different platforms:

```toml
[test.'case(os)']
linux = { runner = ["script.sh"] }
windows = { runner = ["script.cmd"] }
# etc...
```

I'll still implement the several runners idea, but I feel like it's not the right direction. I think @Fabien-Chouteau is of a somewhat similar mind (but I could be wrong).

mosteo:
To me, a table vs an array is a secondary detail. It won't affect 99% of users, and those needing it can make use of it. It's not complicating the basic use case. I don't see anything in my proposal that worsens the behavior you have implemented.

This is something I wrote and removed in previous replies not to broaden the topic even more, but precisely having to write custom scripts per OS is a reason to have a way to avoid them for trivial cases like running two crates rather than one. If we can go an extra mile to keep users in Alire/Ada land, I welcome it.

Incidentally, Fabien joked at FOSDEM that maybe we should embed HAC in Alire for portable Ada scripting... and maybe there's something to the idea ;-)

But, seriously: this PR mustn't do everything at once. You can keep the implementation at a single table. Any extensions can come later.

AldanTanneo:
Fair enough, for now I've added the possibility in the command to run several test runners with said warning, and added some very minimal 'execute legacy' code; I don't do capture however, I think that should be external (the user should redirect output to a file if they want logs), and there is no good way to capture build output without spawning a new `alr` process.

mosteo:
So the PR is ready for final review?

AldanTanneo:
now yes I believe :)

(I will add extensive documentation of the new behaviour, and continue the multiple runners idea, in another PR)

src/alire/alire-test_runner.adb
```diff
@@ -0,0 +1,285 @@
...
+   ---------------------
+
+   procedure Create_Gpr_List
+     (Root : Alire.Roots.Root; List : AAA.Strings.Vector)
+     --  Create a gpr file containing a list of the test files
+     --  (named `Test_Files`).
+
+   is
+      File_Path : constant Alire.Absolute_Path :=
+        Root.Path / "config" / (Root.Name.As_String & "_list_config.gpr");
```
mosteo: Please add an `Alire.Paths.Default_Config_Folder` and refactor this literal string and the one in `Alire.Properties.Configurations`.

src/alire/alire-test_runner.adb
```diff
@@ -0,0 +1,285 @@
...
+
+   procedure Run_All_Tests
+     (Root : Alire.Roots.Root; Test_List : AAA.Strings.Vector; Jobs : Positive)
+   is
+      use GNAT.OS_Lib;
+
+      function Cmp (A, B : Process_Id) return Boolean
+      is (Pid_To_Integer (A) < Pid_To_Integer (B));
+
+      package Map is new
```
mosteo, line 139: Name boxes here. We really need an style guide and some automated checking, sorry about that.

src/alire/alire-test_runner.ads
```diff
@@ -0,0 +1,12 @@
...
+
+with AAA.Strings;
+
+package Alire.Test_Runner is
+   function Run
+     (Root   : in out Alire.Roots.Root;
+      Filter : AAA.Strings.Vector := AAA.Strings.Empty_Vector;
+      Jobs   : Natural := 0) return Integer;
+   --  Run all .adb files in the `src` folder of the given root as
+   --  separate tests. Return the number of failing tests.
```
mosteo, line 12: One whitespace line around the function. Ditto about the style guide ^^'

src/alr/alr-commands-init.adb
```diff
@@ -58,6 +58,7 @@ package body Alr.Commands.Init is
       Src_Directory : constant Virtual_File := Directory / "src";
       Share_Directory : constant Virtual_File :=
          Directory / "share" / Filesystem_String (Lower_Name);
+      Test_Directory : constant Virtual_File := Directory / "tests";
```
mosteo: Please add an `Alire.Paths.Default_Tests_Folder` for this `"tests"`.

src/alr/alr-commands-test.adb
```diff
@@ -1,718 +1,155 @@
...
+      end if;
 
-         function Is_Match (Name : Alire.Crate_Name) return Boolean is
-           (for some I in Args.First_Index .. Args.Last_Index =>
-               AAA.Strings.Contains (+Name, Args (I)));
+      if not Args.Is_Empty and then All_Settings.Length > 1 then
+         Trace.Warning
+           ("arguments cannot be forwarded to test runners when several "
+            & "exist.");
+         Trace.Warning ("use --id=<runner id> to call a specific test runner");
```
mosteo: This `--id` isn't implemented yet, right? So this message is premature.

testsuite/tests/action/masked-error/test.py
```diff
@@ -20,9 +20,9 @@ def check_errors():
 
 # Verify a successful action
 add_action("test", ["echo", "OK"])
-run_alr("test")
+run_alr("index-test")
```
mosteo: This should work with the default `test` now, right?

testsuite/tests/test/custom-runner/test.py
```diff
@@ -0,0 +1,35 @@
...
+
+import os.path
+import uuid
+
+from drivers.alr import run_alr
+from drivers.asserts import assert_match
+from drivers.helpers import replace_in_file
+
+run_alr("init", "--lib", "xxx", "--no-test")
+os.chdir("xxx")
```
mosteo: init_local_crate()

testsuite/tests/test/default-failure/test.py
```diff
@@ -0,0 +1,23 @@
...
+Run a failing test and check it is detected correctly
+"""
+
+import os.path
+
+from drivers.alr import run_alr
+from drivers.asserts import assert_match
+
+run_alr("init", "--lib", "xxx")
+os.chdir("xxx")
```
mosteo: Idem

src/alr/alr-commands-test.adb
```diff
@@ -1,718 +1,155 @@
...
-               end;
-            end loop;
-         end if;
-      end Find_Candidates;
+            Dir      : constant Alire.Relative_Path := To_String (S.Directory);
+            Failures : Integer;
 
-   begin
-      --  Validate command line
+            Guard : Alire.Directories.Guard (Enter (Dir))
```
mosteo: Here, if the tests dir doesn't exist, it ends with  with
```
error: directory "tests" does not exist
error: alr encountered an unexpected error, re-run with -d for details.
error: error location: 0x5736818265a2 ada__directories__set_directory at ???
```
It would be nicer to just point out the missing dir and not consider this unexpected. Thinking of users that may edit the test directory.

src/alire/alire-test_runner.adb
```diff
@@ -0,0 +1,285 @@
...
+      Trace.Info ("Building tests");
+      if Alire.Roots.Build (Root, AAA.Strings.Empty_Vector) then
+         Trace.Info ("Running" & Test_List.Length'Image & " tests");
+         Run_All_Tests (Root, Test_List, Job_Count);
+
+         Trace.Always ("Total:" & Driver.Total_Count'Image & " tests");
+         Ada.Text_IO.Flush;
+         if Driver.Fail_Count /= 0 then
+            Trace.Error ("failed" & Driver.Fail_Count'Image & " tests");
+         end if;
```
mosteo: I'd like a success alternative with `Alire.Put_Success`, "Test run completed successfully" or somesuch.

mosteo (changes requested):
Very minor things for the most part, I'm eager to start using this.

AldanTanneo:
CI fails before reaching tests:

```
ERROR: Checkout of repository at https://github.com/mosteo/yeison failed, re-run with -vv -d for details
```

AldanTanneo:
Btw: I checked and apparently, this works already:

```toml
[[test]]
runner = "alire"

[[test]]
runner = ["echo", "OK"]
```

There needs to be some work to make the output display a bit clearer, but no work required on the properties side apparently!

src/alire/alire-properties-tests.ads
```diff
@@ -0,0 +1,93 @@
...
+
+   function Directory (S : Settings) return Unbounded_Relative_Path
+   is (S.Directory);
+
+   function Jobs (S : Settings) return Natural
+   is (S.Jobs);
+
+   function Default return Settings
+   is (Properties.Property
+       with (Kind => Alire_Runner), UStrings.To_Unbounded_String ("tests"), 0);
```
mosteo: I guess this one should come form `Alire.Paths` too.

src/alire/alire-properties-tests.adb
```diff
@@ -0,0 +1,120 @@
...
+            if Local.Pop (TOML_Keys.Test_Runner, Val) then
+               if Val.Kind = TOML_String then
+                  if Val.As_String = "alire" then
+                     Res.Runner := (Kind => Alire_Runner);
+                  else
+                     Local.Checked_Error
+                       ("invalid builtin runner (accepted values: 'alire'). "
+                        & "External runners must be an array of strings.");
+                  end if;
+               elsif Val.Kind = TOML_Array then
```
mosteo, line 81: I thought the idea was to have both `runner` and `command` to clearly separate builtin runners from custom commands?

src/alire/alire-properties-tests.adb
mosteo: The error is quite obscure without a manual check that the path is relative:
```
error: Cannot continue without a workspace:
error:    Failed to load alire.toml:
error:    Dynamic_Predicate failed at alire-properties-tests.adb:97
```
As for if the path exists, a warning would be fine.

mosteo (changes requested):
I fixed the yeison dependency, so you'll need to pull or merge.

I commented on a couple of things I found in this round.

As for the arrays already working, I didn't immediately remember that, yes, by default a property can be either a single value or an array, the parser resolves expressions for each item and dispatches to the final loader, and cardinality is enforced elsewhere. So your `From_TOML` gets called several times transparently.

AldanTanneo:
Regarding that last comment (because apparently I can't reply in the thread? GitHub is a mystery sometimes), I don't understand why the predicate on `Unbounded_Relative_Path` is done with `Check_Absolute_Path` (and why it works at all when the paths I tested with are not absolute).

From `alire.ads`:

```ada
subtype Unbounded_Relative_Path is UString
  with Dynamic_Predicate =>
    +Unbounded_Relative_Path = "" or else
  not Check_Absolute_Path (+Unbounded_Relative_Path);
```

src/alire/alire-properties-tests.adb
AldanTanneo: Implemented the two fields, with checks that exactly one of them is present

src/alire/alire-properties-tests.adb
AldanTanneo: Added a check with a better error message. I didn't implement the warning if the path does not exist; that check is done at runtime. A warning would be too noisy and would show up every time the manifest is parsed.

mosteo:
> I don't understand why the predicate on `Unbounded_Relative_Path` is done with `Check_Absolute_Path`

`Relative_Path` doesn't bring in any checks by itself, it's used to document. Absolute paths, OTOH, are actually checked to be so. So the check on unbouded relative path is by exclusion, that the path is not absolute.

AldanTanneo:
> > I don't understand why the predicate on `Unbounded_Relative_Path` is done with `Check_Absolute_Path`
> 
> `Relative_Path` doesn't bring in any checks by itself, it's used to document. Absolute paths, OTOH, are actually checked to be so. So the check on unbouded relative path is by exclusion, that the path is not absolute.

Understood that right after and deleted the comment :P reading is hard

mosteo (approved):

mosteo:
I'm going to merge now, thanks for all the effort @AldanTanneo . We'll need to follow up on a few points (like documentation), I will create issues for those.

Issue #1881, opened 2025-03-10
New `alr test` with built-in runner


mosteo:
This is an umbrella issue for follow-ups to the main feature in https://github.com/alire-project/alire/pull/1874.

AldanTanneo:
Question:

Currently, the default behaviour in `alr init` is to name the tests subcrate `<cratename>_tests`. This is a good default for the gprfile and such. However, it's a long name for the package name, especially if the convention is to have tests files like `<cratename>_tests-mytest.adb`. Would it be acceptable to keep the crate name as is, but to have a different package name? The test crate should never be used as a dependency for anything, so naming the package `Tests` or `Test` could be nicer:

```ada
procedure Tests.My_Test_Name is ...
```

Anyway, this is kind of a nitpick but it could be good for usability.

mosteo:
I think this would cause the same problem I indicated when editing several crates at the same time via pins to their testsuites. There will be a name clash in the root test package as the files will be duplicated.

AldanTanneo:
> I think this would cause the same problem I indicated when editing several crates at the same time via pins to their testsuites. There will be a name clash in the root test package as the files will be duplicated.

I see (although I don't really understand the point of pinning testsuites, but that's probably because I haven't used Alire enough with multiple crates). I think there should be a way to easily rename the testsuite, or to set the name at `init` time then.

Fabien-Chouteau:
Hello here; as I was writing some tests the other day, I realized that the simple test case detection scheme might be a bit of a problem for some tests.

Right now, all .adb files are considered test cases and must contain a single "main" procedure. But in Ada, you often have to declare stuff at the "library level," i.e., inside a package, and not in the declarative part of a procedure. For example, access subprograms. So there is often the need to declare a package (spec and body) for the test, in that case, there will be a `.adb` file for the package and the runner will consider this `.adb` as a test case.

@AldanTanneo and I discussed various solutions and we do not agree on the way to go. 

On one hand, @AldanTanneo considers that the built-in test runner should remain as simple as possible. More advanced features would be provided by external runners, potentially developed within the alire-project umbrella but not in the Alire code base. So, the use case presented above would simply not be covered by the built-in runner.

I am more in favor of improving the built-in test runner to handle more complexity, such as the case presented here.
For the problem at hand, we could require a prefix in the file name of test cases (e.g. `.*tc_.*\.adb`), only the `.adb` files matching the pattern would be considered test cases. 

[...]

PR #1887, opened 2025-03-10
fix: minor issues with dynamic properties

mosteo:
I identified these deficiencies while reviewing #1881:

- We weren't preventing case expressions for properties that shouldn't allow them.
- We were checking cardinality of properties only for a subset of properties (those implemented as 'labeled').
- Export of the new `[test]` property would fail if containing dynamic expressions.

##### PR creation checklist


src/alire/alire-releases.adb
```diff
@@ -1187,20 +1189,26 @@ package body Alire.Releases is
...
-            end case;
-         end if;
+      for Prop in Alire.Properties.From_TOML.Cardinality'Range loop
+         declare
+            Toml_Key : constant String := Tomlify (Prop).As_String;
+         begin
+            if Root.Has (Toml_Key) then
+               case Alire.Properties.From_TOML.Cardinality (Prop) is
+                  when Unique | Multiple_Deprecated =>
+                     pragma Assert
```
mosteo: Change assert to checked error

src/alire/alire-properties-from_toml.ads
```diff
@@ -79,6 +79,49 @@ package Alire.Properties.From_TOML is
...
+   --  Cardinalities say if a property must be a single value or an array.
+   --  Since we store properties always as elements in a list, we lose this
+   --  information after load so we use this not only during loading to enforce
+   --  index correctness, but also during exporting to ensure the proper type
+   --  (atom/array) is created.
+
+   type Cardinalities is
+     (Unique,
+      --  Must be a single value
+      Multiple_Deprecated,
```
AldanTanneo: The licenses array form has been deprecated since Alire 1.0, maybe it could be removed in Alire 3.0?

testsuite/tests/test/conditional-tests/test.py
```diff
@@ -0,0 +1,37 @@
...
+from drivers.helpers import append_to_file
+from drivers.asserts import assert_substring
+
+init_local_crate(with_test=False)
+
+# Add a conditional test
+append_to_file(alr_manifest(),
+               [
+                   "[[test]]",
+                   "[test.'case(os)'.'...']",
```
AldanTanneo: From how I understand the TOML format, shouldn't a dynamic property just have the `[test.'case(os)'...]` line, without the `[[test]]`?

src/alire/alire-properties-from_toml.ads
mosteo: Yes, and surely we shouldn't have any incompatible release in the index at this point

testsuite/tests/test/conditional-tests/test.py
mosteo: Both should work. I'll add it to the test.

mosteo: The reason I only used the array form is because I remember when I did dynamic actions that was the problematic one, so I instinctively went directly for it.

mosteo:
Addressed review by @AldanTanneo

Fabien-Chouteau (approved):

PR #1890, opened 2025-03-11
feat: `templates_parser` for file initializations

### LINKED ISSUE #11: Use Templates_Parser for cleaner file generation

### LINKED ISSUE #1896: Replace example test with a more useful one
Replace the default example test with the `Assertions_Enabled` in https://github.com/alire-project/alire/pull/1893

mosteo:
The idea is to use `Templates_Parser` to generate files (currently only during `alr init`), which enhances maintenance, and opens the door to user templates (not in this PR, this only deals with our own hardcoded initializations).

This is achieved this way:

- Files under `templates` replicate the trees we want to generate. These can contain TP variables/filters/etc.
- A new `support/embedder/embedder.sh` script generates embedded resources under `src/templates`. It relies on `awsres` from AWS, but since this is done independently of Alire's build, it doesn't introduce new dependencies (in particular I didn't want to depend on AWS from `alr`).
  - There's a hash-based check to avoid regenerating resources, so normal builds of Alire shouldn't need AWS nor regenerate anything. Contributors can continue building `alr` with the same simplicity.
- `Alire.Templates` provides a couple of wrappers over TP to use the embedded resources.
- `Alire.Templates.Builtins` constants define the structure of folders we want to generate.

User templates could be had  e.g. via `alr init --template=<url>` with little extra effort now that the basis is in place.

This makes the only new dependency to be `templates_parser`, and it doesn't require special compilation as I forked their repo to have the preconfiguration done, so `gprbuild` is still enough to build.

Edit: I initially thought xmlada was also required but it can be disabled in templates_parser.

Fixes #11.
Fixes #1896.


.gitmodules
```diff
@@ -69,3 +69,6 @@
 [submodule "deps/yeison"]
 	path = deps/yeison
 	url = https://github.com/mosteo/yeison
+[submodule "deps/templates-parser"]
+	path = deps/templates-parser
+	url = https://github.com/alire-project/templates-parser
```
Fabien-Chouteau, line 74: Why not use the official repo?

Fabien-Chouteau (commented):
I am not sure about the use of ARE for this use case. We have another way to handle resources for native applications: https://github.com/alire-project/resources

mosteo:
> I am not sure about the use of ARE for this use case. We have another way to handle resources for native applications: https://github.com/alire-project/resources

This is to make `alr` self-contained, as we talked about. It's true that we could clone resources like we clone the community index, but that seems a larger departure from the current situation I don't feel comfortable moving away (yet?)

Or if for example we started using templates for the long help descriptions, it would also be nice not to depend on external files.

.gitmodules
mosteo: Because the official repo is missing a project file which is generated with `make`, and I don't want to introduce that dependency in the build of `alr` from scratch without `alr`... I like depending only on `gprbuild` for as long as possible.

An alternative to templates parser could be [JinTP](https://github.com/rehartmann/JinTP.git) which I just became aware of the other day. It looks more like the liquid templates in jekyll, and seems to be quite light. I don't know if you have experience with it, I don't know how mature it is in real use.

Fabien-Chouteau (approved):

PR #1894, opened 2025-03-12
feat: add .gitignore to `alr init` generated test crate

AldanTanneo:


mosteo:
Is this really necessary? Isn't the `.gitignore` of the main crate applying to the test crate too?

mosteo:
I see that the problem is that our .gitignore is not very well defined and so lots of things aren't properly ignored in subcrates.

AldanTanneo:
Yeah, it only ignores specific folders at the root of the repository, and not in subcrate. I added this as a nested .gitignore, do you think it would be better to modify the existing .gitignore?

mosteo:
No, I think it's okay to be targeted rather than risking ignoring something not created by us.

mosteo (approved):

PR #1902, opened 2025-03-15
fix: enable assertions in test crate

mosteo:
Plain `pragma Assert` weren't enabled by default.

##### PR creation checklist

### DIFF
--- a/src/alr/alr-commands-init.adb
+++ b/src/alr/alr-commands-init.adb
@@ -370,6 +370,7 @@ package body Alr.Commands.Init is
          Put_New_Line;
          Put_Line ("[build-profiles]");
          Put_Line (Lower_Name & " = 'validation'");
+         Put_Line (Test_Lower & " = 'validation'");
          TIO.Close (File);
 
          if not Create (+Full_Name (Test_Directory / (+Test_Lower & ".gpr"))) then
--- a/testsuite/tests/test/default-failure/test.py
+++ b/testsuite/tests/test/default-failure/test.py
@@ -9,10 +9,24 @@ from drivers.asserts import assert_match
 
 init_local_crate("xxx", with_test=True)
 
+# Check with explicit exception
+
 with open("./tests/src/xxx_tests-example_test.adb", "w") as f:
    f.write("""procedure Xxx_Tests.Example_Test is
 begin
    raise Program_Error;
+end Xxx_Tests.Example_Test;
+""")
+
+p = run_alr("test", complain_on_error=False)
+assert_match(".*\[ FAIL \] example_test.*", p.out)
+
+# Check with plain assertion (verify that assertions are evaluated)
+
+with open("./tests/src/xxx_tests-example_test.adb", "w") as f:
+   f.write("""procedure Xxx_Tests.Example_Test is
+begin
+   pragma Assert (False);
 end Xxx_Tests.Example_Test;
 """)
 


Fabien-Chouteau:
@mosteo This is something @AldanTanneo and I discussed before. The use of a `validation` profile is not very user-friendly because of all the noise it brings. That's why the example test uses the `Ada.Assertions.Assert` function (which will always check) instead of the pragma.

On a side note, now that GNATformat is available, we should probably remove the default stylecheck switches from all build profiles.

mosteo:
Let's see if I understand: the idea is that the main crate is in validation mode, but the test crate itself is not?

Is there some documentation where I can get up to speed with gnatformat and intended uses?

AldanTanneo:
The main issue with validation mode is that it enforces very strict style checks, and those are not desirable during the development process (during which you will probably run the tests over and over)

IMO the `'development'` profile should enable all assertions and no style checks, and be used for the test crate.

mosteo:
The mixing of assertions and style checks is certainly unfortunate. To me, validation means the strictest run-time correctness, regardless of style.

I remember @Fabien-Chouteau discouraging assertions enabled in `development` profile as they are very heavy in some crates (I think proof-heavy crates, it's not been a problem in my particular use). If not for that, I too presumed assertions should be enabled during `development`.

My current understanding of these profiles as they are today in Alire is:
- `development`: amenable to `gdb`-style debugging, not particularly safe (as assertions are disabled).
- `validation`: full correctness checks (with the unfortunate side effect of full style checks).

I personally develop with all my crates in validation mode because of assertions, and don't see much use in development profile at all. Both because lack of assertions and because I very rarely use a debugger, so the `-O3` in validation doesn't affect me. 

So, I'm a bit confused because @AldanTanneo's test runner original PR (#1881) does enable validation for the main crate, and I'm not sure I understand what is gained by not using it also in the test suite, so I'm presuming validation for the main crate (as it is defined now) is not wanted either.

From what has been said, it seems to me that what is wanted (by all of us) is a mode with full assertions and correctness checks but without style checks. Short of burdening us immediately with a new mode, that should be validation mode, I think.

(Side note:  I find highly counterintuitive that `pragma Assert` doesn't work by default in the test suite. It's sure to bite someone (as it did to me)).

In conclusion: drop all style checks (perhaps have a new profile `--style` at some point?) and keep crate and tests at validation? Or I'm off-the-mark here?

(Another side note: if we are going to redefine profiles for 3.0, I would vote for enabling assertions by default in `development`. My guess is that there's less people affected by heaviness of assertions (who could disable them in that case) than people surprised by assertions not working in `development` profile. But I may have a biased perception that's plainly wrong in industry code bases.)

AldanTanneo:
I agree that ideally assertions should be enabled both in validation and development. I find the split between development and validation a bit weird (coming from the Rust world, which does things differently - assertions are always checked, but the compiler is very good at eliminating them, and uses them as optimization hints; and there are specific "debug assertions" that are only enabled during development).

But yes, the testing crate should definitely enable all assertions, and build the tested crate with assertions too. I don't know what the best path is for this. Maybe development can be assertions, validation can become assertions + style, and release no assertions? It did surprise me at first that the development profile does not enable assertions, it's quite counterintuitive for newcomers.

mosteo:
> Maybe development can be assertions, validation can become assertions + style, and release no assertions?

This could work for me too. Let's hear from @Fabien-Chouteau, he designed the three profiles we have now.

Fabien-Chouteau:
I agree that enabling all checks would be better for everyone in theory. However, I got bitten by bad performances (like an unusable level of bad), and that's why I didn't enable them in the first version of the build profiles. 

I am ok to change this and enable all checks in `development` but I would like to get more opinions on this.

For the style checks, I've received a lot of negative feedback on the fact that we try to enforce them. In addition to the availability of GNATformat, I think they should be eliminated in all profiles.

PR #1906, opened 2025-03-18
fix: updating pins with `subdir` which track Alire-generated files

Seb-MCaw:
Updates #1788 in line with #1857.

##### PR creation checklist


Seb-MCaw:
Both failures look transient to me, so ready for review @mosteo.

mosteo (approved):

PR #1909, opened 2025-03-19
feat: add test runner selection via `--id`

### LINKED ISSUE #1899: Add optional 'label' to [test]
To allow running only one particular test runner with argument forwarding

AldanTanneo:
Closes #1899

I picked the name `id` as it made more sense to me personally, but it's still possible to change it.

##### PR creation checklist
- [x] `doc/catalog-format-spec.md` has been updated, if applicable.

### DIFF
--- a/doc/catalog-format-spec.md
+++ b/doc/catalog-format-spec.md
@@ -449,6 +449,8 @@ static, i.e. they cannot depend on the context.
     # OR
     command = ["<command>", "<argument>", ...]
 
+    # optional keys
+    id = <string>
     directory = <relative path>
     jobs = <number>
     ```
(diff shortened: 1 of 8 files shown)


mosteo:
> I picked the name `id` as it made more sense to me personally, but it's still possible to change it.

Which other possibilities did you consider? I disliked `name` for being too easy to confuse with a "test name" in the sense of an individual test.

The only other possibility I could think of was "label", which has the advantage of being an actual word.

mosteo (approved):

AldanTanneo:
> The only other possibility I could think of was "label", which has the advantage of being an actual word.

I considered that but there's just something about writing `alr test --label XXX` that doesn't sit right with me. And the field is used to <b>id</b>entify the test, not just label it.

To me 'label' would be something used for display purposes. Does that makes sense to you?

mosteo:
OK, I can see how that can imply the action of labeling. If we don't have any better alternatives, so be it.

PR #1912, opened 2025-03-19
feat: `--builtin` safeguard for `alr settings`

mosteo:
The idea here is that when you want to set a built-in setting, you don't mistakenly set something else because of a typo, or misremembering the key. It works kind of like `overriding` in Ada:

-  `alr settings --set --builtin` will fail if the setting is not a builtin
- `alr settings --set` will warn if setting a builtin without `--builtin`

Works the same for `--get` and `--unset`.

##### PR creation checklist
- [x] `BREAKING.md` has been updated for major changes in `alr`, minor/major in catalog format.


doc/user-changes.md
```diff
@@ -752,394 +765,4 @@ dependencies in a solution, as the solver will discard these combinations.
 
 ### Toolchain management
 
-PR [#775](https://github.com/alire-project/alire/pull/775)
```
Fabien-Chouteau: Why remove all this?

mosteo: By mistake :facepalm:

Fabien-Chouteau (approved):

PR #1914, opened 2025-03-19
feat: find tests in nested folders

### LINKED ISSUE #1895: Allow having tests in nested folders inside `src`
To help with classification of large numbers of tests, as we do for the e3 tests.

The `den` crate, already a dependency, is appropriate for this.

mosteo:
Allow classification of tests in subfolders:
```
[ PASS ] nested/sub/another_test
[ PASS ] nested/example2_test
[ PASS ] example_test
```

Closes #1895

##### PR creation checklist

### DIFF
--- a/src/alire/alire-test_runner.adb
+++ b/src/alire/alire-test_runner.adb
@@ -1,5 +1,5 @@
 with Ada.Containers.Indefinite_Ordered_Maps;
-with Ada.Strings.Fixed;
+with Ada.Containers.Indefinite_Vectors;
 with Ada.Text_IO;
 with GNAT.OS_Lib;
 with System.Multiprocessors;
@@ -101,24 +101,54 @@ package body Alire.Test_Runner is
       end if;
    end Strip_Prefix;
 
-   subtype Portable_Path_Vector is AAA.Strings.Vector;
-   --  Crashes on older GNATs, but for reference:
-   --  with Predicate => (for all Path of Portable_Path_Vector =>
-   --                       VFS.Is_Portable (Path));
+   ------------------
+   -- Strip_Suffix --
+   ------------------
+
+   function Strip_Suffix (Src, Suffix : String) return String is
+   begin
+      if AAA.Strings.Has_Suffix (Src, Suffix) then
+         return Src (Src'First .. Src'Last - Suffix'Length);
+      else
+         return Src;
+      end if;
+   end Strip_Suffix;
+
+   ------------------
+   -- Display_Name --
+   ------------------
+
+   function Display_Name
+     (Name : Portable_Path; Root_Prefix : String) return String
+   is
+      Simple      : constant String :=
+        Strip_Suffix
+          (Strip_Prefix (VFS.Simple_Name (Name), Root_Prefix), ".adb");
+      Parent_Name : constant Portable_Path := VFS.Parent (Name);
+   begin
+      if Parent_Name = "." then
+         return Simple;
+      else
+         return String (Parent_Name) & "/" & Simple;
+      end if;
+   end Display_Name;
+
+   package Portable_Path_Vectors is new
+     Ada.Containers.Indefinite_Vectors (Positive, Portable_Path);
+   subtype Portable_Path_Vector is Portable_Path_Vectors.Vector;
 
    ---------------------
    -- Create_Gpr_List --
    ---------------------
 
-   procedure Create_Gpr_List
-     (Root : Alire.Roots.Root;
-      List : Portable_Path_Vector)
-      --  Create a gpr file containing a list of the test files
-      --  (named `Test_Files`).
+   procedure Create_Gpr_List (Root : Roots.Root; List : Portable_Path_Vector)
+     --  Create a gpr file containing a list of the test files
+     --  (named `Test_Files`).
 
    is
-      File_Path : constant Alire.Absolute_Path :=
-        Root.Path / Alire.Paths.Default_Config_Folder
+      File_Path : constant Absolute_Path :=
+        Root.Path
+        / Paths.Default_Config_Folder
         / (Root.Name.As_String & "_list_config.gpr");
       File      : Text_Files.File := Text_Files.Create (File_Path);
       Lines     : access AAA.Strings.Vector renames File.Lines;
@@ -142,8 +172,7 @@ package body Alire.Test_Runner is
          else
             Lines.Append_To_Last_Line (",");
          end if;
-         Lines.Append_To_Last_Line
-           ("""" & VFS.Simple_Name (Portable_Path (Name)) & ".adb""");
+         Lines.Append_To_Last_Line ("""" & VFS.Simple_Name (Name) & """");
       end loop;
 
       Lines.Append_Line (Indent & ");");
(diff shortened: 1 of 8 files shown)


mosteo:
@AldanTanneo could you please take a look?

src/alire/alire-test_runner.adb
```diff
@@ -254,23 +274,33 @@ package body Alire.Test_Runner is
...
       begin
          if Name'Length > 4
            and then Name (Name'Last - 3 .. Name'Last) = ".adb"
            and then (Filter.Is_Empty
                      or else (for some F of Filter
                               => Ada.Strings.Fixed.Index (Name, F) /= 0))
          then
-            Test_List.Append (Name (Name'First .. Name'Last - 4));
+            Test_List.Append (Name (Name'First + 1 .. Name'Last - 4));
+            --  +1 removes initial '/'
```
AldanTanneo: Why have a +1 here and not use `"src/"` in the call to `Strip_Prefix`?

Edit: Hmm there's the issue of windows file separators. Maybe it would be a lesser evil to do a `Replace ('\', '/')`, as Windows file APIs are mostly compatible with UNIX style path separators anyway

src/alire/alire-test_runner.adb
```diff
@@ -254,23 +274,33 @@ package body Alire.Test_Runner is
...
-         Name : constant String := Adirs.Simple_Name (Dir_Entry);
+         Name : constant String :=
+                  Strip_Prefix (This.Path,
+                                Prefix => Root.Path / "src");
       begin
          if Name'Length > 4
            and then Name (Name'Last - 3 .. Name'Last) = ".adb"
            and then (Filter.Is_Empty
                      or else (for some F of Filter
                               => Ada.Strings.Fixed.Index (Name, F) /= 0))
```
AldanTanneo: Something I missed during the filtering initially: this also matches on the package name (`crate_test-...`) which is hidden during output :grimacing: I don't know if we should do something about that

src/alire/alire-test_runner.adb
mosteo: That was to not do an extra `& Dir_Separator` elsewhere.

I don't think Windows should cause problems here? Or it would have triggered already something in the testsuite.

AldanTanneo: Well, the testsuite did fail on windows because of path separators 😛 (not here though)

src/alire/alire-test_runner.adb
mosteo: I noticed too. Probably yes just in case for some cases it brings in extra unwanted tests.

src/alire/alire-test_runner.adb
mosteo: Haha I saw it immediately after writing.

In the end I've changed it to always use forward slashes as it seems desirable that test output is consistent across OSes.

src/alire/alire-test_runner.adb
```diff
@@ -254,11 +275,17 @@ package body Alire.Test_Runner is
...
+                        Unused_Stop  : in out Boolean)
+      is
+         --  Helper function to append all .adb files in a tree
          --  to the `Test_List` vector
 
-         Name : constant String := Adirs.Simple_Name (Dir_Entry);
+         Name : constant String :=
+                  Strip_Prefix
+                    (This.Path,
+                     Prefix => (Root.Path / "src") & OS_Lib.Dir_Separator);
```
AldanTanneo: We should convert to portable path form here, so filtering is consistent with the output

AldanTanneo:
Also: I formatted `alire-test_runner.adb` using GNATformat all along, I'd appreciate it if it could keep being formatted automatically

src/alire/alire-test_runner.adb
AldanTanneo: I'd have something like

```ada
function Matches_Filter (Name : String) return Boolean;
```

that strips the package name part and does the matching on the test's display name, and call that with

```ada
Name : constant String :=
   Strip_Prefix
     (OS_Lib.To_Portable (This.Path),
      Prefix => Root.Path / "src/"));
```

mosteo: I didn't want to touch the GPR generation, but in the end I went all in and fixed filtering too.

mosteo:
> Also: I formatted `alire-test_runner.adb` using GNATformat all along, I'd appreciate it if it could keep being formatted automatically

How are you doing this?

src/alire/alire-test_runner.adb
mosteo: I did something similar, not sure you have seen it yet. Not refactored to its own function though.

mosteo:
> I'd appreciate it if it could keep being formatted automatically

I understand the sentiment and for this time I've manually run `gnatformat`, but until we mandate it for all our sources you may have to live with a degree of OCD, sorry about that :/

src/alire/alire-test_runner.adb
AldanTanneo: My issue here is that using "nested/test" to filter will not work on windows, as the path is not made portable during filtering

mosteo: Should be fixed now and caught by the new test

src/alire/alire-test_runner.adb
```diff
@@ -294,11 +299,13 @@ package body Alire.Test_Runner is
...
+         Name : constant OS_Lib.Portable_Path_Like :=
+           OS_Lib.To_Portable
+             (Strip_Prefix
+                (This.Path,
+                 Prefix => (Root.Path / "src") & OS_Lib.Dir_Separator));
 
-         Filtering_Name : constant String :=
+         Filtering_Name : constant OS_Lib.Portable_Path_Like :=
            AAA.Strings.Replace
              (Text => Name, Match => Root_Prefix (Root), Subst => "");
```
AldanTanneo: This breaks if we have a structure like

```
src
|_ crate_tests-for_fun
   |_ crate_tests-example.adb
```

Which admittedly is a bit contrived but still.

AldanTanneo: In pseudo Ada:

```ada
function Display_Name (Name : Portable_Path; Root_Prefix : String)
return Portable_Path_Like is
   Simple : String := Strip_Prefix (Simple_Name (Name), Root_Prefix);
   Parent_Name : String := Portable_Path (Parent (Name));
begin
   if Parent_Name = "." then
      return Simple;
   else
      return Parent_Name & "/" & Simple;
   end if;
end Display_Name;
```

And you could reuse the function in the spawn code.

AldanTanneo: https://github.com/AldanTanneo/alire/commit/4d682b19f3cd486ef97d06e8979723b3e976d7e5

My commit that fixes this (and some other structural things)

mosteo: Cool, thanks.

mosteo: > Which admittedly is a bit contrived but still.

Ugh, right.

mosteo:
@AldanTanneo I pulled your [452c9e7](https://github.com/alire-project/alire/pull/1914/commits/452c9e7a4b0f01c0c4d38cb4052f4e0882015f09) so it should be good to go once tests pass.

AldanTanneo (approved):

PR #1924, opened 2025-03-27
dev: override any existing `awsres` in `embedder.sh`

Seb-MCaw:
This would be more convenient for those of us with a more recent (and therefore incompatible) version of `awsres` already on our `PATH`, but I can work around it if you'd prefer to leave as-is.


Seb-MCaw:
Ready for review @mosteo.

mosteo (approved):

PR #1925, opened 2025-03-27
fix: use copy/delete for renames by default

mosteo:
Some temporary renamings are failing seemingly because the temporary location is in a different filesystem. Copy/delete is a safer alternative by default. When we are positive that temporary and destination are in the same filesystem, we still can do cheap renames (e.g. pins, releases in the vault, releases under the workspace).

This PR adds an `Alire.Directories.Rename` that can do both regular renaming and copy/delete. Uses of `Ada.Directories.Rename` are replaced with the Alire one.

Also, for attribute changing on Windows, we use our own recursion now as Window's `attrib` with `/R` fails for certain looping softlinks (as evidenced in the `den` crate test cases).

### DIFF
--- a/src/alire/alire-directories.adb
+++ b/src/alire/alire-directories.adb
@@ -224,29 +224,50 @@ package body Alire.Directories is
    ----------------------
 
    procedure Ensure_Deletable (Path : Any_Path) is
-      use Ada.Directories;
+
+      ---------------------------
+      -- Ensure_Deletable_Item --
+      ---------------------------
+
+      procedure Ensure_Deletable_Item
+        (Path : Any_Path; Unused : in out Boolean)
+      is
+         use all type Den.Kinds;
+      begin
+         case Den.Kind (Path) is
+            when Nothing =>
+               raise Program_Error
+                 with "cannot change attributes of non-existing file: " & Path;
+            when Directory =>
+               Trace.Debug ("Forcing writability of dir " & Path);
+               OS_Lib.Subprocess.Checked_Spawn
+                 ("attrib",
+                  AAA.Strings.Empty_Vector
+                  .Append ("-R") -- Remove read-only
+                  .Append ("/D") -- On dirs
+                  .Append (Path & "\*"));
+            when File | Softlink | Special =>
+               Trace.Debug ("Forcing writability of file " & Path);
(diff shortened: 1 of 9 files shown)


src/alire/alire-directories.adb
```diff
@@ -369,6 +364,41 @@ package body Alire.Directories is
...
+   procedure Rename (Source,
+                     Destination : Any_Path;
+                     Copy_Delete : Boolean := True) is
+   begin
+      Trace.Debug ("Renaming " & Source & " (" & Kind (Source)'Image & ") "
+                   & "into " & Destination
+                   & " using copy/delete=" & Copy_Delete'Image);
+
+      if Exists (Destination) then
+         raise Program_Error with
```
Fabien-Chouteau: Don't we have `Raise_Checked_Error ()` for this?

src/alire/alire-directories.adb
```diff
@@ -369,6 +364,41 @@ package body Alire.Directories is
          raise;
    end Force_Delete;
 
+   ------------
+   -- Rename --
+   ------------
+
+   procedure Rename (Source,
+                     Destination : Any_Path;
+                     Copy_Delete : Boolean := True) is
```
Fabien-Chouteau: Why not detect automatically here if copy/delete is required?

Fabien-Chouteau (approved):

src/alire/alire-directories.adb
mosteo: Yes, in this case it would be probably more adequate, thanks for the catch.

src/alire/alire-directories.adb
mosteo: Frankly, it didn't cross my mind that it could be detected easily in an OS-portable way. I will take a look.

Worst case, I guess we could check whether source and destination are in the same folder.

Fabien-Chouteau: If the issue is location in different filesystems, that should be easy to do with absolute paths on Windows. Or am I missing something?

mosteo: I was thinking more of Linux where mountpoints can be anywhere (and I think this is possible nowadays on Windows too). In the end, I'm thinking of simply trying move first, and if it fails fall back to copy/delete.

PR #1935, opened 2025-04-14
POC: test case detection with headers

Fabien-Chouteau:
This is just a proof of concept to play with a detection of test cases based on a header in the .adb source file of the test for #1881. 

This patch implements both pragmas and TOML front-matter headers:
```ada
pragma Alire_Test (Name, "This is a test");
pragma Alire_Test (Should_Fail, True);
```
```ada
--  +++
--  name = "This is a test"
--  should-fail = true
--  +++
```

Both solutions are relatively similar in code size, **but** the pragma version implemented here is quite incomplete as we would need a proper parser for Ada strings and integer literals (maybe real literals at some point...).

### DIFF
--- a/src/alire/alire-test_runner.adb
+++ b/src/alire/alire-test_runner.adb
@@ -1,5 +1,8 @@
+pragma Ada_2022;
+
+with AAA.Enum_Tools;
 with Ada.Containers.Indefinite_Ordered_Maps;
-with Ada.Containers.Indefinite_Vectors;
+with Ada.Containers.Vectors;
 with Ada.Strings.Fixed;
 with Ada.Text_IO;
 with GNAT.OS_Lib;
@@ -11,6 +14,8 @@ with Alire.Paths;
 with Alire.Utils.Text_Files;
 with Alire.VFS;
 
+with TOML;
+
 with CLIC.TTY;
 
 with Den.Walk;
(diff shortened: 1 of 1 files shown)


mosteo:
Thanks, Fabien. I will take a look in depth ASAP.

mosteo:
> Both solutions are relatively similar in code size, **but** the pragma version implemented here is quite incomplete as we would need a proper parser for Ada strings and integer literals (maybe real literals at some point...).

Thanks for the POC, Fabien, and I apologize about the long silence. I see it's quite straightforward.

Ultimately I'm unsure about which way to go. Well, if this were only me, I know i would go with pragmas no matter the extra complexity, "because Ada" and because this is the kind of fun new rabbit hole I like. But it's not only me, and I see the advantages of a simpler, fully general solution from the start without hidden risks.

Maybe we could ask the community on this? Perhaps more eyes can come with more use cases that tilt the scales.

Fabien-Chouteau:
Renewing my request for feedback on this :)

mosteo:
Yes! Sorry. Restacking closer to the top.

mosteo:
Hi Fabien, sorry for the suuuper late reply. I did look at the implementations in several occasions, even if I never reached back... 

I don't have anything substantive to say, besides that I'd prefer to go the pragma way on this one. In that regard, I would like to integrate the pragma processing in [lml](https://github.com/mosteo/lml_ada), so to have a (limited) but transparent conversion from pragmas to a standard JSON-like object. I can work it on top of this POC.

PR #1963, opened 2025-06-02
feat: alr self-update command

AldanTanneo:
##### PR creation checklist

### DIFF
--- a/alire.toml
+++ b/alire.toml
@@ -27,15 +27,15 @@ gnatcoll = "^21"
 lml = "~0.1"
 minirest = "~0.3"
 optional = "~0.1.1"
+resources = "~0.1"
 semantic_versioning = "^3.0"
 simple_logging = "^2.0"
 si_units = "~0.2.2"
+spdx = "~0.2"
 stopwatch = "~0.1"
 templates_parser = "^24" # Latests pre-Ada 2022 version
 toml_slicer = "~0.1"
 uri_ada = "^2.0"
-spdx = "~0.2"
-resources = "~0.1"
 
 [gpr-set-externals]
 CLIC_LIBRARY_TYPE="static"  # Has problems with "static-pic" for some reason
@@ -102,6 +102,10 @@ commit = "9a9c660f9c6f27f5ef75417e7fac7061dff14d78"
 url    = "https://github.com/mosteo/optional.git"
 commit = "7b8132a09a6c3c467409ab15d34fac605b1e5711"
 
+[pins.resources]
+url    = "https://github.com/alire-project/resources"
+commit = "d3a83f35d7c0b586119fc19a568289cf21c87aa0"
+
 [pins.semantic_versioning]
 url    = "https://github.com/alire-project/semantic_versioning"
 commit = "4861e32bd8a2f0df038d3ecc9a72b6381e7a34cc"
@@ -131,10 +135,6 @@ commit = "3e5cbdb5673b85a1da6344a41764ef1cbafe3289"
 url    = "https://github.com/mosteo/yeison"
 subdir = "yeison_12"
 commit = "20033cf90a3eb2eff69d9ff9d0820291aed79b57"
-
-[pins.resources]
-url    = "https://github.com/alire-project/resources"
-commit = "d3a83f35d7c0b586119fc19a568289cf21c87aa0"
 
 # To disable version updating, export ALR_VERSION_DONT_PATCH with any value
 
--- a/alire.toml
+++ b/alire.toml
@@ -35,6 +35,7 @@ templates_parser = "^24" # Latests pre-Ada 2022 version
 toml_slicer = "~0.1"
 uri_ada = "^2.0"
 spdx = "~0.2"
+resources = "~0.1"
 
 [gpr-set-externals]
 CLIC_LIBRARY_TYPE="static"  # Has problems with "static-pic" for some reason
@@ -130,6 +131,10 @@ commit = "3e5cbdb5673b85a1da6344a41764ef1cbafe3289"
 url    = "https://github.com/mosteo/yeison"
 subdir = "yeison_12"
 commit = "20033cf90a3eb2eff69d9ff9d0820291aed79b57"
+
+[pins.resources]
+url    = "https://github.com/alire-project/resources"
+commit = "d3a83f35d7c0b586119fc19a568289cf21c87aa0"
 
 # To disable version updating, export ALR_VERSION_DONT_PATCH with any value
 
--- a/src/alire/alire-github.adb
+++ b/src/alire/alire-github.adb
@@ -23,6 +23,8 @@ package body Alire.GitHub is
 
    Repos       : constant String := "repos";
    Pulls       : constant String := "pulls";
+   Releases    : constant String := "releases";
+   Latest      : constant String := "latest";
 
    -------------------
    -- Community_API --
@@ -32,6 +34,13 @@ package body Alire.GitHub is
    is (Repos
        / Index.Community_Organization
        / Index.Community_Repo_Name);
+
+   --------------------
+   -- Alire_Repo_API --
+   --------------------
+
+   function Alire_Repo_API return String
+   is (Repos / Index.Community_Organization / "alire");
 
    -----------------
    -- JSON_Escape --
@@ -451,5 +460,13 @@ package body Alire.GitHub is
       --  TODO: do we need to additionally request a review, or simply by
       --  removing the draft status we'll get a notification?
    end Request_Review;
+
+   function Get_Latest_Alire_Release return String is
+      Response : constant GNATCOLL.JSON.JSON_Value :=
+        API_Call (Alire_Repo_API / Releases / Latest);
+      --  public endpoint
+   begin
+      return Response.Get ("tag_name");
+   end Get_Latest_Alire_Release;
 
 end Alire.GitHub;
--- a/src/alire/alire-github.adb
+++ b/src/alire/alire-github.adb
@@ -473,4 +473,38 @@ package body Alire.GitHub is
       return Response.Get ("tag_name");
    end Get_Latest_Alire_Release;
 
+   function Check_Alire_Binary_Release (Tag, Archive : String) return Outcome
+   is
+      Response : constant Minirest.Response :=
+        API_Call (Alire_Repo_API / Releases / "tags" / Tag);
+   begin
+      if Response.Succeeded then
+         declare
+            use GNATCOLL.JSON;
+
+            Data   : constant JSON_Value :=
+              GNATCOLL.JSON.Read (Response.Content.Flatten (""));
+            Assets : constant JSON_Array := Data.Get ("assets");
+            Len    : constant Natural := Length (Assets);
+         begin
+            if (for some I in 1 .. Len
+                => Get (Assets, I).Get ("name") = Archive)
+            then
+               return Outcome_Success;
+            else
+               return
+                 Outcome_Failure
+                   ("could not find artifact '"
+                    & Archive
+                    & "' in release "
+                    & Tag,
+                    Report => False);
+            end if;
+         end;
+      else
+         return
+           Outcome_Failure ("could not find release " & Tag, Report => False);
+      end if;
+   end Check_Alire_Binary_Release;
+
 end Alire.GitHub;
--- a/src/alire/alire-github.adb
+++ b/src/alire/alire-github.adb
@@ -473,6 +473,10 @@ package body Alire.GitHub is
       return Response.Get ("tag_name");
    end Get_Latest_Alire_Release;
 
+   --------------------------------
+   -- Check_Alire_Binary_Release --
+   --------------------------------
+
    function Check_Alire_Binary_Release (Tag, Archive : String) return Outcome
    is
       Response : constant Minirest.Response :=
--- a/src/alire/alire-github.adb
+++ b/src/alire/alire-github.adb
@@ -461,6 +461,10 @@ package body Alire.GitHub is
       --  removing the draft status we'll get a notification?
    end Request_Review;
 
+   ------------------------------
+   -- Get_Latest_Alire_Release --
+   ------------------------------
+
    function Get_Latest_Alire_Release return String is
       Response : constant GNATCOLL.JSON.JSON_Value :=
         API_Call (Alire_Repo_API / Releases / Latest);
--- a/src/alr/alr-commands-self_update.adb
+++ b/src/alr/alr-commands-self_update.adb
@@ -309,6 +309,7 @@ package body Alr.Commands.Self_Update is
 
          if Proceed = UI.Yes then
             Install_Alr (Dest_Base, Tmp_Dir, Extract_Dir / "bin" / Alr_Bin);
+            Alire.Put_Success ("updated alr [" & Tag_String (T) & "]");
          end if;
       end;
    exception
--- a/src/alr/alr-commands-self_update.adb
+++ b/src/alr/alr-commands-self_update.adb
@@ -1,3 +1,5 @@
+with Ada.Exceptions;
+
 with Alire.GitHub;
 with Alire.OS_Lib.Download;
 with Alire.OS_Lib.Subprocess;
@@ -170,11 +172,9 @@ package body Alr.Commands.Self_Update is
       use Alire.OS_Lib.Operators;
 
       Dest_Bin   : constant Any_Path := Dest_Base / Alr_Bin;
-      Backup_Bin : constant Any_Path := Dest_Base / ("alr.old" & Exe);
+      Backup_Bin : constant Any_Path :=
+        Dest_Base / (Dirs.Temp_Name (Length => 16) & Exe);
    begin
-      if Dirs.Is_File (Backup_Bin) then
-         Dirs.Adirs.Delete_File (Backup_Bin);
-      end if;
       if Dirs.Is_File (Dest_Bin) then
          Dirs.Adirs.Rename (Dest_Bin, Backup_Bin);
       end if;
@@ -183,10 +183,13 @@ package body Alr.Commands.Self_Update is
          Dirs.Adirs.Copy_File (Extract_Bin, Dest_Bin);
       exception
          --  if operation failed and a backup was made, restore previous
-         when others =>
+         when E : others =>
             if Dirs.Is_File (Backup_Bin) then
                Dirs.Adirs.Rename (Backup_Bin, Dest_Bin);
             end if;
+            Reportaise_Command_Failed
+              ("could not copy extracted file to destination: "
+               & Ada.Exceptions.Exception_Message (E));
       end;
 
       case Plat.Current.Operating_System is
--- a/src/alr/alr-commands-self_update.adb
+++ b/src/alr/alr-commands-self_update.adb
@@ -1,4 +1,5 @@
 with Ada.Exceptions;
+with Ada.Text_IO;
 
 with Alire_Early_Elaboration;
 with Alire.GitHub;
@@ -52,16 +53,17 @@ package body Alr.Commands.Self_Update is
    ---------------------
 
    function Get_Version_Tag (Cmd : Command) return Tag is
+      First_Self_Updatable_Alr : constant Semver.Version :=
+        Semver.New_Version (3, 0, 0);
    begin
       if Cmd.Nightly then
          return (Kind => Nightly);
-      elsif Cmd.Force_Version /= null and then Cmd.Force_Version.all /= "" then
+      elsif Cmd.Release /= null and then Cmd.Release.all /= "" then
          declare
-            V     : constant Semver.Version :=
-              Semver.Parse (Cmd.Force_Version.all);
+            V     : constant Semver.Version := Semver.Parse (Cmd.Release.all);
             V_Img : constant String := Semver.Image (V);
          begin
-            if V_Img /= Cmd.Force_Version.all then
+            if V_Img /= Cmd.Release.all then
                Trace.Info ("version string parsed as '" & V_Img & "'");
             end if;
             if V < Alire.Version.Current then
@@ -71,8 +73,10 @@ package body Alr.Commands.Self_Update is
                   & " (current: "
                   & Semver.Image (Alire.Version.Current)
                   & ")");
-               Trace.Warning
-                 ("previous versions may not have the `self-update` command");
+               if V < First_Self_Updatable_Alr then
+                  Trace.Warning
+                    ("this version will not have the `self-update` command");
+               end if;
             end if;
             return (Specific_Version, V);
          end;
@@ -83,15 +87,15 @@ package body Alr.Commands.Self_Update is
             V        : constant Semver.Version :=
               Semver.Parse (Tag_Name (Tag_Name'First + 1 .. Tag_Name'Last));
          begin
-            if Alire.Version.Current.Pre_Release = "dev" then
+            if Alire.Version.Current.Pre_Release /= "" then
                Trace.Info
-                 ("Detected nightly version. Use --force="
+                 ("Detected nightly version. Use --release="
                   & Semver.Image (V)
                   & " to update to the latest stable release.");
                return (Kind => Nightly);
             elsif V < Alire.Version.Current then
                Trace.Warning
-                 ("you are currently on a pre-release (v"
+                 ("you are currently on a preview version (v"
                   & Semver.Image (Alire.Version.Current)
                   & ")");
                Trace.Warning
@@ -100,7 +104,7 @@ package body Alr.Commands.Self_Update is
             elsif V = Alire.Version.Current then
                Trace.Info ("You are already using the latest version of alr!");
                Trace.Info
-                 ("To reinstall the current version, use --force="
+                 ("To reinstall the current version, use --release="
                   & Semver.Image (V));
                raise Abort_With_Success;
             end if;
@@ -110,7 +114,7 @@ package body Alr.Commands.Self_Update is
    exception
       when Semver.Malformed_Input =>
          Reportaise_Command_Failed
-           ("specified invalid alr version: " & Cmd.Force_Version.all);
+           ("specified invalid alr version: " & Cmd.Release.all);
    end Get_Version_Tag;
 
    ----------------
@@ -178,8 +182,7 @@ package body Alr.Commands.Self_Update is
    -- Install_Alr --
    -----------------
 
-   procedure Install_Alr (Dest_Base, Extract_Bin : Any_Path) is
-      use AAA.Strings;
+   procedure Install_Alr (Dest_Base, Extracted_Bin : Any_Path) is
       use Alire.OS_Lib.Operators;
 
       Dest_Bin   : constant Any_Path := Dest_Base / Alr_Bin;
@@ -187,11 +190,18 @@ package body Alr.Commands.Self_Update is
         Dest_Base / (Dirs.Temp_Name (Length => 16));
    begin
       if Dirs.Is_File (Dest_Bin) then
-         Dirs.Adirs.Rename (Dest_Bin, Backup_Bin);
+         begin
+            Dirs.Adirs.Rename (Dest_Bin, Backup_Bin);
+         exception
+            when E : others =>
+               Reportaise_Command_Failed
+                 ("could not back up existing binary: "
+                  & Ada.Exceptions.Exception_Message (E));
+         end;
       end if;
 
       begin
-         Dirs.Adirs.Copy_File (Extract_Bin, Dest_Bin);
+         Dirs.Adirs.Copy_File (Extracted_Bin, Dest_Bin);
       exception
          --  if operation failed and a backup was made, restore previous
          when E : others =>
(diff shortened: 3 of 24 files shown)


src/alire/alire-github.adb
```diff
@@ -452,4 +461,12 @@ package body Alire.GitHub is
       --  removing the draft status we'll get a notification?
    end Request_Review;
 
+   function Get_Latest_Alire_Release return String is
```
Fabien-Chouteau, line 468: Sub-program comment box.

src/alr/alr-commands-self_update.adb
```diff
@@ -0,0 +1,316 @@
...
+         declare
+            V     : constant Semver.Version :=
+              Semver.Parse (Cmd.Force_Version.all);
+            V_Img : constant String := Semver.Image (V);
+         begin
+            if V_Img /= Cmd.Force_Version.all then
+               Trace.Info ("version string parsed as '" & V_Img & "'");
+            end if;
+            if V < Alire.Version.Current then
+               Trace.Warning
```
Fabien-Chouteau, line 71: Maybe request user confirmation here.

src/alr/alr-commands-self_update.adb
```diff
@@ -0,0 +1,316 @@
...
+                 ("you are currently on a nightly release (v"
+                  & Semver.Image (Alire.Version.Current)
+                  & ")");
+               Trace.Warning
+                 ("upgrading to latest will downgrade Alire to v"
+                  & Semver.Image (V));
+            elsif V = Alire.Version.Current then
+               Trace.Error
+                 ("you are already using the latest version of Alire");
+               Reportaise_Command_Failed
```
Fabien-Chouteau: I don't think it's an error, but more information.

src/alr/alr-commands-self_update.adb
```diff
@@ -0,0 +1,316 @@
...
+      if Dirs.Is_Directory (Path) then
+         return Path;
+      elsif Dirs.Adirs.Simple_Name (Path) /= Alr_Bin then
+         Reportaise_Command_Failed
+           ("invalid location: does not point to an existing directory or a "
+            & "file named `"
+            & Alr_Bin
+            & "`");
+      else
+         declare
```
Fabien-Chouteau, line 177: What is this care covering? Given `path/to/something/plop.txt`, it's going to install in `path/to/something/`?

src/alr/alr-commands-self_update.adb
```diff
@@ -0,0 +1,316 @@
...
+   -------------
+   -- Execute --
+   -------------
+
+   overriding
+   procedure Execute (Cmd : in out Command; Args : AAA.Strings.Vector) is
+      Dest_Path : constant String :=
+        (if Cmd.Location /= null and then Cmd.Location.all /= ""
+         then Cmd.Location.all
+         else Alire.OS_Lib.Subprocess.Locate_In_Path ("alr"));
```
Fabien-Chouteau: The `alr` that is running is not necessarily in PATH, is it?
I think we have other ways to find the path to the running executable.

src/alr/alr-commands-self_update.adb
```diff
@@ -0,0 +1,316 @@
...
+         Proceed         : UI.Answer_Kind;
+      begin
+         Dirs.Adirs.Create_Directory (Tmp_Dir);
+
+         Download_Result :=
+           Alire.OS_Lib.Download.File (Download_Url, Archive, Tmp_Dir);
+
+         if not Download_Result.Success then
+            Trace.Error ("could not download Alire binary for this platform.");
+            if T.Kind = Specific_Version then
```
Fabien-Chouteau: Maybe something we can do automatically. Not sure it's worth the complexity...

Fabien-Chouteau (commented):
Left a couple of comments.

How can we test this feature?

AldanTanneo:
> How can we test this feature?

You could use the --location argument to install the 'update' to a temporary folder, that's what I use for testing

src/alr/alr-commands-self_update.adb
AldanTanneo: The information is printed out, and then the user can give confirmation for the whole operation; is that not enough?

src/alr/alr-commands-self_update.adb
AldanTanneo: I don't understand what you mean; currently when given a path such as `path/to/something/plop.txt`, it will fail because `plop.txt` is not an existing directory or a valid `alr` binary. It's a bit strict maybe?

src/alr/alr-commands-self_update.adb
AldanTanneo: I don't know of a cross platform way to find the running executable, but I'm very much open to better ways of doing this!

src/alr/alr-commands-self_update.adb
AldanTanneo: By do automatically, do you mean: first fetch the release in the API to check that an artifact exists? Or something else

src/alr/alr-commands-self_update.adb
AldanTanneo: Maybe this should not be an error (it should exit with 0), but I think the default behaviour should still be to abort the command

src/alr/alr-commands-self_update.adb
Fabien-Chouteau: Ok, yes, that makes sense.

src/alr/alr-commands-self_update.adb
Fabien-Chouteau: Definitely, we don't want to reinstall the same version.

src/alr/alr-commands-self_update.adb
Fabien-Chouteau: I don't understand which cases this 3rd branch is handling.

1. Path is an existing directory
1. Path is an existing alr binary
1. ?

AldanTanneo: It's more like
1. Path is an existing directory: ok, return dir name
2. Path is not an existing directory, if it's a file named something other than `alr`, fail
3. Path is not an existing directory AND the file is named `alr`: check that the parent of that file (even if the file itself doesn't exist yet) is an existing directory

That way we can do `--location=~/.local/bin/`, or `--location=~/.local/bin/alr` even if the `alr` binary doesn't yet exist in that folder

src/alr/alr-commands-self_update.adb
Fabien-Chouteau: Here's the GNATcoll function to do that: https://github.com/AdaCore/gnatcoll-core/blob/7bd599e2cb9bc3088ce7d8f87b2f991f00ce6c87/core/src/gnatcoll-utils.ads#L264

src/alr/alr-commands-self_update.adb
Fabien-Chouteau: Check if the version exists, then check if the build exists.

src/alr/alr-commands-self_update.adb
AldanTanneo: `Executable_Path` does not exist in the current Alire pin of GNATCOLL, only `Executable_Location`, but that function does funky stuff:

> A special case is done for "bin" directories, which are skipped.

I'm not sure how I can use it properly then. Unless we update GNATCOLL in Alire.

Also, on a pickier note, there are almost no mentions of GNATCOLL in the `alr` package tree (admittedly, not as much in `alire`); I'd like to avoid it because it's a huge and not very modular dependency.

src/alr/alr-commands-self_update.adb
AldanTanneo: fixed in the [latest commit](https://github.com/alire-project/alire/pull/1963/commits/193f33b20232eded3fde1b48371b2bb94054fb60) by raising a special exception -- I don't know if that's really good but it works :P

src/alr/alr-commands-self_update.adb
AldanTanneo: Alternatively, we use `Ada.Command_Line.Command_Name`, but that might not be the correct command name (for instance with Linux' `execve`).

Fabien-Chouteau: @mosteo it's probably time to switch to the vanilla GNATcoll.

mosteo: > I think we have other ways to find the path to the running executable.

Wasn't the `resources` crate doing this? I might be mistaken. It doesn't depend of GNATCOLL.

> @mosteo it's probably time to switch to the vanilla GNATcoll.

Actually I was more inclined to try and not depend on GNATCOLL at all because it brings lots of baggage and it raises our minimum GNAT requirement above what Debian stable ships (although the new stable is expected later this year I think). Most uses are easily replaceable, but a few not so much. It might be a pipe dream and then we can simply go all-in. It's been a while I haven't been happy about keeping using that old version, if only for technical debt.

src/alr/alr-commands-self_update.adb
AldanTanneo: Done in https://github.com/alire-project/alire/pull/1963/commits/de7fc5d8dc4ed35437975675f7f2661852c28b5a

src/alr/alr-commands-self_update.adb
AldanTanneo: Solved in https://github.com/alire-project/alire/pull/1963/commits/ef9c67a4b36f49514c6d8450fdfe7f31f363c0b9, had to modify workflow files to fetch recursive modules

Fabien-Chouteau:
Thanks @AldanTanneo, only missing one or two tests now.

alire.toml
```diff
@@ -35,6 +35,7 @@ templates_parser = "^24" # Latests pre-Ada 2022 version
 toml_slicer = "~0.1"
 uri_ada = "^2.0"
 spdx = "~0.2"
+resources = "~0.1"
```
mosteo: I see you're not the first offender :-) but please keep dependencies in alphabetical order

alire.toml
```diff
@@ -131,6 +132,10 @@ url    = "https://github.com/mosteo/yeison"
 subdir = "yeison_12"
 commit = "20033cf90a3eb2eff69d9ff9d0820291aed79b57"
 
+[pins.resources]
+url    = "https://github.com/alire-project/resources"
+commit = "d3a83f35d7c0b586119fc19a568289cf21c87aa0"
```
mosteo: Likewise for pin ordering

src/alire/alire-github.adb
```diff
@@ -452,4 +461,50 @@ package body Alire.GitHub is
...
+
+   function Get_Latest_Alire_Release return String is
+      Response : constant GNATCOLL.JSON.JSON_Value :=
+        API_Call (Alire_Repo_API / Releases / Latest);
+      --  public endpoint
+   begin
+      return Response.Get ("tag_name");
+   end Get_Latest_Alire_Release;
+
+   function Check_Alire_Binary_Release (Tag, Archive : String) return Outcome
```
mosteo, line 480: Subprogram box

src/alr/alr-commands-self_update.ads
```diff
@@ -0,0 +1,40 @@
...
+
+   overriding
+   function Long_Description (Cmd : Command) return AAA.Strings.Vector
+   is (AAA.Strings.Empty_Vector
+       & "Update the Alire binary to the latest version, or to a version "
+       & "specified on the command line.");
+
+   overriding
+   function Usage_Custom_Parameters (Cmd : Command) return String
+   is ("[--location=<path/to/alr>] [--nightly] [--force=<version>]");
```
mosteo: I'm surprised this `--force` doesn't conflict with the global one :thinking: 

Even so, I prefer `--release` for example.

src/alr/alr-commands-self_update.adb
```diff
@@ -0,0 +1,536 @@
...
+
+   procedure Windows_Copy_And_Relaunch
+     (Cmd : Command; Dest_Base, Exe_Path : Any_Path)
+   with Pre => Plat.Current.On_Windows
+   is
+      --  Windows hack explanation:
+      --  When we detect that the self update will overwrite the currently
+      --  running binary, we do a little trick: we copy it to the temp folder
+      --  and relaunch it with proper arguments. It will then be able to
+      --  overwrite the old binary.
```
mosteo: Maybe add a comment that this is safer than renaming the running executable, as this way there's always a valid `alr` where intended.

src/alr/alr-commands-self_update.adb
```diff
@@ -0,0 +1,536 @@
...
+      --  to the command invocation ('Magic_Arg_Windows'), which contains a
+      --  hash of the build commit (for sanity checking).
+
+      use AAA.Strings;
+      use Alire.OS_Lib.Operators;
+
+      package AEE renames Alire_Early_Elaboration;
+
+      Copied_Bin    : constant Any_Path :=
+        Plat.Folders.Temp / (Dirs.Temp_Name (Length => 16) & Exe);
```
mosteo: I worry that some security policies will prevent running an exec from a temp folder (I have not experienced it, but I seem to have read something like that at some point?)

src/alr/alr-commands-self_update.adb
```diff
@@ -0,0 +1,536 @@
...
+      Relaunch_Args : Vector :=
+        Empty_Vector
+        & "/C"
+        & "start"
+        & "Alire Self-updater"
+        & String'("""" & Copied_Bin & """");
+      --  the `start` command in cmd.exe will launch a detached process, in a
+      --  separate console. see the exception section of `Execute` to see
+      --  the
```
mosteo: Incomplete comment

src/alr/alr-commands-self_update.adb
```diff
@@ -0,0 +1,536 @@
...
+      if AEE.Switch_VV then
+         Relaunch_Args.Append ("-vv");
+      elsif AEE.Switch_V then
+         Relaunch_Args.Append ("-v");
+      elsif AEE.Switch_Q then
+         Relaunch_Args.Append ("-q");
+      end if;
+      if UI.Not_Interactive then
+         Relaunch_Args.Append ("-n");
+      end if;
```
mosteo: IIRC we used to have something similar in `Alr.Spawn` to recreate the switches. Although we don't want to ever spawn an `alr` given the headaches (except for this case where it's truly inevitable), maybe move this bit of logic there just in case.

src/alr/alr-commands-self_update.adb
```diff
@@ -0,0 +1,536 @@
...
+      if Cmd.Nightly then
+         Relaunch_Args.Append ("--nightly");
+      elsif Cmd.Force_Version /= null and then Cmd.Force_Version.all /= "" then
+         Relaunch_Args.Append
+           (String'("""--force=" & Cmd.Force_Version.all & """"));
+      end if;
+
+      Relaunch_Args.Append (Magic_Arg_Windows);
+      Alire.OS_Lib.Subprocess.Checked_Spawn ("cmd.exe", Relaunch_Args);
+      OS_Lib.Bailout; --  quickly exit after `start` launched the second alr
```
mosteo, line 306: Is there a race condition here that could prevent the child to overwrite if the parent hasn't yet exited? 

I presume a proper solution would require the child enumerating running processes, but short of that, maybe add a small delay in the child?

src/alr/alr-commands-self_update.adb
```diff
@@ -0,0 +1,536 @@
...
+        Alire.OS_Lib.Subprocess.Unchecked_Spawn
+          ("cmd.exe",
+           Empty_Vector
+           & "/C"
+           & "start"
+           & "delayed pause"
+           & "/B"
+           & "cmd.exe"
+           & "/C"
+           & "ping -n 2 127.0.0.1 > nul & pause");
```
mosteo: Wouldn't a pure Ada solution with a dynamically allocated task doing the delay work here? (I understand here there's no concern like in the previous case in which the deletion requires a bg shell, but this is doing nothing.)

src/alr/alr-commands-self_update.adb
```diff
@@ -0,0 +1,536 @@
...
+            end if;
+            if V < Alire.Version.Current then
+               Trace.Warning
+                 ("downgrading to version "
+                  & V_Img
+                  & " (current: "
+                  & Semver.Image (Alire.Version.Current)
+                  & ")");
+               Trace.Warning
+                 ("previous versions may not have the `self-update` command");
```
mosteo: Nitpick but if this goes into 3.0 you could warn only when that's the case.

Given that I want the new testing out the door ASAP, I doubt there will be a 2.2

src/alr/alr-commands-self_update.adb
```diff
@@ -0,0 +1,536 @@
...
+      else
+         declare
+            Tag_Name : constant String :=
+              Alire.GitHub.Get_Latest_Alire_Release;
+            V        : constant Semver.Version :=
+              Semver.Parse (Tag_Name (Tag_Name'First + 1 .. Tag_Name'Last));
+         begin
+            if Alire.Version.Current.Pre_Release = "dev" then
+               Trace.Info
+                 ("Detected nightly version. Use --force="
```
mosteo: Hardcoded `--force` may bite later if switch name changed

src/alr/alr-commands-self_update.adb
```diff
@@ -0,0 +1,536 @@
...
+            return (Specific_Version, V);
+         end;
+      else
+         declare
+            Tag_Name : constant String :=
+              Alire.GitHub.Get_Latest_Alire_Release;
+            V        : constant Semver.Version :=
+              Semver.Parse (Tag_Name (Tag_Name'First + 1 .. Tag_Name'Last));
+         begin
+            if Alire.Version.Current.Pre_Release = "dev" then
```
mosteo: Rather than equal to `"dev"` I'd just check that pre-release string is not empty.

src/alr/alr-commands-self_update.adb
```diff
@@ -0,0 +1,536 @@
...
+               Trace.Info
+                 ("Detected nightly version. Use --force="
+                  & Semver.Image (V)
+                  & " to update to the latest stable release.");
+               return (Kind => Nightly);
+            elsif V < Alire.Version.Current then
```
mosteo, line 99: I'm not sure this has to be a pre-release? Given line 86.

src/alr/alr-commands-self_update.adb
```diff
@@ -0,0 +1,536 @@
...
+            end if;
+         end;
+      end if;
+   end Dest_Path_Validate;
+
+   -----------------
+   -- Install_Alr --
+   -----------------
+
+   procedure Install_Alr (Dest_Base, Extract_Bin : Any_Path) is
```
mosteo: `s/Extract/Extracted/` (if I'm correctly understanding the intention here).

src/alr/alr-commands-self_update.adb
```diff
@@ -0,0 +1,536 @@
...
+      use AAA.Strings;
+      use Alire.OS_Lib.Operators;
+
+      Dest_Bin   : constant Any_Path := Dest_Base / Alr_Bin;
+      Backup_Bin : constant Any_Path :=
+        Dest_Base / (Dirs.Temp_Name (Length => 16));
+   begin
+      if Dirs.Is_File (Dest_Bin) then
+         Dirs.Adirs.Rename (Dest_Bin, Backup_Bin);
+      end if;
```
mosteo: No handling of rename failure?

src/alr/alr-commands-self_update.adb
```diff
@@ -0,0 +1,536 @@
...
+         when Plat.MacOS =>
+            Alire.OS_Lib.Subprocess.Checked_Spawn
+              ("xattr",
+               Empty_Vector & "-d" & "com.apple.quarantine" & Dest_Bin);
+            Alire.OS_Lib.Subprocess.Checked_Spawn
+              ("chmod", Empty_Vector & "+x" & Dest_Bin);
+
+         when Plat.Windows | Plat.OS_Unknown =>
+            null;
+      end case;
```
mosteo: Refactor this to `Alire.OS_Lib.Mark_Executable` or so?

mosteo (changes requested):
Nothing very substantive in my review, but there was quite a bit to parse here.

I think just in case it should always keep a copy of the old version as `alr-x.x.x-shortcommit[.exe]`. The error management here is very comprehensive but if a user downgrades to a version without `self-update` this could save some hassle. And there's always Murphy.

In any case, neat effort!

src/alr/alr-commands-self_update.adb
AldanTanneo: On Linux /tmp is sometimes mounted as noexec and I've heard of that, but on windows I'm not familiar with the issue. Will research it further.

src/alr/alr-commands-self_update.adb
AldanTanneo: Well, the child will always have _some_ delay, because it has to
- reparse CLI args
- fetch the release and download the binary

and even after all that, unless it's running in non-interactive mode, it will ask for confirmation to overwrite.

mosteo: Ok, I was thinking that the first thing would be the cleanup.

src/alr/alr-commands-self_update.adb
AldanTanneo: I wanted it to be asynchronous (start /B will launch a detached process but keep it in the same console), because that way exception messages would have time to show up... I would welcome other ideas though.

mosteo: I may be confused here about what you need. Must the alr process have exited in this case, or is it just a matter of it continuing but not exiting for a while? If the latter (this is what I was thinking when I read this part):
```
   task type X;
   task body X is
   begin
      delay 3.0;
   end X;
   type AX is access all X;
   MX : constant AX := new X;
```
The dynamic task is not waited for by its creator, but it is by the Ada runtime when everything has completed.

If the former then sure I cannot think of anything.

AldanTanneo: It must continue and do the top-level exception handling, but it does not need to have exited. I'll check the behaviour of this solution, thanks!

src/alr/alr-commands-self_update.adb
AldanTanneo: Well yes but there would be something weird happening: the latest release from github would be lower than the current version, even though it's not a prerelease

Shall I say 'preview version'?

src/alr/alr-commands-self_update.ads
AldanTanneo: Changed it to `--release`, it is indeed a better name

src/alr/alr-commands-self_update.adb
AldanTanneo: It works very well :)

src/alr/alr-commands-self_update.adb
AldanTanneo: Changed the name, but still hardcoding the string value for now... Should I move every switch name in a constant?

src/alr/alr-commands-self_update.adb
AldanTanneo: I updated the wording to 'preview' version (aka, a version above latest that somehow doesn't have a pre-release tag in its version string)

src/alr/alr-commands-self_update.adb
AldanTanneo: What's expected here, a function that returns a string vector?

src/alr/alr-commands-self_update.adb
AldanTanneo: I found some references to this, although I expect it to happen only rarely on machines where the user is not the administrator. I don't really know if we could consider this an acceptable risk. We would need people to test it out to see how widespread the issue is.

src/alr/alr-commands-self_update.adb
mosteo: OK, I hadn't realized the implications, but this way everything is covered.

src/alr/alr-commands-self_update.adb
mosteo: What about just using the same folder where the other `alr.exe` is?

src/alr/alr-commands-self_update.adb
mosteo: That would be OK, yes.

src/alr/alr-commands-self_update.adb
mosteo: Not every switch. I typically do it only when a switch name appears in more than one place.

src/alr/alr-commands-self_update.adb
AldanTanneo: Somewhat related question (while reading the global switch code), why do we never enable color output on windows?

mosteo: We are using the color flag for the use of both ANSI and Unicode, and on Windows we were getting many reports of strange characters being printed. We haven't investigated further if there's an easy way of detecting TTY capabilities on Windows...

My understanding is that recent Windows versions support both ANSI and Unicode, but I'm not really sure if this is enabled by default on both CMD/PowerShell, and since when.

AldanTanneo:
Hmm. Testing is going to be... complex. Even ignoring network failures that can happen, apparently sending many unauthenticated requests to the github api in parallel results in a ratelimit (we could add the option to use a GitHub token when one is present in the environment?)

Seb-MCaw:
> Hmm. Testing is going to be... complex. Even ignoring network failures that can happen, apparently sending many unauthenticated requests to the github api in parallel results in a ratelimit (we could add the option to use a GitHub token when one is present in the environment?)

If I understand correctly, everything ultimately calls out to `curl`; would it be feasible to mock `curl` with [a `MockCommand`](https://github.com/alire-project/alire/blob/63824271d9a23071260ad454546f8d24201638c7/testsuite/drivers/helpers.py#L386)? 

In addition/alternatively, you could keep online tests but marked as [`skip_local`](https://github.com/alire-project/alire/blob/63824271d9a23071260ad454546f8d24201638c7/testsuite/skels/no-index/test.yaml#L6).

AldanTanneo:
@mosteo if the GITHUB_TOKEN for github actions only has read (not write) permissions for the repo, can we add it to the environment for tests? I think the GitHub API might need it to not be rate-limited (see `testsuite/tests/self_update/latest/test.py` to see how I inject the auth into the curl invocation when needed)

mosteo:
> can we add it to the environment for tests?

I guess there's no problem. Also I suppose we can create a dedicated token with only the required permissions?

AldanTanneo:
The `alr` binary available in release artifacts for 2.1 doesn't run on Ubuntu 22.04 (arm) :cry:

mosteo:
> The `alr` binary available in release artifacts for 2.1 doesn't run on Ubuntu 22.04 (arm) 😢

I think we used ubuntu-24.04-arm for that release. Oh well, we may have a 2.1.1 shortly that would fix that.

testsuite/tests/self_update/latest/test.yaml
```diff
@@ -0,0 +1,3 @@
+driver: python-script
+indexes:
+    compiler_only_index: {}
```
mosteo, line 3: IIUC the mock curl, this is still going online, right? Then this test should be marked with `skip_network` (see in `/testsuite/skels/no-index/test.yaml`)

src/alr/alr-commands-self_update.ads
```diff
@@ -0,0 +1,43 @@
...
+   function Usage_Custom_Parameters (Cmd : Command) return String
+   is ("");
+
+   overriding
+   procedure Setup_Switches
+     (Cmd    : in out Command;
+      Config : in out CLIC.Subcommand.Switches_Configuration);
+
+private
+   Switch_Location : constant String := "location";
```
mosteo: Blank line after `private`

testsuite/tests/self_update/specific_version/test.yaml
```diff
@@ -0,0 +1,3 @@
+driver: python-script
+indexes:
+    compiler_only_index: {}
```
mosteo, line 3: Same about `skip_network`

testsuite/tests/self_update/specific_version/test.py
```diff
@@ -0,0 +1,64 @@
...
+    return p.stdout.decode(errors="replace")
+
+
+with MockCommand("curl", curl_script, "curl_override"):
+    out = run_alr(["self-update", "--release=2.1.0"])
+
+    if (
+        "alr-2.1.0-bin-aarch64-linux.zip" in out
+        and __import__("platform").freedesktop_os_release().get("VERSION_ID") == "22.04"
+    ):  # HACK: no working alr 2.1.0 for ubuntu 22.04
```
mosteo: `for ubuntu 22.04 ARM`?

testsuite/tests/self_update/specific_version/test.py
```diff
@@ -0,0 +1,64 @@
...
+        print("SUCCESS")
+        __import__("sys").exit()
+
+    out = None
+    for i in range(10):
+        out = run_alr(["--version"])
+        if "2.1" in out:
+            break
+        time.sleep(1)
```
mosteo: Where is the concurrency coming from?

mosteo (changes requested):

testsuite/tests/self_update/specific_version/test.py
AldanTanneo: The windows version, when it's self replacing, spawns another console / cmd process, so we need to wait until it's finished 

I don't know if this is the best way to do it though

mosteo: Ah yes. I was looking only in the test.

This may be OK, but please add a comment before that block explaining what's going on.

mosteo (approved):

mosteo:
Ready for merging, @AldanTanneo ?

AldanTanneo:
I would think so! Further issues can go in another PR

PR #1976, opened 2025-07-06
fix: ANSI-aware transient progress messages

mosteo:
Fixes the extra invisible length that resulted in spurious empty lines for some commands (mainly when updating dependencies).


Seb-MCaw:
[The pin in `alire.toml` is still on commit `a8561d3`.](https://github.com/alire-project/alire/blob/02575f1f5b2e7e3c3fc949f79ef12cf72fe506b3/alire.toml#L110)

mosteo:
> [The pin in `alire.toml` is still on commit `a8561d3`.](https://github.com/alire-project/alire/blob/02575f1f5b2e7e3c3fc949f79ef12cf72fe506b3/alire.toml#L110)

Well spotted, Seb, thanks.

Issue #1314, opened 2023-01-28
Command lines containing spaces inside arguments not supported in editor.cmd


mgrojo:
Related to #1313

As commented in the emacs ada-mode users list:

>> Thanks. Now, I am able to set it in this way:
>>
>> alr config --set --global editor.cmd 'emacs
>> --eval=(ada-build-prompt-select-prj-file"${GPR_FILE}") ${GPR_FILE}'

>> Note that there is no space between the function and the string
>> argument. Otherwise, Alire splits the argument and Emacs is not able
>> to evaluate it correctly. Don't know if that could be considered a bug
>> in Alire or not, but I couldn't find a proper way to escape the space.
> 
> The space problem sounds like a bug, but it could be in your shell
> rather than in Alire. It's simpler just to edit the alire config file
> directly; it's ~/.config/alire/config.toml

When the config file is edited and space added:

```
$ alr edit
Editing crate with: ['emacs' '--eval=(ada-build-prompt-select-prj-file', '"play_2048.gpr")', 'play_2048.gpr']
```

And then Emacs errors:

`command-line-1: End of file during parsing`

That can be attributed to Alire, it's splitting the command assuming spaces as separator, so command lines containing spaces inside arguments are not supported. Using shell syntax (e.g. single quotes) inside the config file doesn't work either.

It's a very limited need and maybe difficult to solve it portably, but I think it deserves its own issue.

mgrojo:
This is where the command line is split:
https://github.com/alire-project/alire/blob/49c04043a3a8d259257729191f5bfe4a104385e5/src/alr/alr-commands-edit.adb#L67

I guess fixing it is not easy and would imply user-visible changes, but the workaround of joining function and string luckily worked.

mosteo:
Thanks, @mgrojo, certainly worth to keep tabs on this but, as you say, not trivial unless we locate some lib with ready-to-use cleverer splitting of strings.

PR #1993, opened 2025-08-04
feat: Add shell-style quoting and escaping to edit and run commands

Mivirl:
Addresses https://github.com/alire-project/alire/issues/1314

Allows specifying a custom editor command with quoted spaces

`${CRATE_ROOT}` and `${GPR_FILE}` are still replaced directly, not using parameter expansion.

The argument to `alr run -a` has been changed to use the same quoting rules as `alr edit` instead of using the OS-specific rules provided by `GNAT.OS_Lib.Argument_String_To_List`.

Quoting and token recognition uses the behavior for the shell command language from the POSIX.1-2024 standard with a few changes:
- [2.2.2 Single-Quotes](https://pubs.opengroup.org/onlinepubs/9799919799/utilities/V3_chap02.html#tag_19_02_02)
  - No changes
- [2.2.3 Double-Quotes](https://pubs.opengroup.org/onlinepubs/9799919799/utilities/V3_chap02.html#tag_19_02_02)
  - The dollar-sign is not used for parameter expansion, command substitution, or arithmetic expansion
  - The backquote is not used for command substitution
  - The backslash is an escape character within double quotes only when immediately followed by `\` or `"`
- [2.3 Token Recognition](https://pubs.opengroup.org/onlinepubs/9799919799/utilities/V3_chap02.html#tag_19_03)
  - Here-documents are not supported
  - Parameter expansion, command substitution, and arithmetic expansion are not supported
  - Comments are not supported
  - Alias substitution is not supported

##### PR creation checklist
- [x] `BREAKING.md` has been updated for major changes in `alr`, minor/major in catalog format.


src/alire/alire-os_lib-subprocess.adb
```diff
@@ -175,6 +327,37 @@ package body Alire.OS_Lib.Subprocess is
       Dim_Output          : Boolean := True) return Integer
    is (Spawn (Command, Arguments, Understands_Verbose, Dim_Output));
 
+   ---------------
+   -- Spawn_Raw --
+   ---------------
+
+   procedure Spawn_Raw (Command   : String;
+                        Arguments : AAA.Strings.Vector)
```
mosteo, line 336: If I'm reading the diff correctly, there are now two `Spawn_Raw` at `OS_Lib` and `OS_Lib.Subprocess`? Is there any strong reason that precludes having only one of them?

mosteo (changes requested):
A neat improvement. Only, see the comment about the duplicated `Spawn_Raw`.

src/alire/alire-os_lib-subprocess.adb
Mivirl: I don't have a strong reason, I just left the original `Spawn_Raw` since I wasn't sure about other changes I would need to make if I removed it. The main thing that made me hesitate is that one's in `Alr.OS_Lib` and the other's in `Alire.OS_Lib.Subprocess`, where I wasn't sure about using a new exception, so I had the original `Spawn_Raw` wrap the call and raise the same exception as before.

But it's clearer just to have one `Spawn_Raw`, so I can do that instead. Would it be better to have a separate `Child_Failed` exception or just raise a `Checked_Error` (and have `Checked_Spawn_Raw` and `Unchecked_Spawn_Raw` variants for consistency)? The exception is only used here: https://github.com/alire-project/alire/blob/a06aac7cf88f77d2fc4bb9dd7ae9de2155c3236b/src/alr/alr-commands.adb#L635-L650 and the only difference I see is how the error is surfaced.

mosteo: In truth, `Child_Failed` is a remnant from old times when `alr` used to respawn itself for most operations. We could think of getting rid of it in a later cleanup. I think what you did is OK now.

mosteo (approved):

mosteo:
Ideally we would need a Python test for the end-user behavior. @Mivirl, are you able/willing to do it?

testsuite/tests/edit/split-arguments/test.py
```diff
@@ -0,0 +1,25 @@
...
+from subprocess import run
+
+import os, re, shutil
+
+alr_path = os.environ['ALR_PATH']
+
+p = run_alr('settings', '--global', '--set', 'editor.cmd', 'sh -c "echo Argument splitting test"')
+
+target = 'noop_1.0.0_filesystem'
+p = run_alr('get', 'noop=1.0-default', quiet=True)
```
mosteo: You don't need to retrieve a crate, it's simpler to use `init_local_crate()` (which will `cd` into the created crate too)

testsuite/tests/edit/split-arguments/test.py
```diff
@@ -0,0 +1,25 @@
...
+os.chdir(target)
+
+p = run([alr_path, "edit"], capture_output=True)
+assert_match('Argument splitting test', p.stdout.decode(), flags=re.S)
```
mosteo: Why not use `p.out` from `p = run_alr()` here?

testsuite/tests/edit/split-arguments/test.py
```diff
@@ -0,0 +1,25 @@
...
+
+os.chdir('..')
+shutil.rmtree(target)
```
mosteo: No need to do clean-up, the test driver takes care of it.

mosteo (changes requested):
Thanks for the tests, Mivirl. Only minor comments about one of them.

I'm not sure why the Windows test is failing. There's this curious effect where the last 'e' from example is disappearing:

https://github.com/alire-project/alire/actions/runs/18571537329/job/52946369421?pr=1993#step:17:1581

testsuite/tests/edit/split-arguments/test.py
Mivirl: I tried doing that first, but switched when the output wasn't captured with `run_alr`. I see now the issue was just that I didn't realize `quiet` defaults to `True`. Explicitly setting it to `False` works fine.

Mivirl:
Thanks for the feedback!

> I'm not sure why the Windows test is failing. There's this curious effect where the last 'e' from example is disappearing:
> 
> https://github.com/alire-project/alire/actions/runs/18571537329/job/52946369421?pr=1993#step:17:1581

Not sure why that's happening either. I don't currently have a windows environment ready to debug it, but I'm going to set one up so I can figure it out.

Seb-MCaw:
By default, Windows GNAT [does some postprocessing on command line arguments](https://docs.adacore.com/live/wave/gnat_ugn/html/gnat_ugn/gnat_ugn/platform_specific_information.html#disabling-command-line-argument-expansion). I don't know if it's functioning as intended, but the handling of quotes is decidedly weird.

Adding
```Ada
Do_Argv_Expansion : Integer := 0;
pragma Export (C, Do_Argv_Expansion, "__gnat_do_argv_expansion");
```
to the declarative part of `Echo_Arguments` appears to be sufficient to get the test to pass (though I believe this only disables a subset of the postprocessing).

Mivirl:
> By default, Windows GNAT [does some postprocessing on command line arguments](https://docs.adacore.com/live/wave/gnat_ugn/html/gnat_ugn/gnat_ugn/platform_specific_information.html#disabling-command-line-argument-expansion). I don't know if it's functioning as intended, but the handling of quotes is decidedly weird.
> 
> Adding
> 
> ```ada
> Do_Argv_Expansion : Integer := 0;
> pragma Export (C, Do_Argv_Expansion, "__gnat_do_argv_expansion");
> ```
> 
> to the declarative part of `Echo_Arguments` appears to be sufficient to get the test to pass (though I believe this only disables a subset of the postprocessing).

Thanks, that's good to know

Just adding reference links below to a few commits involving this feature as I'm looking into it:
- [Add support for wildcard expansion for Unicode parameters on Win32](https://gcc.gnu.org/git/?p=gcc.git;a=commitdiff;h=46e26fe337a12800faa9ef010eebb49266a014f7)
- [Do not expand quoted arguments](https://gcc.gnu.org/git/?p=gcc.git;a=commitdiff;h=6d47b1e3e982e90a53c20a3f6a7a7ed3cd78132d)
- [Add weak symbol __gnat_do_argv_expansion](https://gcc.gnu.org/git/?p=gcc.git;a=commitdiff;h=c6c8d059421efe3d025483c11c66941a8d1926a4)
- [Current version](https://gcc.gnu.org/git/?p=gcc.git;a=blob;f=gcc/ada/rtinit.c;h=6a135529870b2deaadb3cef461666b50f439a667;hb=refs/heads/master#l426)

Mivirl:
> By default, Windows GNAT [does some postprocessing on command line arguments](https://docs.adacore.com/live/wave/gnat_ugn/html/gnat_ugn/gnat_ugn/platform_specific_information.html#disabling-command-line-argument-expansion). I don't know if it's functioning as intended, but the handling of quotes is decidedly weird.
> 
> Adding
> 
> ```ada
> Do_Argv_Expansion : Integer := 0;
> pragma Export (C, Do_Argv_Expansion, "__gnat_do_argv_expansion");
> ```
> 
> to the declarative part of `Echo_Arguments` appears to be sufficient to get the test to pass (though I believe this only disables a subset of the postprocessing).

Thanks! I've added this to `Echo_Arguments`, since I think it makes sense to disable for this test.

I've also submitted a [bug report for GNAT](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=122367) about the disappearing 'e'.

mosteo (approved):

mosteo:
Looks good to me. @Seb-MCaw, since you seem familiar with these issues, any final comment?

Seb-MCaw:
> you seem familiar with these issues

Not at all; this is the first time I've encountered this argument expansion behaviour.

The PR looks good to me (except for the very minor detail that `origins.archive.download_cmd` is not mentioned in the changelogs).

It's fortunate we've run into this, though, since it raises the wider issue of `alr` performing the same postprocessing:
```
C:\xxx> python echo_args.py 'Hello \"W\"o\"r\"l\"d\"
'Hello
"W"o"r"l"d"

C:\xxx> alr exec -- python echo_args.py 'Hello \"W\"o\"r\"l\"d\"
Hell
World

```
I guess this at least needs to be documented somewhere, if not disabled?

---

Edit: On closer inspection, the `"` effect appears to be a `GNAT.OS_Lib.Spawn` behaviour, since I still get:
```
C:\> alr \"W\"o\"r\"l\"d\"
ERROR: Unrecognized command: "W"o"r"l"d"
```
Nonetheless, this is Windows-only, and isn't mentioned in our documentation as far as I'm aware.

mosteo:
> I guess this at least needs to be documented somewhere, if not disabled?

For the disabling, do you mean with `__gnat_do_argv_expansion`? Any collateral damage that we may expect from that?

Seb-MCaw:
> For the disabling, do you mean with __gnat_do_argv_expansion?

Yes; that's the only way I'm aware of.

> Any collateral damage that we may expect from that?

Obviously this would disable wildcard expansion etc. for all `alr` commands, and there are no doubt some situations where the current behaviour could potentially be convenient. The only results of a search for `__gnat_do_argv_expansion` in the GCC sources are the `if` statements surrounding the wildcard expansion and quote removal logic, so there aren't any other effects which are immediately obvious. I'm really not a toolchain expert, though, so there could definitely be ramifications I'm missing.

mosteo:
I think we would be better served by not having unsuspected expansions done for us. I suppose this comes from Windows where the shell is not doing the expansion? *Frustration noises*.

mosteo:
If I've not lost track at some point, the only thing remaining is to add
```
Do_Argv_Expansion : Integer := 0;
pragma Export (C, Do_Argv_Expansion, "__gnat_do_argv_expansion");
```
somewhere, perhaps in `Alire_Early_Elaboration`, and crossing fingers.

Seb-MCaw:
> the only thing remaining is to add ...

Ah, I see. I (and presumably Mivirl) was assuming that would be out-of-scope for this PR.

Note that this would be a potentially breaking change (albeit a fairly obscure one), so I guess it would need another brief entry in `user-changes.md`?

mosteo:
Ah, it's OK to me to leave that for another PR. Also it would make it better isolated if it turns out to be a bad idea. I'll create an issue for that

mosteo (approved):

PR #2003, opened 2025-08-29
Fix workflow on aarch64 linux

AldanTanneo:
Try to fix Linux ARM nightly builds

### DIFF
--- a/.github/workflows/ci-release.yml
+++ b/.github/workflows/ci-release.yml
@@ -69,6 +69,30 @@ jobs:
       with:
         submodules: recursive
 
+    # Install GNAT, we only need a system compiler for Ubuntu ARM until we have
+    # Alire for ARM64/Linux (on Ubuntu 22.04) release.
+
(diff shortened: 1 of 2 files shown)


mosteo:
We cannot yet remove the special case of aarch64 as the build was done in Ubuntu 24 and we hit the missing glib version thing. Any new release should fix that.

AldanTanneo:
Btw: the `macos-13` runner will be retired by the end of the year. It's going to become harder and harder to support x86_64 macos... (the same issue applies to GNAT FSF builds)

mosteo (approved):

mosteo:
Merged, thanks.

mosteo:
> Btw: the `macos-13` runner will be retired by the end of the year. It's going to become harder and harder to support x86_64 macos... (the same issue applies to GNAT FSF builds)

OK, thanks for the heads-up. I guess well test what we can as long as we can...

PR #2008, opened 2025-09-27
Fix wording

godunko:
##### PR creation checklist
- [ ] A test is included, if required by the changes.


mosteo:
This is kind of fuzzy. A crate is only the name (`mycrate`) whereas a release includes a version. On disk, you always have a particular version, so even if its a temporary one, its closer to a release than simply a crate. It's true that people use them interchangeably, or default to crate even when talking about releases...

Issue #1670, opened 2024-04-23
Two alr processes in the same folder


reznikmm:
ALS launches two `alr` processes in the same file to find a project file and environment (`alr --non-interactive printenv` and `alr --non-interactive show`). It seems they unable to work properly.

**To Reproduce**
1. Unpack reproducer.tar.gz from [ALS #1192](https://github.com/AdaCore/ada_language_server/issues/1192). I used a docker container built by vscode to have a clean environment.
2. Change directory to `no-code-navigation` and run 
   ```sh
   for J in `seq 1 10`; do alr -n show > /tmp/alr-$J.log 2>&1 & done; wait
   ```
3. For failed `alr` process you will see `Exit 1`. Check `/tmp/alr-$J.log` for their output

**Expected Behavior**
Generally speaking it would be nice to success all concurrent `alr` executions, but I see that it's probable hard to achieve and verify by CI/tests. So I'm going to change ALS to launch `alr` processes in a sequence.

**`alr` version**
> alr version:               2.0.1

**`alr` logs**
```
'alr --non-interactive show' failed:
Warning: Could not completely remove /root/.local/share/alire/releases/xmlada_24.0.0_ae5a015b/alire/alire.toml.upstream
ERROR: file "/root/.local/share/alire/releases/xmlada_24.0.0_ae5a015b/alire/alire.toml.upstream" could not be deleted
ERROR: alr encountered an unexpected error, re-run with -d for details.
ERROR: error location: 0x6347e939877f ada__directories__delete_file at ???

'alr --non-interactive printenv' failed:
Warning: Could not completely remove /root/.local/share/alire/releases/xmlada_24.0.0_ae5a015b/alire/alire.toml.upstream
ERROR: file "/root/.local/share/alire/releases/xmlada_24.0.0_ae5a015b/alire/alire.toml.upstream" could not be deleted
ERROR: alr encountered an unexpected error, re-run with -d for details. ERROR: error location: 0x62f95b2fa77f ada__directories__delete_file at ???

'alr --non-interactive printenv' failed:
ERROR: file "alire" not found ERROR:
alr encountered an unexpected error, re-run with -d for details. ERROR: error location: 0x579f6a2450c9 ada__directories__rename at ???

Warning: Could not completely remove /root/.local/share/alire/releases/gnatcoll_24.0.0_11c512d1
Warning: Contents listing failed with code:  2
[...]

PR #2011, opened 2025-10-04
fix: remove race condition across two instances of `alr`

mosteo:
Temporary files now include the PID of the creator process


Fabien-Chouteau:
If tmp filenames are randomized, why do we have this issue in the first place?

mosteo:
This happened when two processes were run in the same folder, and since we are using the clock to initialize the seed I guess clashes were not impossible. But also, the removal of the temporary folder on finalization was a race condition in which another alr would try to create a temp file and the folder was already gone. I seem to remember @reznikmm reporting something along these lines.

Fabien-Chouteau:
I am worried about the security risk of using PID in temp files names. But I guess PID + random is ok.

mosteo:
> I am worried about the security risk of using PID in temp files names. But I guess PID + random is ok.

I can derive a hash from the PID to be extra sure.

eliericha:
> This happened when two processes were run in the same folder, and since we are using the clock to initialize the seed I guess clashes were not impossible. But also, the removal of the temporary folder on finalization was a race condition in which another alr would try to create a temp file and the folder was already gone. I seem to remember @reznikmm reporting something along these lines.

For traceability, it was this issue #1670

PR #2039, opened 2025-12-11
fix macos CI

AldanTanneo:
adds gcc 15 to toolchain CI and switch x64 macos image to `macos-15-intel`.

also updates Github actions templates to use `macos-15-intel`.

Needed in #2038


mosteo (approved):

mosteo:
Thanks, @AldanTanneo. I had this in my radar since months ago and still I missed the deadline.

PR #2049, opened 2025-12-22
replace --no-color with a more versatile --color=WHEN switch

AldanTanneo:
The `--color` switch accepts `always`, `auto` and `never`.

`--color=always` is useful when one wants to keep color output in a redirection. Notably: in GitHub actions?

The default value is `auto`, which has the same behaviour as before without the `--no-color` switch.

Disabling colors is done with the `never` value. It also follows the `NO_COLOR` environment variable, like before, but the `always` value has priority.

The `--no-color` switch will still work, but is deprecated.

##### PR creation checklist

### DIFF
--- a/doc/user-changes.md
+++ b/doc/user-changes.md
@@ -17,7 +17,7 @@ The new global switch allows specifying `--color=always`, for instance when
 one wants to keep color output in a redirection. The default value is `auto`,
 which has the same behavior as before without the `--no-color` switch.
 
-Disabling colors in done with the `never` value. It also follows the
+Disabling colors is done with the `never` value. It also follows the
 `NO_COLOR` environment variable, like before, but the `always` value has
 priority.
 
--- a/doc/user-changes.md
+++ b/doc/user-changes.md
@@ -11,6 +11,19 @@ stay on top of `alr` new features.
 The solver timeout has been increased from 5 seconds to 10 seconds by default.
 This should avoid hitting it on complex dependencies on slower machines.
 
+### Replace `--no-color` with a new `--color[=WHEN]` switch
+
+The new global switch allows specifying `--color=always`, for instance when
+one wants to keep color output in a redirection. The default value is `auto`,
+which as the same behavior a before with the `--no-color` switch.
+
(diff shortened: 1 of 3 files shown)


doc/user-changes.md
```diff
@@ -6,6 +6,19 @@ stay on top of `alr` new features.
 
 ## Release `3.0`
 
+### Replace `--no-color` with a new `--color[=WHEN]` switch
+
+The new global switch allows specifying `--color=always`, for instance when
+one wants to keep color output in a redirection. The default value is `auto`,
+which as the same behavior a before with the `--no-color` switch.
```
mgrojo: Did you mean this?
```suggestion
which has the same behavior as before without the `--no-color` switch.
```

doc/user-changes.md
```diff
@@ -6,6 +6,19 @@ stay on top of `alr` new features.
...
+
+Disabling colors in done with the `never` value. It also follows the
```
mgrojo:
```suggestion
Disabling colors is done with the `never` value. It also follows the
```

doc/user-changes.md
AldanTanneo: Yeah I did, I must have been very tired when I wrote this haha

Fabien-Chouteau (approved):

src/alr/alr-commands.adb
```diff
@@ -140,6 +143,28 @@ package body Alr.Commands is
...
+      if Stripped_Value = "always" then
+         Color_Config := Always;
+
+      elsif Stripped_Value = "never" then
+         Color_Config := Never;
+
+      elsif Stripped_Value /= "auto" and then Stripped_Value /= "" then
+         Trace.Warning
+           ("Unknown value '" & Stripped_Value & "' passed for "
+            & TTY.Terminal (Switch) & ". Defaulting to 'auto'.");
```
mosteo: You can suggest here valid values using `Did_You_Mean`, see around line 514 of this same file.

AldanTanneo: Done!

mosteo (approved):

PR #2062, opened 2026-01-13
feat: structured test runner output

AldanTanneo:
POC PR: this is a proposed structured output format for the built-in test runner.

It works but I don't know if we want something else (some standard format?). The structured output is still disabled for external runners.

Also added test duration in regular and structured output.

### DIFF
--- a/src/alire/alire-test_runner.adb
+++ b/src/alire/alire-test_runner.adb
@@ -1,3 +1,4 @@
+with Ada.Calendar;
 with Ada.Containers.Indefinite_Ordered_Maps;
 with Ada.Containers.Indefinite_Vectors;
 with Ada.Strings.Fixed;
@@ -27,9 +28,12 @@ package body Alire.Test_Runner is
 
       procedure Init;
       --  Initialize the driver
-      procedure Pass (Test_Name : String);
+      procedure Pass (Test_Name : String; Start_Time : Ada.Calendar.Time);
       --  Report a passing test with a message
-      procedure Fail (Test_Name, Reason : String; Output : AAA.Strings.Vector);
+      procedure Fail
+        (Test_Name, Reason : String;
+         Start_Time        : Ada.Calendar.Time;
+         Output            : AAA.Strings.Vector);
       --  Report a failing test with a message and its output
       function Total_Count return Natural;
       --  Get the total number of tests that have been run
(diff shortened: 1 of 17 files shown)


src/alire/alire-test_runner.adb
```diff
@@ -2,68 +2,132 @@ with Ada.Containers.Indefinite_Ordered_Maps;
...
+with LML.Output.Factory;
 
 package body Alire.Test_Runner is
-
    use Alire.Utils;
 
-   protected Driver is
-      --  Protected driver for synchronising stats and output
+   package Driver is
+      --  Driver for synchronising stats and output
```
mosteo, line 27: Why isn't the synchronization needed anymore?

mosteo:
This looks good to me.

> some standard format?

Actually, Alire used to output results in JUnit format, and the code is still at `Alr.Testing.JUnit`. Not sure what demand for that is there these days.

src/alire/alire-test_runner.adb
AldanTanneo: It's a leftover of when I wanted to use tasking to spawn tests in parallel (baaaad idea). There is no concurrent code here, we do asynchronous spawning/waiting of processes.

mosteo: OK, thanks.

AldanTanneo:
I added test durations in structured and regular output.

> Actually, Alire used to output results in JUnit format, and the code is still at `Alr.Testing.JUnit`. Not sure what demand for that is there these days.

Well, the XML format used by JUnit is not one of the structured outputs Alire supports. @Fabien-Chouteau thoughts?

Maybe a readable machine format is already good enough (I think cargo does the same with its JSON output: it's an ad-hoc format and people can build adapters around it). Also, I don't want to stray too far from the initial goal, which is to provide a minimal test runner. If people want a more complex one with big platform integrations and so on, it's not really the scope of this test runner.

Fabien-Chouteau:
My only input is that we should aim for a format that is natively supported by the likes of Gitlab and GitHub (e.g. here https://github.com/marketplace/actions/test-reporter).

AldanTanneo:
Then I think we'll want a specific switch on the test command to do that. The Alire 'structured output' format can be JSON, TOML or YAML, it doesn't make much sense to conform to a schema in 3 different formats.

We can still have the structured outputs for parsing test output directly, and then a `--junit=<filename>` switch to output test results to a file.

mosteo:
Yes, I think a specific switch here makes sense, the general structured output is mainly intended for Alire-specific output.

The action linked by Fabien is not something native from GitHub but an adapter of many formats to whatever GitHub uses? Anyway I read it is what is generally used.

Trivia, in the very beginning I implemented the junit format because the CI suite at the time (I can't even remember the name anymore, it was well before GH Actions) understood that one natively.

src/alire/alire-utils.adb
```diff
@@ -294,4 +294,62 @@ package body Alire.Utils is
...
+   ---------------------
+   -- Format_Duration --
+   ---------------------
+
+   function Format_Duration (D : Duration) return String is
+      I, F : Long_Integer;
+
+      --  account for rounding up
+      SECONDS_MAX : constant := 59.9995;
+      MINUTES_MAX : constant := 3599.5;
```
mosteo: We don't usually use all caps for constants. Is this gnatformat doing?

testsuite/tests/test/structured-output/test.py
```diff
@@ -0,0 +1,69 @@
...
+structure_tests(data)
+
+p = run_alr("--format=yaml", "test", complain_on_error=False)
+data = yaml.safe_load(p.out)
+structure_tests(data)
+
+p = run_alr("--format=toml", "test", complain_on_error=False)
+try:
+    import tomllib
+    data = tomllib.loads(p.out)
```
mosteo: If common enough this can be added to the testsuite requirements.txt. It already contains `toml` in fact.

mosteo (approved):
See a couple of minor comments.

src/alire/alire-utils.adb
AldanTanneo: No that's just me, force of habit from other languages. I'll change it.

testsuite/tests/test/structured-output/test.py
AldanTanneo: Oh I didn't check for it, sorry

I knew tomllib was in the standard library after a certain python version but yeah using a dedicated library does make sense if we're using older pythons on some platforms.

Issue #2066, opened 2026-01-17
[RFC] Check if an alire project already exists in directory skip project questions


AdaOrbit:
Currently when running `alr init foo` alire asks a lot of questions about the crate like title, description, license, etc and creates the project directory with `foo/alire.toml`. But when running the same command twice alire asks _all_ the questions again just to fail at the end with 

```
warn: foo/alire.toml already exists.
error: Cannot create 'foo/alire.toml'
``` 
This is somewhat annoying since you spend some time to fill out all the questions and get an error at the end that could have been checked before. 

Proposal: check beforehand if an alire project already exists. If if does: don't even bother to ask the developer for all the crate descriptions.

PR #2075, opened 2026-01-30
Check if directory is free to be used

AdaOrbit:
This PR solves https://github.com/alire-project/alire/issues/2066 by checking if the chosen project name is free.
for `--in-place` it looks if `alire.toml` already exists. On normal project creation it checks for `<project name>/alire.toml`

##### PR creation checklist
- [X] A test is included, if required by the changes.


AdaOrbit:
@AldanTanneo @mosteo ready for review!

AldanTanneo:
Ideally, there would be an error (overridable with `--force`) if the target directory is not empty (we don't want to overwrite existing files, or fail with a partially initialized crate.)

Is this done already elsewhere?

AdaOrbit:
The current behavior _is_ to partially initialize a crate if `alire.toml` already exists. Runnig `alr init --no-skel foo` and then `alr init foo` results in `foo.gpr`, `share/` and `src/foo.ads` being created but then erroring out because `alire.toml` exists and _skipping_ `alire/` creation.

And if I understand it correctly `alire.toml` is what makes a directory an alire crate. What I do to mitigate this is by checking if the `alire.toml` file already exists. Thus stopping the user from partial creation. 

I tried finding something similar that does those checks in the code base but couldn't find anything. (Maybe you know more and could guide me?)

Also a comment on the `--force` command for init: I don't think that it is a good idea. It should rather be the developers responsibility on ensure all the files that alire is going to "overwrite" are already gone and usable. Or else we end up with users who ask "I used --force and `main.gpr`, `src/main.ads` and more got overwriten. How to fix?"

AldanTanneo:
> Also a comment on the `--force` command for init: I don't think that it is a good idea. It should rather be the developers responsibility on ensure all the files that alire is going to "overwrite" are already gone and usable. Or else we end up with users who ask "I used --force and `main.gpr`, `src/main.ads` and more got overwriten. How to fix?"

What I meant is not using `--force` to overwrite files, but to say "try anyway" when the directory is not empty. It would still fail if it tries to overwrite existing user files.

Detecting when the directory is not empty would avoid the partial crate initialization altogether, and then using `--force` would potentially end up with a partially initialized crate (with skeleton files) but the error would not be `alr`'s fault. Does that make sense?

AldanTanneo:
Of course, this can be on top of `alire.toml` detection, as the crate creation would fail unconditionally if there is already an `alire.toml` file present.

AdaOrbit:
I should look more into the control flow to understand the current design better. I'll take a look at it again and see where it's best to check and add features.

AdaOrbit:
I'm gonna work on this again but I'd rather start from new.

PR #2078, opened 2026-02-17
feat: add a --list flag to `alr test`

AldanTanneo:
The `--list` flag outputs a list (well, a map in structured output mode) of tests matching the input filter. This allows external tools (such as the `ada_language_server`) to discover ahead of time what tests are available, and potentially to offer users to run them.

### DIFF
--- a/src/alire/alire-properties-tests.adb
+++ b/src/alire/alire-properties-tests.adb
@@ -139,9 +139,16 @@ package body Alire.Properties.Tests is
 
             if Local.Pop (TOML_Keys.Test_Id, Val) then
                if Val.Kind /= TOML_String
-                  or else Id_Set.Contains (Val.As_String)
+                 or else Id_Set.Contains (Val.As_String)
                then
                   Local.Checked_Error ("id must be a non-empty unique string");
+               end if;
+               if (for some C of Val.As_String =>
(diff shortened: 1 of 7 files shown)


AldanTanneo:
Maybe the issue with the 'corrupted linked list' is that gcc-13 and the system glibc (on ubuntu latest) are mismatched?

mosteo:
> Maybe the issue with the 'corrupted linked list' is that gcc-13 and the system glibc (on ubuntu latest) are mismatched?

I hadn't never seen this up to the particular PR where it started. It is also always the same test. So it seems to be something obscure but reproducible. I would expect such a mismatch to be more evident sooner? But I don't really know.

mosteo (approved):

AldanTanneo:
Hm actually in the current form this will output the list for the first declared alire runner and then bailout... I should add some checks when using the flag :thinking:

Fabien-Chouteau (approved):
Looks good to me as is.

There's one thing I would do to make this more future proof, `Get_File_List` should return a list/array of records with more info than only a path. Of course I am thinking about my #1935 proposal here, but beyond that it makes sense to compute the display name, source path, bin path, in a single place.

AldanTanneo:
I will make a followup PR to change the format of both the structured test output and the structured test list output then. (I implemented the structured output as an object instead of just a string for potentially adding fields in the future, so this was on my mind :+1:)

Issue #2082, opened 2026-02-25
Tweak test & test list output format


AldanTanneo:
Followup to #2078 

- [ ] move the test output and test list output from a map of tests to a list (we don't check the unicity of test ids currently)
- [ ] add more info to `Get_File_List` output, notably the expected location/name of the executable

PR #2083, opened 2026-02-25
Remove test list gpr [Don't merge]

Fabien-Chouteau:
While trying #2082 I realized that to build an executable with gprbuild you can just pass the source file on the command line:
```
# gprbuild main.adb
```
So I did a quick check and it turns out you can do this even if main.adb is not listed in `for Main use ();` in the GPR file.

Using this feature, we can get rid of the system that produces a GPR file with the list of tests to build.

Advantages:

- Less code
- The test crate is "buildable" right away (see below). No need to run Alire to create the missing GPR file
- The less GPR files there are, the happier I am ^^
 
Drawbacks:

- Building the tests crate without passing args will not build any executable...

So it's not clear whether or not this a positive change. I just wanted to record this option.

Ping @AldanTanneo

### DIFF
--- a/src/alire/alire-properties-tests.adb
+++ b/src/alire/alire-properties-tests.adb
@@ -139,9 +139,16 @@ package body Alire.Properties.Tests is
 
             if Local.Pop (TOML_Keys.Test_Id, Val) then
                if Val.Kind /= TOML_String
-                  or else Id_Set.Contains (Val.As_String)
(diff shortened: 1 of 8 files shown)


AldanTanneo:
This doesn't seem to work for nested tests right now. I like the idea but I don't know if it's very feasible to pass a long testsuite entirely on the command line. My first attempt at this as you know was through environment variables...

PR #2099, opened 2026-04-09
feat: omit compiler version from linked dependencies' build hashes

Seb-MCaw:
In offline environments, or other contexts where the community index is not available, it is currently necessary to explicitly set `dependencies.shared` to `false` (or construct a private index, with its own `gnat_external` definition) to build a crate with any dependencies, even if all dependencies are pins to local paths. In particular, this is required for `alr test`'s build of a nested test crate.

This PR omits the compiler version from the build hash of linked dependencies, so that default settings permit building a crate with no index-fetched dependencies without an index configured.

As far as I can tell, the fact that compiler version changes are no longer propagated to the build hashes of any shared dependents is unproblematic, as the compiler version is still hashed directly by such dependents themselves. This directly contradicts the existing comment, though, so I'm not at all confident that I have understood correctly.

##### PR creation checklist

### DIFF
--- a/src/alire/alire-builds-hashes.adb
+++ b/src/alire/alire-builds-hashes.adb
@@ -257,13 +257,18 @@ package body Alire.Builds.Hashes is
             Add_Externals;     -- GPR externals
             Add_Environment;   -- Environment variables
 
-            --  In the root crate we can skip compiler detection, as it has no
-            --  bearing on the hash or config regeneration. This allows most
-            --  operations in a crate without dependencies to succeed even in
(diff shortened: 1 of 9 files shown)


mosteo:
> As far as I can tell, the fact that compiler version changes are no longer propagated to the build hashes of any shared dependents is unproblematic, as the compiler version is still hashed directly by such dependents themselves. This directly contradicts the existing comment, though, so I'm not at all confident that I have understood correctly.

I think your analysis is right. The comment likely didn't account for the dependent also having the compiler version in its hash.

mosteo (approved):

PR #2101, opened 2026-04-14
add flags to forward gprbuild arguments with alr test

AldanTanneo:
This feature is made with the integration of GNATcoverage in mind. Users can now enter
`alr test -B--src-subdirs=gnatcov-instr...` and the like.

I plan to publish a small wrapper program in the index to have easy code instrumentation and coverage based on `alr test` :)

##### PR creation checklist

### DIFF
--- a/src/alire/alire-test_runner.adb
+++ b/src/alire/alire-test_runner.adb
@@ -552,9 +552,11 @@ package body Alire.Test_Runner is
    ---------
 
    function Run
-     (Root   : in out Roots.Root;
-      Filter : AAA.Strings.Vector := AAA.Strings.Empty_Vector;
-      Jobs   : Natural := 0) return Integer
+     (Root       : in out Roots.Root;
+      Filter     : AAA.Strings.Vector := AAA.Strings.Empty_Vector;
+      Jobs       : Natural := 0;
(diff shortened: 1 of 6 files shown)


Fabien-Chouteau (approved):
You told me this is in particular to allow the use of GNATcov, is there a way (potentially in addition to this) to make the process more automatic when doing coverage?

AldanTanneo:
> You told me this is in particular to allow the use of GNATcov, is there a way (potentially in addition to this) to make the process more automatic when doing coverage?

I'd like to have a wrapper script/command that would run the right commands to run tests with coverage instrumentation.

Another, more general idea would be to have a GPRFLAGS environment variable that Alire would pick up, and would apply regardless; unless one already exists in GPRbuild? I know ADAFLAGS is used in the Alire generated GPR files.

Fabien-Chouteau:
> I'd like to have a wrapper script/command that would run the right commands to run tests with coverage instrumentation.


Ideally this would be just a switch for `alr test` with the built-in runner.

> Another, more general idea would be to have a GPRFLAGS environment variable that Alire would pick up, and would apply regardless; unless one already exists in GPRbuild? I know ADAFLAGS is used in the Alire generated GPR files.

I don't think there's something like this for GPRbuild.

mosteo (approved):

PR #2107, opened 2026-05-07
feat: workflow improvements for immutable releases

mosteo:
Github requires a release to be not a draft to trigger actions. This can be a quagmire if some build fails, as release watchers will have been notified, immutable releases don't allow uploading assets for manual fixing, etc.

With this patch, we instead trigger release actions by pushing a tag. Once builds succeed, the release is automatically created in draft mode, which doesn't yet notify anybody and allows fixing the release if something goes wrong. We manually publish the draft once everything is OK.

Tested in https://github.com/mosteo/alire/actions/runs/25383957370


.github/workflows/ci-release.yml
```diff
@@ -11,8 +11,9 @@ on:
       - '**.md'
       - '**.rst'
       - '**.txt'
-  release:
-    types: [published]
+  push:
+    tags:
+      - 'v[0-9]*.[0-9]*.[0-9]*'
```
AldanTanneo:
```suggestion
      - 'v[0-9]+.[0-9]+.[0-9]+'
```

.github/workflows/ci-release.yml
```diff
@@ -186,22 +187,26 @@ jobs:
...
+      upload-url: ${{ steps.create_release.outputs.upload_url }}
       version-without-v: ${{ steps.get_version.outputs.version-without-v }}
 
     steps:
-      - name: Retrieve upload URL for the release
-        id: get_release
-        uses: bruceadams/get-release@v1.3.2
+      - name: Create draft release
+        id: create_release
+        uses: softprops/action-gh-release@v2
```
AldanTanneo: External actions should probably be pinned to a specific commit hash, especially in a sensitive context like creating a release.

.github/workflows/ci-release.yml
mosteo: Right, thanks.

PR #2108, opened 2026-05-09
fix: Always parse user.github_login as a String 

AdaOrbit:
PR fixes https://github.com/alire-project/alire/issues/2102 by implementing `Set_User_GitHub_Login is new CLIC.Config.Edit.Set_Typed`. The `CLIC` gitsubmodule had to be updated to fix an error. ~~But most importantly this PR introduces **pragma Ada_2022** to `alire-settings-edit.ads` since `String'Image` is needed for `Set_Typed`.~~

~~If anyone knows how to fix the Image "issue" I'd be open to fix it but as far as I could see there is no other simple fix :/~~

The login edge cases were added to the testsuite.

EDIT: I removed the ada2022 pragma as gcc^10 and gcc^11 use "pramga Ada_2020" instead of "pragma Ada_2022" and "re implementing" the 'Image attribute was not that hard _if_ you know how to do it.

##### PR creation checklist
- [X] A test is included, if required by the changes.
- [ ] `doc/user-changes.md` has been updated, if there are user-visible changes.
- [ ] `doc/catalog-format-spec.md` has been updated, if applicable.
- [ ] `BREAKING.md` has been updated for major changes in `alr`, minor/major in catalog format.

PR #2110, opened 2026-05-10
Remove obsolescent ASCII

jquorning:
Replace obsolescent `ASCII` package with `Ada.Characters.Latin_1`.

Added compiler configuration file `alire.adc`. GNAT does not flag `ASCII` as obsolescent with `-gnatwa/-gnatwj` even though it is. `pragma Restrictions (No_Obsolescent_Features)` does.

There are still around 10 grep references to `ASCII` in the code in output to user and comments.

### DIFF
--- a/src/alr/alr.ads
+++ b/src/alr/alr.ads
@@ -1,3 +1,4 @@
+with Ada.Characters.Latin_1;
 private with Ada.Strings.UTF_Encoding.Wide_Wide_Strings;
 
 with Alire;
@@ -6,6 +7,8 @@ with Simple_Logging;
 with CLIC.TTY;
 
 package Alr with Preelaborate is
+
+   package Latin_1 renames Ada.Characters.Latin_1;
 
    --  Nothing of note in this root package. Entities declared here are
    --  generally useful everywhere or in many packages: Exceptions for
--- a/src/alr/alr.ads
+++ b/src/alr/alr.ads
@@ -8,7 +8,7 @@ with CLIC.TTY;
 
 package Alr with Preelaborate is
 
-   package Latin_1 renames Ada.Characters.Latin_1;
+   package Latin_1 renames Alire.Latin_1;
 
    --  Nothing of note in this root package. Entities declared here are
    --  generally useful everywhere or in many packages: Exceptions for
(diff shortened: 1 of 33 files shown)


AldanTanneo:
Not entirely convinced this is useful, but if we want to do this: a reexport of the `Latin_1` package should probably be introduced at the root `Alire` package?

jquorning:
[`ASCII/ISO 646`](https://docs.adacore.com/live/wave/arm95/html/arm95/arm95-j-5.html) is in [Annex J](https://docs.adacore.com/live/wave/arm95/html/arm95/arm95-j.html) since nearly forever and should not be used in new programs.

`-gnatwj/-gnatwa` does not flag `ASCII` usage but `No_Obsolescent_Features` restriction [does](https://docs.adacore.com/gnat_ugn-docs/html/gnat_ugn/gnat_ugn/building_executable_programs_with_gnat.html#index-255).

The PR is just a clean-up.

jquorning:
Anyone got an idea what is wrong [during testing](https://github.com/alire-project/alire/actions/runs/25709066223/job/75485302642?pr=2110#step:9:1675)?

It looks like @AdaOrbit ran into same problems with [#2108](https://github.com/alire-project/alire/pull/2108).

`workflows__get-build-run-clean` returns error code 1 from `clean`. 

It works on my machine, Debian 13.

AdaOrbit:
YEAH! I am currently breaking my head against the wall trying to figure it why it wont work. But I have no idea...

jquorning:
> YEAH! I am currently breaking my head against the wall trying to figure it why it wont work. But I have no idea...

It comforted me also to see I'm not alone.

There has been a couple of updates to the testsuite lately, but they look okay. What is more interesting is that `gprbuild` in the index was updated yesterday. 

I wondered why a simple update suddenly caused a lot of failures.

It has something to do with `-r` and `-p` flags for `gprclean`, I think. `obj/development` directory is not deleted.

src/alr/alr.ads
```diff
@@ -7,6 +8,8 @@ with CLIC.TTY;
 
 package Alr with Preelaborate is
 
+   package Latin_1 renames Ada.Characters.Latin_1;
```
AldanTanneo:
```suggestion
   package Latin_1 renames Alire.Latin_1;
```

AldanTanneo (approved):

mosteo (approved):

mosteo (approved):

PR #2117, opened 2026-05-15
dev: remove empty dirs within `obj` even if gprclean didn't

mosteo:
Should fix the recent breakage in the testsuite for a change in behavior in `gprclean`.


testsuite/tests/workflows/get-build-run-clean/test.py
```diff
@@ -23,6 +24,20 @@
...
+# obj/development/xxx.cswi
+# This is likely a bug in gprclean, so instead of checking that obj/ is empty,
+# as we did before, we check that remaining files have one of those extensions.
+
+GPRCLEAN_BUG_EXTENSIONS = {'.stderr', '.stdout', '.bexch', '.cswi'}
+spurious = [f for f in contents('obj')
+            if os.path.isfile(f)
+            and os.path.splitext(f)[1] not in GPRCLEAN_BUG_EXTENSIONS]
+assert not spurious, \
+    "Unexpected files after clean: " + str(spurious)
```
mosteo, line 41: @Fabien-Chouteau is this known/intended?

Fabien-Chouteau: I don't know. I'll try to get the correct info :)

Fabien-Chouteau: So "gprclean2" is not cleaning the artifacts produced by "gprbuild1". We mix both in the GNAT-FSF-Builds so that's why.

jquorning: Maybe it has something to do with the `-p` switch? (not `-P`)

mosteo: It's not related to `-p`, at least according to my tests, although I wasn't thorough. Plus, I've just fixed some changes in wording in the output of gprbuild that seem to affect only the Windows version (?).

AldanTanneo (approved):

PR #2119, opened 2026-05-15
feat: don't look for containing test runner if legacy tests exist

mosteo:
Fixes a corner case where we would go looking for a test runner even if there are legacy tests in the current crate.

Also adds `--here` to disable this 'clever' behavior in case some other obscure case arises.


AldanTanneo:
Wasn't the `--legacy` flag already doing what the `--here` flag does?

edit: hm not exactly, however it seems that the `--legacy` flag may be redundant now? Unless a crate has both the old test actions and the new test runner sections.

mosteo:
> edit: hm not exactly, however it seems that the `--legacy` flag may be redundant now?

Yes, indeed the `--legacy` switch is already marked as obsolescent. So at some point we'll remove it.

PR #2135, opened 2026-06-03
feat: test configuration via pragmas

mosteo:
Based on #1935. I actually rebased that one so that has resulted in a somewhat messy `alire-tests_runner.adb`, as it had diverged somewhat from master, sorry about that.

There are a bunch of TODOs but I didn't want this one to grow anymore, so those should be easy to add in a follow-up.

##### PR creation checklist


src/alire/alire-test_runner.adb
```diff
@@ -538,16 +878,23 @@ package body Alire.Test_Runner is
                 (This.Path,
                  Prefix => (Root.Path / "src") & OS_Lib.Dir_Separator));
       begin
-         if AAA.Strings.Has_Suffix (String (Name), ".adb")
-           and then Matches_Filter (Name)
-         then
-            Test_List.Append (Name);
+         --  TODO: skip files that also have an ".ads" counterpart
```
Fabien-Chouteau, line 891: Why do you want to skip them?

templates/crate_test/tests/test_pragmas.adc
```diff
@@ -0,0 +1,5 @@
+--  `pragma Alire_Test (...)` clauses in test sources are read by the alr test
```
Fabien-Chouteau: Can we have a way to do this without a `.adc` file ? Like a compiler switch.

I don't like having to introduce yet another kind of file if we can avoid it.

testsuite/tests/test/pragmas/test.py
```diff
@@ -0,0 +1,335 @@
...
+# Named form: `pragma Alire_Test (Key => Value)` is accepted as
+# equivalent to the positional `(Key, Value)` form.
+
+write_test(
+    "named",
+    "null;",
+    prelude='pragma Alire_Test (Name => "pragma_named");\n',
+)
+
+# No Alire_Test pragma at all (overkill but...)
```
Fabien-Chouteau: In my POC, the absence of the `pragma Alire_Test` indicates that this file is not a test.

For me it was an important part of the POC because it means you can write test support packages without having them compiled and tested (which would lead to an error because they are pacakges and not procedures).

And the files that are actual tests are clearly identified.

testsuite/tests/test/pragmas/test.py
```diff
@@ -0,0 +1,335 @@
...
+assert_eq(0, data["summary"]["failures"])
+assert_eq(1, data["summary"]["skipped"])
+assert_eq("skip", find_test(data["tests"], "bogus_key")["status"])
+assert_substring(
+    "unknown Alire_Test pragma key: bogus",
+    find_test(data["tests"], "bogus_key")["reason"],
+)
+
+
+print("SUCCESS")
```
Fabien-Chouteau: That's a lot of tests in a single file. It would be nice to split.

templates/crate_test/tests/test_pragmas.adc
Fabien-Chouteau: `-gnatwG` ‘Suppress warnings on unrecognized pragmas.’

This switch suppresses warnings for unrecognized pragmas.

src/alire/alire-test_runner.adb
mosteo: Each main procedure is converted into a test, and if there is a .ads, that must be a support package. Currently, it tries to convert the body into a test and it fails (IIRC).

mosteo: Actually this should allows us to get rid fo the tests/common/ folder which is now used just to store the root test package.

templates/crate_test/tests/test_pragmas.adc
mosteo: there are two ways I can think of: using -gnatwG (if not that one, there's for sure one to ignore unknown pragmas), but that disables the warning for all unknown pragmas, not only ours.

Or we can add `pragma Ignore_Pragma (Alire_Test);` to the root test package. Seems indeed simpler and better.

testsuite/tests/test/pragmas/test.py
mosteo: I see. The current behavior is that anything in `tests/src` is a test. I proposed to ignore packages, that can obviously not be a test. For the rare cases where a stand-alone .adb is not a test, I had introduced `Auxiliary_File`.

So I understand the idea would be to be able to use, for example, `pragma Alire_Test;` even with arguments to explicitly mark the file as a test.

testsuite/tests/test/pragmas/test.py
mosteo: Will do.

src/alire/alire-test_runner.adb
Fabien-Chouteau: > if there is a .ads, that must be a support package

You can have `.ads` files for the `.adb` of procedures too.
In my opinion, it's better to ask for at least one pragma to mark the source file as a test.

templates/crate_test/tests/test_pragmas.adc
Fabien-Chouteau: I don't think it's a problem to ignore all unknown pragmas in the tests. But if `pragma Ignore_Pragma (Alire_Test);` in  the root package works for all sources then it's an ok solution I would say.

testsuite/tests/test/pragmas/test.py
Fabien-Chouteau: I would say that you have to at least provide a test name:
```ada
pragma Alire_Test (Name => "My Test");
```

src/alire/alire-test_runner.adb
mosteo: Created #2138 to track this one.

testsuite/tests/test/pragmas/test.py
mosteo: Since this is a change of the current implementation, I propose to address it in a follow-up since it is orthogonal to the main purpose of this PR. Tracked in #2139.

templates/crate_test/tests/test_pragmas.adc
mosteo: Changed to using `Ignore_Pragma`.

Fabien-Chouteau (approved):

src/alire/alire-test_runner.adb
mosteo: > You can have `.ads` files for the `.adb` of procedures too.

Well, go figure... This one got me good.

PR #2157, opened 2026-07-29
feat: explicit declaration of tests

mosteo:
At least one `pragma Alire_Test` is now needed for a main to be considered a test. A main without it is a (recoverable) error (to prevent forgetting to enable a test and silently missing it among many others). 

The pragmas parser also returns now the kind of unit, so specs, generics, packages are safely detected and ignored, and giving them a pragma is also considered an error. This in turn allows removing the extra `common/` folder which was a necessity of previously blindly considering anything `*.adb` a test, as now everything can live under `src/`

Includes a fix for an unrelated issue with Windows' tar triggered by a new dependency source archive.

##### PR creation checklist


doc/user-changes.md
```diff
@@ -6,6 +6,34 @@ stay on top of `alr` new features.
 
 ## Release `3.0`
 
+### Built-in tests must be declared with `pragma Alire_Test`
```
AldanTanneo: Since the test runner is part of 3.0 as well, shouldn't this changelog entry be merged into the test runner one?

mosteo: Yes, it's probably better to have a single entry. This one was written more with the people that are already using it in mind, but it will be confusing.

Issue #107, opened 2019-04-25
Optional dependencies


mosteo:
cargo/opam have them, I can see the use case, need has not arisen yet

An optional dependency should fork the solver into two solution trees, one with and another without it (in case the dependency is available but makes the solution impossible).

PR #2167, opened 2026-08-08
Add Cargo-style additive package features

yrashk:
## Summary

This PR adds explicit, additive package features to Alire, including feature-gated dependency edges, dependency-feature forwarding, root feature selection, and resolved feature configuration for Ada, C, and GPR.

The central model is deterministic and package-controlled:

- features are selected explicitly by manifests or CLI options;
- feature requests from all dependency edges are unified additively;
- selected features may activate dependency edges marked `optional = true`;
- weak forwarding such as `crate?/feature` forwards a feature only when another feature or non-optional edge activates that dependency.

Here, `optional = true` is a property of an edge controlled by package features. It does not mean that the solver should opportunistically include a dependency when solvable. That best-effort model is the separate proposal described in [#107](https://github.com/alire-project/alire/issues/107).

## Motivation

A crate may expose integrations that should not become unconditional dependencies.

For example, a standalone benchmarking or reporting crate may support integration with another library while remaining independently usable. Consumers should be able to request that integration explicitly, and transitive crates should be able to forward the corresponding features without forcing activation.

Today, crate configuration can change generated source and GPR values, but it cannot change the dependency graph. Conditional dependencies are limited to platform properties. Consequently, Alire cannot currently represent this kind of package capability without making the integration unconditional or creating a separate crate.

## Manifest model

Dependencies that need feature-related settings use an inline table:

```toml
[[depends-on]]
backend = { version = "^1", optional = true }
reporter = { version = "^2", features = ["json"], default-features = false }
```

Features are declared in a top-level table:

```toml
[features]
default = []
backend-support = ["dep:backend"]
backend-diagnostics = ["backend?/diagnostics"]
serialization = ["reporter/json"]
```

Feature references have four forms:

- `name` selects another local feature;
- `dep:crate` activates a feature-gated dependency edge;
- `crate/feature` activates the dependency when necessary and requests its feature;
- `crate?/feature` requests the dependency feature only if the dependency is activated elsewhere.

Feature names are case-insensitive. Feature selection is strictly additive: enabling one feature cannot disable another.

## Resolution semantics

Feature selection participates in dependency solving rather than being applied after a solution is chosen.

For each selected release, Alire unifies:

- requested features from every dependency edge;
- whether any edge requests the release's `default` feature.

The solver then expands the resulting feature closure. Newly activated dependencies and forwarded feature requests may introduce further releases or feature requests, so expansion continues to a fixed point.

A release that does not define an explicitly requested feature is rejected as a candidate. Inactive feature-gated edges do not appear in the resulting solution.

This produces one unified feature set per selected crate release, including when multiple dependency names are fulfilled by the same release through `provides`.

## CLI behavior

Root features can be selected during a build:

```console
alr build --features=feature-a,feature-b
alr build --no-default-features
```

Feature settings can also be recorded on a newly added dependency edge:

```console
alr with crate --features=feature-a --no-default-features
```

Root selections are persisted in the lockfile. Commands without feature switches adopt the stored selection instead of silently reverting to defaults.

## Build and GPR integration

Resolved feature state is exposed through generated configuration:

- Ada: `Feature_<Name>`
- GPR: `Feature_<Name>`
- C: `FEATURE_<NAME>`

Inactive feature-gated dependencies are excluded from:

- the resolved solution;
- build ordering;
- dependency graphs;
- `GPR_PROJECT_PATH`;
- generated automatic GPR `with` clauses;
- dependency build hashes.

Unified feature selections and default-feature state are included in shared-build hash inputs, preventing feature variants from sharing incompatible cached builds.

GNAT and GPRbuild do not require changes. They consume the resolved dependency closure and generated configuration through Alire's existing integration paths.

## Persistence and compatibility

Dependency-edge feature requests are stored in solution states and lockfiles. The root feature selection is stored separately so subsequent commands can reconstruct the same graph.

Feature-bearing index entries are gated on proposed index format 1.5. Older clients should therefore reject an unsupported index before encountering feature syntax they cannot parse.

Feature-free manifests retain their existing dependency representation, and default lockfiles do not gain feature-only fields.

The index-format version remains subject to maintainer confirmation before this PR is marked ready.

## Implementation outline

The implementation is divided across these existing boundaries:

1. **Feature model and validation**
   - Adds package feature definitions and selection types.
   - Validates local references, dependency activation, forwarding, normalized-name collisions, and weak forwarding.

2. **Dependency representation**
   - Adds feature requests, default-feature selection, and the feature-gated edge flag to dependencies and dependency states.
   - Preserves this information through TOML and lockfile serialization.

3. **Solver**
   - Unifies requests across edges.
   - Expands activated dependencies and forwarded features to a fixed point.
   - Rejects release candidates that do not provide requested features.

4. **Resolved integration**
   - Uses the unified selection for dependency graphs, automatic GPR integration, generated configuration, and build hashing.
   - Keeps inactive edges outside all downstream build inputs.

5. **CLI**
   - Adds root feature switches to `alr build`.
   - Adds dependency-edge feature switches to `alr with`.

## Draft status and review questions

This is opened as a draft to establish agreement on the architecture and semantics before treating the format as final.

In particular:

- Is explicit additive feature selection the desired model for package-controlled capabilities?
- Is index format 1.5 the appropriate compatibility boundary?
- Should root feature selection remain process-level CLI state, or be owned more directly by the root/solver API?
- Are additional restrictions needed for feature-activated toolchain dependencies?

Before marking the PR ready, the manifest schema and `doc/user-changes.md` should be updated, and additional focused coverage should be considered for shared-build hash separation, automatic GPR `with` generation, and lockfile compatibility.

##### PR creation checklist

- [x] `BREAKING.md` has been updated for catalog-format compatibility.


yrashk:
CI follow-up after the feature fixes:

- 1ef445e6 restores legacy build hashes by adding feature inputs only for releases that actually define features.
- 3ae05ca1 compares dependency states by persisted semantics, excluding the presentation-only feature-syntax flag that caused phantom solution changes after lockfile reloads.
- The first fresh matrix then passed every feature-related check. Its only failure was in the two live self-update tests, where GitHub reported that the shared runner IP had exhausted its anonymous API limit.

I initially tried mapping the workflow-provided GITHUB_TOKEN to Alire’s GH_TOKEN in 34e45188. A fresh ARM release job showed that this credential is not reliably accepted by the release API on every runner (HTTP 401), so de109cd4 removes that mapping and replaces it with the narrower workaround: both self-update tests capture their command output and report a skip only when Alire explicitly says the GitHub API rate limit was exceeded. Every other self-update error remains a failure. The latest-release test now also keeps output available on failure, avoiding the unrelated NameError that previously obscured the actual API error.

The workaround is intentionally scoped to the two tests that depend on the live GitHub releases API. Its exact rate-limit path was verified locally with both tests reporting SKIP, and the corrected aarch64-linux Release job passed its full suite. A separate aarch64-macos retry failed before building the PR because the third-party architecture checker triggered ADA.IO_EXCEPTIONS.DEVICE_ERROR: Broken pipe while piping gprbuild --version into grep -q. Commit d3f364ff documents the narrow skip inline and triggers a clean retry of that unrelated flake.

yrashk:
@mosteo ping?

mosteo:
Sorry @yrashk , this is a substantial patch and it may take some quality time to review. I'll get to it ASAP.

yrashk:
I do understand it is substantial, just wanted to make sure it gets looked at :) 

Would be great to have this feature.

yrashk:
I have prepared an upstream follow-up in #2177 that explores gated delivery for this feature.

The idea is to allow package features to merge and receive real-world testing without changing behavior for existing users: support remains disabled unless `ALIRE_GATE_FEATURES` is present.

This may also be a useful general pattern for substantial features – letting them ship early and evolve through opt-in use instead of remaining indefinitely in PRs, while keeping the default user experience unchanged.

The temporary gate is isolated in an independently revertible commit, with permanent index compatibility fixes kept separately.

PR #2169, opened 2026-08-22
fix: avoid rehashing shared dependencies during clean

yrashk:
## Problem

Plain `alr clean` exports the current build environment before asking `Root.Build` to regenerate configuration. With shared dependencies, `Build_Prepare` computes hashes after that early export.

If a dependency has `Set` environment values containing `${CRATE_ROOT}`, hash compilation sees the exported absolute path of the existing shared build. Conflict checking is disabled during hashing, so those stale absolute paths become new hash inputs. Alire deploys a second shared build directory and the later checked export fails because the old and new `<CRATE>_ALIRE_PREFIX` and environment paths differ.

`Root.Build` already exports the build environment after shared build synchronization and configuration generation, before returning at the `Generation` stop point.

## Concrete reproduction

The regression creates a local indexed dependency `envdep=1.0.0` with a file origin and this manifest content:

```toml
project-files = ["envdep.gpr"]

[environment]
ENVDEP_INCLUDE.set = "${CRATE_ROOT}/include"
ENVDEP_SHARE.set = "${CRATE_ROOT}/share"

[[actions]]
type = "pre-build"
command = ["sh", "-c", "touch pre_build_ran"]
```

A fresh consumer enables shared dependencies, adds `envdep`, and runs:

```console
$ alr build --stop-after=generation
$ alr clean
```

After generation, the first shared deployment H1 has normalized hash inputs:

```text
environment:ENVDEP_INCLUDE=envdep_1.0.0_filesystem/include
environment:ENVDEP_SHARE=envdep_1.0.0_filesystem/share
```

On affected versions, plain `alr clean` exports the H1 environment before generation. Hashing then creates H2 with absolute inputs pointing back into H1:

```text
environment:ENVDEP_INCLUDE=/tmp/.../builds/envdep_1.0.0_filesystem/<H1>/include
environment:ENVDEP_SHARE=/tmp/.../builds/envdep_1.0.0_filesystem/<H1>/share
```

The final checked export fails with:

```text
Trying to set an already defined environment variable
ENVDEP_SHARE is already defined as /tmp/.../<H1>/share
but new value is /tmp/.../<H2>/share
```

Both H1 and H2 contain `alire/flags/complete_copy`; neither contains `pre_build_ran`, confirming that the dependency pre-build action is not involved.

The complete disposable case is in `testsuite/tests/clean/shared-env-rehash`.

## Solution

Remove only the redundant early export from `Alr.Commands.Clean.Execute`.

Add a disposable indexed shared-dependency regression with two crate-relative `Set` values. The test verifies that:

- hash inputs remain normalized;
- plain clean retains exactly one shared build directory;
- dependency pre-build does not run;
- `gprclean` still receives the environment exported after generation.

## Tests

- New `clean/shared-env-rehash` regression fails before the fix with the H1/H2 conflict and passes after it.
- Full clean-focused selector: 6/6 passed.
- Source rebuild with GNAT 16.1.0 and GPRbuild 26 passed.
- `build/hashes/*`: 6/7 passed; `transitive-linked-dep` reached its unrelated final mocked compiler-version step and failed in generated mock `main.adb` with an access-check `Constraint_Error`.


yrashk:
CI status: `clean__shared-env-rehash` passes across the matrix, including Windows. There have been no code changes after `d913d662`; `acdb1f11` and `7ce5eada` are no-content CI retriggers. Each retrigger run had only the unrelated Windows `self_update__specific_version` flake (expected 2.1, still saw 3.0.0-dev), on different GCC jobs. Could a maintainer please rerun only failed job `96986682038` in run `32554596648`? No further whole-suite retriggers are planned.

mosteo (approved):
@yrashk thanks for this fix.

PR #2174, opened 2026-08-25
fix: distinguish local pre-releases in the cache

Seb-MCaw:
Previously, local directory or source archive origins would be deployed to `<cache>/releases/<crate>_X.Y.Z_filesystem` regardless of any pre-release part of the version. Subsequent attempts to deploy another version differing only in the pre-release part would generate the same deployment directory, and therefore report success and the requested version (including in the generated `config/` and manifest), but continue using the previously deployed version's sources.

This PR includes the prerelease part, and the archive hash where available, in the cache directory name.

The build metadata is still omitted, since sanitising both `-` and `+` to `_` would introduce a new class of collision; as far as I can tell, `alr` treats crates differing only in the build metadata as equivalent anyway.

There remains a residual collision risk on case-insensitive filesystems, but this now raises an explicit error. I suppose this could alternatively trigger a full delete and re-deploy, but that didn't seem worth the effort for something so obscure.

##### PR creation checklist


Seb-MCaw:
Ready for review.

mosteo (approved):

PR #2176, opened 2026-08-29
Add misstyping suggestions from clic

AdaOrbit:
When misstyping a command like `alr cleam` instead of `alr clean`, the user will be informed of the closest valid command as long as the levenshtein edit distance is less then 3. If the distance is greater then 3 the default help message is printed. 

I chose 3 as the default here since it's good enough to find common spelling mistakes without "over"-searching probable commands.

Since it's a cosmetic change I didn't Include tests but I'm willing to add some if requested.

Also CLIC has been updated to v0.4 since I needed to add some features in order to get it to work. Until CLIC gets updated I point the dependency to my CLIC fork in order to get the builds working. And a parts of the alire utils got moved to clic since it was needed to determine the closest command. See https://github.com/alire-project/clic/pull/38

**DO NOT ACCEPT YET!!!** Merge the CLIC update first

##### PR creation checklist
- [X] A test is included, if required by the changes.
- [X] `doc/user-changes.md` has been updated, if there are user-visible changes.


AdaOrbit:
No clue why this one check failed... 

Except for the CLIC dependency I'm ready for review!

@mosteo DO NOT MERGE YET!!!

mosteo (changes requested):
This looks good but indeed we will need a test for it. I'll look into the clic PR now.

mosteo:
https://github.com/alire-project/clic/pull/38 is merged so you can update the pins here while you are at it.

AdaOrbit:
Need to rerun that one test but other then that it's good to go! @mosteo

AdaOrbit:
while writing the test I found another issue in clic...

commands that are three letters or shorter can't be tested. With length three the line https://github.com/alire-project/clic/blob/33f71b9381dde2bc32741be288b123f80a7fcb53/src/clic-utils.adb#L62 only returns true if Min_Dist is zero.

Glad that you made me write the test before submitting :+1:

mosteo:
You can bump clic now.

PR #2182, opened 2026-09-09
fix: clarify that index --del expects an index name, not a position

### LINKED ISSUE #2113: `alr index --del 1` errors with: Given index not found: 1
This is because an index name is expected, so the error message is misleading.

kokhlo:
## Root cause

`alr index --del <name>` matched the argument against configured index *names*. When the name is not found, `Delete` in `src/alr/alr-commands-index.adb` reported `Given index not found: 1`, which does not say that the argument is expected to be a name — and since `alr index --list` displays a positional `#` column, users naturally read it as "index #1 is missing" (`alr index --list` output shows `1 foo ...`, hence #2113).

## Fix

Replaced the message with one that names the expectation and points at how to discover valid names:

```
There is no configured index named '1'. --del expects an index name, not a position; use `alr index --list` to see configured index names
```

Wording follows the style of neighbouring messages (single-quoted names, backtick-quoted commands, e.g. `alr-toml_index.adb`).

## Testing

Repro on unpatched master (`alr index --add /tmp/empty-repo --name foo` then `alr index --del 1`):

```
ERROR: Given index not found: 1
```

Same scenario on this branch:

```
ERROR: There is no configured index named '1'. --del expects an index name, not a position; use `alr index --list` to see configured index names
```

New e3 regression test `testsuite/tests/index/index-del-not-found`:

```
$ python run.py tests/index/index-del-not-found
INFO     Found 1 tests
INFO     PASS            index__index-del-not-found
INFO     Summary:
  PASS         1
```

Sibling index tests (`python run.py tests/index`, network/docker disabled): **PASS 54, FAIL 1, SKIP 2** — the single failure (`index__branch-mismatch`) reproduces identically on unpatched master (pre-existing, unrelated: test expects community branch 'master' while fixtures now use 'main').

Fixes #2113


testsuite/tests/index/index-del-not-found/test.py
```diff
@@ -0,0 +1,30 @@
+"""
+Test that `alr index --del` expects an index name, not a position, and that
+deleting a name that is not configured gives a clear error.
+"""
+
+import re
```
AdaOrbit: `re` is not used. Please remove it.

src/alr/alr-commands-index.adb
```diff
@@ -72,7 +72,10 @@ package body Alr.Commands.Index is
       end loop;
 
       if not Found then
-         Reportaise_Command_Failed ("Given index not found: " & Name);
+         Reportaise_Command_Failed
+           ("There is no configured index named '" & Name & "'. "
+            & "--del expects an index name, not a position; use "
```
AldanTanneo, line 77: I think the message would be better without the "--del expects an index name, not a position;" sentence; ideally it should only be shown if the user entered a number, not any misspelled name. The current phrasing was not good, but simply having

```ada
"There is no configured index named '" & Name & "'. "
& "Use `alr index --list` to see configured index "
& "names"
```
should be enough, no?

AldanTanneo:
Also (a personal remark): I'm reluctant to merge a fix to an issue labelled "good first issue" with something that was most probably made by an LLM. So-called "good first issues" are meant for new human contributors to get to know the project and figure out how to contribute, and using an LLM for that defeats the purpose.

